PR #3, opened 2021-11-26
Generate github wiki

mandolaerik:
- Refactor script to avoid side-effects on toplevel
- Generate github wiki


andreashedstrom-intel (approved):
I didn't look too closely at the newly added code. It should be obvious from tests/local runs if it works or not. Approving so you can submit it when you think it's ready.

PR #4, opened 2021-11-29
Remove dollar from DML 1.4 error messages -- SIMICS-18421

lwaern-intel:


py/dml/logging.py
```diff
@@ -84,6 +87,10 @@ def report_error():
     if failure == max_errors:
         exit(2)
 
+def dollar(site=None):
```
mandolaerik, line 90: Never called with site=None, it seems

py/dml/objects.py
```diff
@@ -256,8 +256,8 @@ def logname_anonymized(self, indices=(), relative='device'):
         if self.isindexed():
             suff = "".join('[%s]' % i for i in
                            indices[-self.local_dimensions():])
-            suff += "".join('[$%s]' % idxvar for idxvar in
-                            self._idxvars[len(indices):])
+            suff += "".join(f'[{dollar(self.site)}{idxvar}]'
+                            for idxvar in self._idxvars[len(indices):])
```
lwaern-intel, line 260: This _is_ breaking! `logname_anonymized` is used unsaturated in order to generate event class names! That enough is a reason to revert this change.

py/dml/logging.py
lwaern-intel: True. Logic scrapped.

py/dml/objects.py
lwaern-intel: Reverted.

mandolaerik (approved):

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9560416

PR #5, opened 2021-11-29
Fix reset port templates

svenwest:
Signal interface API docs say that raise and lower represent the
rising and falling edge of the signal, and that it is an error for an
initiator to call raise twice, without an intervening call to lower.

Continuation of this PR
https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/942

### DIFF
--- a/lib/1.4/utility.dml
+++ b/lib/1.4/utility.dml
@@ -192,14 +192,49 @@ template _init_val_power_on_reset is (init_val, power_on_reset) {
     }
 }
 
+/**
+### state_signal
+
+<section>
+
+#### Description
+
+Implements a signal interface with saved state. The current state of the signal
+is stored in the saved boolean `high` and an error is logged if the signal is
+raised or lowered when already high or low. The methods `signal_raise` and
+`signal_lower` can be overridden to add side effects.
+
+#### Related Templates
+
+poreset, hreset, sreset
+
+</section>
+*/
+template state_signal {
+    implement signal {
+        saved bool high;
+        method signal_raise() default {
+            if (high)
+                log error: "%s raised when already high", qname;
+            else
+                high = true;
+        }
+        method signal_lower() default {
+            if (!high)
+                log error: "%s lowered when already low", qname;
+            else
+                high = false;
+        }
+    }
+}
+
 // Instantiate this on top level to support power-on reset.
 template poreset is power_on_reset {
-    port POWER {
+    port POWER is state_signal {
         implement signal {
             method signal_raise() {
+                default();
                 dev.power_on_reset();
-            }
-            method signal_lower() {
             }
         }
     }
@@ -234,12 +269,11 @@ template _init_val_hard_reset is (init_val, hard_reset) {
 
 // Instantiate this on top level to support hard reset.
 template hreset is hard_reset {
-    port HRESET {
+    port HRESET is state_signal {
         implement signal {
             method signal_raise() {
+                default();
                 dev.hard_reset();
-            }
-            method signal_lower() {
             }
         }
     }
@@ -301,19 +335,18 @@ for power-on reset, hard reset and soft reset, respectively.
 
 #### Related Templates
 
-power\_on\_reset, hard\_reset, soft\_reset
+state\_signal, power\_on\_reset, hard\_reset, soft\_reset
 
 </section>
 */
 
 // Instantiate this on top level to support soft reset.
 template sreset is soft_reset {
-    port SRESET {
+    port SRESET is state_signal {
         implement signal {
             method signal_raise() {
+                default();
                 dev.soft_reset();
-            }
-            method signal_lower() {
             }
         }
     }
--- a/test/1.4/lib/T_state_signal.dml
+++ b/test/1.4/lib/T_state_signal.dml
@@ -1,0 +1,51 @@
+/*
+  © 2021 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+dml 1.4;
+
+device test;
+
+import "utility.dml";
+
+saved bool on_raise_called;
+saved bool on_lower_called;
+
+method on_raise() {
+    on_raise_called = true;
+}
+
+method on_lower() {
+    on_lower_called = true;
+}
+
+port both is state_signal {
+    implement signal {
+        method signal_raise() {
+            default();
+            on_raise_called = true;
+        }
+        method signal_lower() {
+            default();
+            on_lower_called = true;
+        }
+    }
+}
+
+port up is state_signal {
+    implement signal {
+        method signal_raise() {
+            default();
+            on_raise_called = true;
+        }
+    }
+}
+
+port down is state_signal {
+    implement signal {
+        method signal_lower() {
+            default();
+            on_lower_called = true;
+        }
+    }
+}
--- a/test/1.4/lib/T_state_signal.py
+++ b/test/1.4/lib/T_state_signal.py
@@ -1,0 +1,58 @@
+# © 2021 Intel Corporation
+# SPDX-License-Identifier: MPL-2.0
+
+import stest
+stest.expect_equal(obj.on_raise_called, False)
+stest.expect_equal(obj.on_lower_called, False)
+stest.expect_equal(obj.port.both.signal_high, False)
+
+obj.port.both.iface.signal.signal_raise()
+stest.expect_equal(obj.port.both.signal_high, True)
+stest.expect_equal(obj.on_raise_called, True)
+stest.expect_equal(obj.on_lower_called, False)
+
+obj.on_raise_called = False
+obj.on_lower_called = False
+
+with stest.expect_log_mgr(obj=obj.port.both, regex="already high"):
+    obj.port.both.iface.signal.signal_raise()
+stest.expect_equal(obj.port.both.signal_high, True)
+stest.expect_equal(obj.on_raise_called, True)
+
+obj.port.both.iface.signal.signal_lower()
+stest.expect_equal(obj.port.both.signal_high, False)
+stest.expect_equal(obj.on_lower_called, True)
+
+obj.on_raise_called = False
+obj.on_lower_called = False
+
+with stest.expect_log_mgr(obj=obj.port.both, regex="already low"):
+    obj.port.both.iface.signal.signal_lower()
+stest.expect_equal(obj.port.both.signal_high, False)
+stest.expect_equal(obj.on_lower_called, True)
+
+obj.on_raise_called = False
+obj.on_lower_called = False
+
+obj.port.up.iface.signal.signal_raise()
+stest.expect_equal(obj.on_raise_called, True)
+stest.expect_equal(obj.on_lower_called, False)
+
+obj.on_raise_called = False
+obj.on_lower_called = False
+
+obj.port.up.iface.signal.signal_lower()
+stest.expect_equal(obj.on_raise_called, False)
+stest.expect_equal(obj.on_lower_called, False)
+
+
+obj.port.down.iface.signal.signal_raise()
+stest.expect_equal(obj.on_raise_called, False)
+stest.expect_equal(obj.on_lower_called, False)
+
+obj.on_raise_called = False
+obj.on_lower_called = False
+
+obj.port.down.iface.signal.signal_lower()
+stest.expect_equal(obj.on_raise_called, False)
+stest.expect_equal(obj.on_lower_called, True)
--- a/RELEASENOTES-1.4.docu
+++ b/RELEASENOTES-1.4.docu
@@ -237,4 +237,10 @@
       The internal representation of values of template types has been changed.
       It is no longer allowed to cast such values to integer or pointer types.
   </add-note></build-id>
+  <build-id value="next"><add-note>Added a new
+      template <tt>state_signal</tt> which implements a signal
+      interface with saved state. It logs an <tt>error</tt> if the
+      signal is raised or lowered when already high or low. The
+      templates <tt>hreset</tt>, <tt>sreset</tt> and <tt>poreset</tt>
+      use the <tt>state_signal</tt> template.</add-note></build-id>
 </rn>
--- a/test/1.4/lib/T_regdisp.py
+++ b/test/1.4/lib/T_regdisp.py
@@ -57,6 +57,7 @@ def test_some(mem, obj, port, allow_partial, allow_overlapping,
     def access(offset, length, partial = False, overlapping = False,
                illegal = False):
         obj.ports.HRESET.signal.signal_raise()
+        obj.ports.HRESET.signal.signal_lower()
         write(offset, length, partial, overlapping, illegal)
         read(offset, length, partial, overlapping, illegal)
 


svenwest:
Verification build (pending) here:
https://teamcity01-ir.devtools.intel.com/buildConfiguration/SimicsCiSimics_MonoTargets_Verification/9537286

lib/1.4/utility.dml
```diff
@@ -192,15 +192,50 @@ template _init_val_power_on_reset is (init_val, power_on_reset) {
...
+        }
+    }
+}
+
 // Instantiate this on top level to support power-on reset.
 template poreset is power_on_reset {
-    port POWER {
+    port POWER is state_signal {
         implement signal {
             method signal_raise() {
```
makaro1x, line 235:
```suggestion
            method signal_raise() default {
```

Without this `default` `signal_raise` can not actually be overridden.

lib/1.4/utility.dml
```diff
@@ -234,13 +269,12 @@ template _init_val_hard_reset is (init_val, hard_reset) {
 
 // Instantiate this on top level to support hard reset.
 template hreset is hard_reset {
-    port HRESET {
+    port HRESET is state_signal {
         implement signal {
             method signal_raise() {
```
makaro1x, line 274:
```suggestion
            method signal_raise() default {
```

lib/1.4/utility.dml
```diff
@@ -301,20 +335,19 @@ for power-on reset, hard reset and soft reset, respectively.
...
 
 </section>
 */
 
 // Instantiate this on top level to support soft reset.
 template sreset is soft_reset {
-    port SRESET {
+    port SRESET is state_signal {
         implement signal {
             method signal_raise() {
```
makaro1x, line 347:
```suggestion
            method signal_raise() default {
```

svenwest: Agreed. Fixed and updated documentation of hreset, sreset and poreset templates.

svenwest:
New verification build
https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=9537687

test/1.4/lib/T_state_signal.py
```diff
@@ -0,0 +1,58 @@
...
+stest.expect_equal(obj.on_raise_called, True)
+stest.expect_equal(obj.on_lower_called, False)
+
+obj.on_raise_called = False
+obj.on_lower_called = False
+
+obj.port.up.iface.signal.signal_lower()
+stest.expect_equal(obj.on_raise_called, False)
+stest.expect_equal(obj.on_lower_called, False)
+
```
JonatanWaern, line 47: Nitpickery: as you demonstrate, you can skip resetting the variables after the expect_equal confirms an off-value.

test/1.4/lib/T_state_signal.dml
```diff
@@ -0,0 +1,51 @@
...
+  © 2021 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+dml 1.4;
+
+device test;
+
+import "utility.dml";
+
+saved bool on_raise_called;
```
JonatanWaern, line 11: Technically, we dont want to use saved variables for interacting with the model since the name and representation of saved variable attributes are not guaranteed. I would prefer to declare proper attributes for these

svenwest: Ok, I'm not 100% happy with the fact - but I'll change this and raise the issue somewhere else.

mandolaerik: @JonatanWaern, I disagree: We may not document how saved variables map on attributes, but *we* know the mapping, and it will never change for simple variables, and *if* we change it, then updating the tests is the least of our problems (customer-side checkpoint compatibility is a bigger issue)

svenwest:
> it's hard to implement `signal` correctly;

Not really, but ok - I've added a template for a signal connect too.

lib/1.4/utility.dml
```diff
@@ -192,15 +192,105 @@ template _init_val_power_on_reset is (init_val, power_on_reset) {
...
+</section>
+*/
+template signal_connect is connect {
+    interface signal {
+        saved bool high;
+    }
+
+    method set(conf_object_t *obj) default {
+        /* this is only called if the attribute actually changes */
+        default(obj);
```
mandolaerik, line 264: do we also need to lower the signal to the previous obj?

lib/1.4/utility.dml
```diff
@@ -192,15 +192,105 @@ template _init_val_power_on_reset is (init_val, power_on_reset) {
...
+        if (signal.val && signal.high)
+            signal.signal_raise();
```
mandolaerik, line 266: If instantiating from scratch, this must be delayed (finalization phase or whatever that corresponds to with the new class functions).  If restoring a checkpoint, should not re-raise.

lib/1.4/utility.dml
```diff
@@ -296,25 +385,27 @@ template _init_val_soft_reset is (init_val, soft_reset) {
...
+Implemented on the top level to get standard reset behaviour, for power-on
+reset, hard reset and soft reset, respectively. The `signal_interface` methods
+`signal_raise` and `signal_lower` can be overridden to add additional
+side-effects. The default implementation calls the apropriate reset method on
+`signal_raise` and does nothing on `signal_lower`.
 
 #### Related Templates
 
-power\_on\_reset, hard\_reset, soft\_reset
+state\_signal, power\_on\_reset, hard\_reset, soft\_reset
```
mandolaerik, line 396: Lists of templates should be linkified (not part of this PR)

mandolaerik: Also, what's state_signal?

lib/1.4/utility.dml
```diff
@@ -192,15 +192,105 @@ template _init_val_power_on_reset is (init_val, power_on_reset) {
...
+    }
+
+    method set(conf_object_t *obj) default {
+        /* this is only called if the attribute actually changes */
+        default(obj);
+        if (signal.val && signal.high)
+            signal.signal_raise();
+    }
+
+    method raise() default {
```
mandolaerik, line 269: Since these are idempotent, perhaps they should be called drive_high/drive_low or `set_level(uint1 high)` or similar?

lib/1.4/utility.dml
```diff
@@ -192,15 +192,105 @@ template _init_val_power_on_reset is (init_val, power_on_reset) {
...
+        if (!signal.high) {
+            signal.high = true;
+            if (signal.val)
+                signal.signal_raise();
```
mandolaerik, line 273: in some cases it can also be problematic to signal something upon connect, e.g. if connect is modified in flight for some hybrid substitution. But it's what the interface dictates, so I suppose we have no choice.

lib/1.4/utility.dml
mandolaerik: More fun: There is no documented guarantee on the attribute setting order between the `signal_high` attribute and this connect on instantiation; in principle this means that the value of `signal.high` is not reliable.

However, to be fair this code is part of the DML distribution, so you can safely rely on the undocumented attribute order, which is alphabetical, so indeed signal.high will be false. This deserves more comments and testing.

lib/1.4/utility.dml
```diff
@@ -192,15 +192,105 @@ template _init_val_power_on_reset is (init_val, power_on_reset) {
...
+            else
+                high = true;
+        }
+        method signal_lower() default {
+            if (!high)
+                log error: "%s lowered when already low", qname;
+            else
+                high = false;
+        }
+    }
```
mandolaerik, line 228: As seen in your original connect code below, also experienced modelers can construct a signal connect that calls raise or lower early during checkpoint restore. If this happens before the `high` attribute is restored, then the tattle logic will be broken.

I can't immediately see any realistic cases where this would lead to false errors being reported; however, if a connect (incorrectly) raises the signal upon checkpoint restore, then the (righteous) error message will happen or not depending on which object gets it attributes restored first, and nondeterministic errors are unfriendly.

svenwest: Indeed, if hot-plugging must be supported it becomes quite tricky.
Considering that hot-plugging is quite rare, especially for `signal` we could make the template not support it.
On the other hand, since the template becomes a bit pointless without it, since the basic use case without hot plugging is almost trivial.

syssimics:
Verification [#9894846](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9894846): fail

syssimics:
Verification [#9894898](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9894898): fail

syssimics:
Verification [#9895901](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9895901): pass

gwikland:
With this change cmn-600.dml could be rewritten right?

Code now looks like:
`is hard_reset;
in each init_val {
    is _init_val_hard_reset;
}
method hard_reset() {
    default();
    memory_space.map_dummy_target();
}

is soft_reset;
in each init_val {
    is _init_val_soft_reset;
}
method soft_reset() {
    default();
    memory_space.map_dummy_target();
}

port HRESET {
    implement signal {
        method signal_raise() {
            reset_asserted = true;
            hard_reset();
        }
        method signal_lower() {
            reset_asserted = false;
            memory_space.map_default_target();
        }
    }
}`

Can now be rewritten as?

`is hreset;
method hard_reset() {
    default();
    memory_space.map_dummy_target();
}

is sreset;
method soft_reset() {
    default();
    memory_space.map_dummy_target();
}

port HRESET {
    implement signal {
       method signal_lower() {
            default();
            memory_space.map_default_target();
        }
    }
}`

svenwest:
> With this change cmn-600.dml could be rewritten right?

Yes, that and a lot of other code that doesn't treat xRESET as a single pulse.

svenwest:
Ping @mandolaerik

lib/1.4/utility.dml
```diff
@@ -197,15 +197,111 @@ template _init_val_power_on_reset is (init_val, power_on_reset) {
...
+the `signal_lower` method will be called on the old object, and the
+`signal_raise` method will be called on the new object. Similarly, if the
+device is created with `signal.high` set to `true`, the `signal_raise` method
+will be called on the connected object in the finalize phase. This behaviour
+can be changed by overriding the `set` method and/or the `post_init`
+method. The template defines the following method:
+
+* `set_level(uint1 high)`: Sets the level of the signal, by calling
+   `signal_raise` or `signal_lower`, as required. Also sets `signal.high`
+   to `high`
```
mandolaerik, line 256: missing `.`

test/1.4/lib/T_signal_templates.py
```diff
@@ -0,0 +1,32 @@
+# © 2022 Intel Corporation
+# SPDX-License-Identifier: MPL-2.0
+
+import stest
+import simics
+simics.SIM_run_command('log-level 2')
+obj.port.insig.iface.signal.signal_raise()
+stest.expect_equal(obj.port.insig.signal_high, True)
+with stest.expect_log_mgr(log_type="spec-viol", obj=obj.port.insig, regex="already high"):
+    obj.port.insig.iface.signal.signal_raise()
```
mandolaerik, line 10: if you want to test all aspects of init/SIM_is_restoring_state logic, then you can consider saying `/// INSTANTIATE-MANUALLY` in the DML file and do `SIM_create_object` here.

lib/1.4/utility.dml
mandolaerik: Still need to test (and comment) the attribute order invariant, i.e. `SIM_create_object(..., [["foo", stub_obj], ["foo_signal_high", True]])` raises the signal.
Looking again at the code, it seems that the init order between attribute/connect/register objects is indeed alphabetical, but looks like all `saved` attributes are initialized after all attributes/connects/registers. Which will cause the test to fail. Maybe it's OK to change init order so `saved` comes first, and rely on that.

svenwest: This is in `post_init` so all attributes, connects, registers, saved's has been set. There's no dependence on order.
```
dml 1.4;
device meh;
import "utility.dml";
connect foo is signal_connect;
```
```
simics> @stub_obj = SIM_create_object('signal-bus', 'signal_bus', [])
simics> @meh = SIM_create_object('meh', 'meh', [['foo', stub_obj], ['foo_signal_high', True]])
simics> @stub_obj.level
1
```

mandolaerik: Right! sorry for the noise.

lib/1.4/utility.dml
svenwest: Done.

test/1.4/lib/T_signal_templates.py
svenwest: Thanks for the tip, added tests for init and checkpoint load.

syssimics:
Verification [#9926648](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9926648): pass

test/1.4/lib/T_signal_templates.py
```diff
@@ -0,0 +1,84 @@
...
+obj = simics.pre_conf_object('test', 'test',
+                             outsig=stub,
+                             outsig_signal_high=True)
+simics.SIM_add_configuration([stub, obj], None)
+stub = simics.SIM_get_object(stub.name)
+obj = simics.SIM_get_object(obj.name)
+stest.expect_equal(stub.level, 1)
+
+# loading a checkpoint doesn't call raise
+with tempfile.TemporaryDirectory() as d:
```
mandolaerik, line 34: IIRC, the TemporaryDirectory will be deleted when you exit the block; better for debug to create a dir under scratchdir that survives the test run.

svenwest: stest.scratch didn't seem reliable when I did `make test-dmlc`, it tried to create the checkpoint in `./scratch/checkpoint` which didn't exist.

Wherever stest.scratch decides to put files is a mystery to me, finding it after a failed test would take significantly longer for me than to rerun the test locally after modifying the test to use a hardcoded path for the checkpoint.

lib/1.4/utility.dml
```diff
@@ -197,15 +197,111 @@ template _init_val_power_on_reset is (init_val, power_on_reset) {
...
+    }
+
+    method post_init() default {
+        if (!SIM_is_restoring_state(dev.obj) && signal.val && signal.high)
+            signal.signal_raise();
+    }
+
+    method set(conf_object_t *obj) default {
+        // this method is only called if the attribute actually changes
+        local bool hotplug = SIM_object_is_configured(dev.obj);
```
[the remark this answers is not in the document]
svenwest: It won't be called if the attribute value didn't change, which it shouldn't do on reverse-execution, or if it does then it _is_ a hotplug.

lib/1.4/utility.dml
svenwest: objects are usually not set as configured until _after_ `post_init`

test/1.4/lib/T_signal_templates.py
mandolaerik: Grepping, it seems that there's a global variable `scratchdir` that you can access from `T_*.py`. Some other tests do that.

lib/1.4/utility.dml
mandolaerik: right

test/1.4/lib/T_signal_templates.py
svenwest: Sure, if that's what you want. I still won't know where `scratchdir` is so it makes no huge difference to me :-)

lib/1.4/utility.dml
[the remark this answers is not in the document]
svenwest: Got it, will do.

syssimics:
Verification [#9927540](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9927540): fail

lib/1.4/utility.dml
```diff
@@ -239,13 +345,12 @@ template _init_val_hard_reset is (init_val, hard_reset) {
 
 // Instantiate this on top level to support hard reset.
 template hreset is hard_reset {
-    port HRESET {
+    port HRESET is signal_port {
```
mandolaerik, line 348: This is an incompatible change, it makes HRESET less forgiving which probably will give massive failures in SSM code.

I suggest that you break out the reset parts to a separate PR and merge the new templates only, and make sure to run VP pretests before merging the reset template changes.

mandolaerik: Agreed, but our users are also extra sensitive to incompatibilities these days so better to play safe, and no need to delay the more urgent stuff waiting for pretests.

svenwest: Updating the reset-templates is the urgent part of this PR though, right now they're often unusable since the raise/lower methods of the signal interface can't be overridden.

mandolaerik: I suppose adding that `default` KW is orthogonal from adding the template, though. Switching behaviour to signal_port breaks compatibility just as much regardless.

svenwest: Yes it is, I'm just reminiscing about where all this started ;-)

mandolaerik (approved):

syssimics:
Verification [#9929877](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9929877): fail

PR #6, opened 2021-11-29
Make attributes internal if missing _both_ doc & desc

svenwest:


svenwest:
Continuation of
https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/983

svenwest:
Test L DML run here:
https://teamcity01-ir.devtools.intel.com/buildConfiguration/SimicsCiSimics_MonoTests_TestLDml/9535687

lib/1.4/dml-builtins.dml
```diff
@@ -539,7 +539,8 @@ template _conf_attribute {
 
     param configuration             default "optional";
     param persistent                default false;
-    param internal                  default !defined documentation;
+    param internal                  default (!defined documentation &&
+                                             !defined desc);
```
lwaern-intel, line 543: The style requires the binary operator after the line break, not before. Parentheses are optional, but seemingly encouraged, so you can keep those if you want to.

test/1.4/lib/T_internal.dml
```diff
@@ -0,0 +1,49 @@
...
+attribute a2 is bool_attr {
+    param documentation = "long desc";
+}
+attribute a3 is bool_attr {
+    param internal = false;
+}
+
+method get_internal(const char *cls, const char *attr) -> (bool) {
+    return (SIM_get_attribute_attributes(SIM_get_class(cls), attr)
+            & Sim_Attr_Internal) == Sim_Attr_Internal;
```
lwaern-intel, line 35: You can use `SIM_object_class` together with the `obj` param of the device object or the banks instead. I think that's cleaner.

lwaern-intel (changes requested):

svenwest:
Verification build is here:
https://teamcity01-ir.devtools.intel.com/buildConfiguration/SimicsCiSimics_MonoTargets_Verification/9537095

JonatanWaern (approved):

lib/1.4/dml-builtins.dml
mandolaerik: I wholeheartedly agree! I have constantly been grudging internally about this ever since I joined in 2007, along with the c-style's indent level and `\n{`. I got used to it, and I always help reminding others to also obey, but I will never get used to it. A nice style is better than an ugly style, but an ugly style is better than an inconsistent mix of a nice style and an ugly style.

PR #7, opened 2021-11-29
Improve generation of array de/serialization functions -- SIMICS-18440

lwaern-intel:
Due to how serializable types work, we can't entirely avoid creating dedicated de/serialization functions for each array type which is implicitly serialized in the program -- however, these functions now simply serve as thin wrappers over calls to _deserialize_array and _serialize_array, which this PR introduces into dmllib.h.

We've discussed revamping the architecture of serialize.py to permit returning set_error_t in the future -- perhaps we could make use of that opportunity to also revamp de/serialization function generation to avoid the generation of these thin wrappers.

### DIFF
--- a/RELEASENOTES.docu
+++ b/RELEASENOTES.docu
@@ -1681,4 +1681,8 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
   <build-id value="6132"><add-note> Unicode BiDi control characters
       are no longer permitted in DML source files, for security
       reasons. </add-note></build-id>
+  <build-id value="next"><add-note> Addressed code generation efficiency issues
+      when using array types in <tt>saved</tt> variables or parameters to
+      methods called with <tt>after</tt> <bug number="SIMICS-18440"/>.
+  </add-note></build-id>
 </rn>
--- a/RELEASENOTES.docu
+++ b/RELEASENOTES.docu
@@ -1684,5 +1684,7 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
   <build-id value="next"><add-note> Addressed code generation efficiency issues
       when using array types in <tt>saved</tt> variables or parameters to
       methods called with <tt>after</tt> <bug number="SIMICS-18440"/>.
-  </add-note></build-id>
+
+      Serialized representation of <tt>uint8[(be|le)_t]</tt> array types has
+      been changed. </add-note></build-id>
 </rn>
--- a/RELEASENOTES.docu
+++ b/RELEASENOTES.docu
@@ -1686,5 +1686,7 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
       methods called with <tt>after</tt> <bug number="SIMICS-18440"/>.
 
       Serialized representation of <tt>uint8[(be|le)_t]</tt> array types has
-      been changed. </add-note></build-id>
+      been changed. The old serialized representation is still accepted for
+      deserialization in order to preserve checkpoint compatibility.
+  </add-note></build-id>
 </rn>
--- a/include/simics/dmllib.h
+++ b/include/simics/dmllib.h
@@ -1834,33 +1834,35 @@ static set_error_t
 _deserialize_array_aux(attr_value_t val, uint8 *data, size_t elem_size,
                        const uint32 *dimsizes, uint32 dims,
                        size_t total_elem_count,
-                       _deserializer_t deserialize_elem) {
+                       _deserializer_t deserialize_elem,
+                       bool elems_are_bytes) {
     uint32 len = *dimsizes;
 
-    // Deserialize the final dimension as bytes if deserialize_elem is NULL
-    if (dims == 1 && !deserialize_elem) {
-        if (unlikely(!SIM_attr_is_data(val))) {
-            SIM_attribute_error("Invalid serialized representation of byte "
-                                "array: not represented as data");
-            return Sim_Set_Illegal_Type;
-        } else if (unlikely(SIM_attr_data_size(val) != len)) {
+    // Allow the final dimension to be represented as data if elems_are_bytes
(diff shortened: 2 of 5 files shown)


RELEASENOTES.docu
```diff
@@ -1681,4 +1681,12 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
   <build-id value="6132"><add-note> Unicode BiDi control characters
       are no longer permitted in DML source files, for security
       reasons. </add-note></build-id>
+  <build-id value="next"><add-note> Addressed code generation efficiency issues
+      when using array types in <tt>saved</tt> variables or parameters to
+      methods called with <tt>after</tt> <bug number="SIMICS-18440"/>.
+
+      Serialized representation of <tt>uint8[(be|le)_t]</tt> array types has
+      been changed. The old serialized representation is still accepted for
+      deserialization in order to preserve checkpoint compatibility.
```
lwaern-intel, line 1690: Is this paragraph warrented? Erik says it isn't necessary, but it's not harmful either.

include/simics/dmllib.h
```diff
@@ -1780,6 +1783,124 @@ _get_port_saved_variable(lang_void *saved_access, conf_object_t *_portobj,
...
+            return error;
+        }
+    }
+    return Sim_Set_Ok;
+}
+
+UNUSED static set_error_t
+_deserialize_array(attr_value_t in_val, void *out_data, size_t elem_size,
+                   const uint32 *dimsizes, uint32 dims,
+                   _deserializer_t deserialize_elem, bool elems_are_bytes) {
```
lwaern-intel, line 1885: There's a discrepancy between `_serialize_array` and `_deserialize_array` in that `elems_are_bytes` is signaled to `_serialize_array` by providing `NULL` as the `serialize_elem` function pointer. For `_deserialize_array`, it needs to be its own dedicated parameter, as in that case `_deserialize_array` must be flexible in whether it receives an integer list or data value as the final dimension, so it needs `deserialize_elem`.

Is this discrepancy ugly enough to change `_serialize_array` so that it also has a dedicated parameter for `elems_are_bytes`?

test/1.4/serialize/T_saved_declaration.dml
```diff
@@ -35,6 +35,27 @@ typedef struct {
...
+saved uint8 saved_byte_array_via_list[4];
+saved uint8_be_t saved_nested_byte_array_via_list[4][3];
+
+// The final dimension of int8 array types -- both implicit and explicit
+// endian -- is serialized as an integer list, and can be deserialized either
+// through data or an integer list
+saved char saved_signed_byte_array_via_data[4];
+saved int8_be_t saved_nested_signed_byte_array_via_data[4][3];
+saved char saved_signed_byte_array_via_list[4];
+saved int8_be_t saved_nested_signed_byte_array_via_list[4][3];
```
lwaern-intel, line 58: Lots of test boilerplate. Should these be cut down?

RELEASENOTES.docu
JonatanWaern: I don't think its necessary. Mentioning that the old rep is still valid but not what it is seems weird, and I dont want us to mention what the serialization type is (since we don't guarantee it).

include/simics/dmllib.h
JonatanWaern: I think its fine.

test/1.4/serialize/T_saved_declaration.dml
lwaern-intel: Conclusion is: it's fine.

RELEASENOTES.docu
lwaern-intel: Conclusion is: doesn't matter.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9548892

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9549508

include/simics/dmllib.h
```diff
@@ -1785,6 +1788,124 @@ UNUSED static uint64 _identity_to_key(_identity_t id) {
...
+        if (unlikely(SIM_attr_data_size(val) != len)) {
+            SIM_c_attribute_error(
+                "Invalid serialized value of byte array: expected data of %u "
+                "bytes, got %u", len, SIM_attr_data_size(val));
+            return Sim_Set_Illegal_Value;
+        }
+        memcpy(data, SIM_attr_data(val), len);
+        return Sim_Set_Ok;
+    } else if (unlikely(!SIM_attr_is_list(val))) {
+        if (data_allowed) {
```
mandolaerik, line 1853: I think this code is unreachable for attributes, since the declared attribute type captures the error earlier. But I suppose it's still useful for `after` event data. What will be the effect of SIM_attribute_error in this case?

lwaern-intel: If `SIM_attribute_error` is called during deserialization of event data Simics will recognize it and checkpoint restoration will fail (and the error will be printed).

py/dml/serialize.py
```diff
@@ -62,6 +68,21 @@ def declare_variable(site, name, type, init_expr = None):
     return (ctree.mkDeclaration(site, name, type, init = init),
             ctree.mkLocalVariable(site, symtab.LocalSymbol(name, name, type)))
 
+def prepare_array_de_serialization(site, t):
```
JonatanWaern, line 71: nitpick: "de_serialization" is incorrect

JonatanWaern (approved):

py/dml/serialize.py
lwaern-intel: This was actually intentional, meant to reflect the slash in `de/serialization` -- but yeah, it's questionable. I'd welcome a better suggestion.

PR #8, opened 2021-11-29
Fix register_view.description of multi-dimensional banks

JonatanWaern:


lib/1.4/dml-builtins.dml
```diff
@@ -1359,28 +1359,43 @@ template bank is (object, shown_desc, _qname) {
...
+            local uint64 num = 0;
+            foreach r in (this._each_register) {
+                if (r.offset != unmapped_offset) {
+                    _reginfo_cache[num] = r;
+                    num++;
                 }
                 num++;
             }
         }
         throw;
```
lwaern-intel, line 1384: This `throw` statement doesn't belong.

lwaern-intel (changes requested):
Looks sane enough except for a stray `throw`.

lib/1.4/dml-builtins.dml
JonatanWaern: fixt. mustve snuck in while migrating patches

JonatanWaern:
Accidentally force-pushed a non-rebased version, which seems to break the pull request unrecoverably (forcepushing new, rebased, version doesnt fix). Will dupe to new pull request

PR #15, opened 2021-12-01
Make subsequent_log use trait identity as key

JonatanWaern:


py/dml/codegen.py
```diff
@@ -1573,10 +1573,19 @@ def stmt_log(stmt, location, scope):
...
             key = mkLit(site, '(uint64)NULL', TInt(64, False))
         else:
-            # Normally, we are not allowed to cast "this.trait" to an uint,
-            # but since we _internally_ know it is a pointer this is justified
-            key = mkLit(site, "(uint64)(%s.trait)" % lookup_var(
-                site, scope, "this").read(), TInt(64, False))
+            # Acquire a key based on trait _identity_t
+            identity = TraitObjIdentity(site, lookup_var(site, scope, "this"))
+            # identity is opaque, so we need to cheat a little bit
+            top_bits = mkShL(site,
```
mandolaerik, line 1579: Maybe the test should also cover the array logic? i.e. extend g to `g[i<2]` and check that it will log once on `[0]` and once on `[1]`.

py/dml/codegen.py
```diff
@@ -1573,10 +1573,19 @@ def stmt_log(stmt, location, scope):
...
+                             mkLit(site,
+                                   "%s.id" % identity.read(),
+                                   TInt(32, False)),
+                             mkIntegerConstant(site, 32, False))
+            bottom_bits = mkLit(site,
+                                "%s.encoded_index" % identity.read(),
+                                TInt(32, False))
+            key = mkAdd(site, top_bits, bottom_bits)
```
mandolaerik, line 1587: I would use the equivalent `mkOr`, but it's a matter of taste

py/dml/codegen.py
JonatanWaern: reasonable. will do

JonatanWaern: Done

py/dml/codegen.py
```diff
@@ -1573,10 +1573,19 @@ def stmt_log(stmt, location, scope):
         if location.method():
             key = mkLit(site, '(uint64)NULL', TInt(64, False))
         else:
-            # Normally, we are not allowed to cast "this.trait" to an uint,
-            # but since we _internally_ know it is a pointer this is justified
-            key = mkLit(site, "(uint64)(%s.trait)" % lookup_var(
-                site, scope, "this").read(), TInt(64, False))
+            # Acquire a key based on trait _identity_t
+            identity = TraitObjIdentity(site, lookup_var(site, scope, "this"))
+            # identity is opaque, so we need to cheat a little bit
```
lwaern-intel, line 1578: An alternative is to do the necessary manipulation through a method in `dmllib.h` instead.

RELEASENOTES-1.4.docu
```diff
@@ -237,4 +237,8 @@
       The internal representation of values of template types has been changed.
       It is no longer allowed to cast such values to integer or pointer types.
   </add-note></build-id>
+  <build-id value="next"><add-note> Fixed an issue where a log
+      statement could fail to switch log levels when declared within
```
lwaern-intel, line 241: "used" is probably the better word than "declared" here.

py/dml/codegen.py
JonatanWaern: This seems reasonable. It puts the conversion function, which depends on representation defined in dmllib, in the same place

include/simics/dmllib.h
```diff
@@ -1780,6 +1780,11 @@ _get_port_saved_variable(lang_void *saved_access, conf_object_t *_portobj,
                                   acc->getter);
 }
 
+
+UNUSED static uint64 _identity_to_key(_identity_t id) {
+        return id.id << 32 || id.encoded_index;
```
lwaern-intel, line 1785: `|` not `||`. I also suspect the implicit types involved are incorrect: that should be fixed by this:
```
return (uint64)id.id << 32 | id.encoded_index;
```

py/dml/codegen.py
```diff
@@ -1573,10 +1573,13 @@ def stmt_log(stmt, location, scope):
         if location.method():
             key = mkLit(site, '(uint64)NULL', TInt(64, False))
```
lwaern-intel, line 1574: This can be modified to instead be based on the indices provided to the method call, which will allow objects of an array to be independent of one another. You can leverage `encode_indices` from ctree for this purpose.

lwaern-intel (changes requested):

include/simics/dmllib.h
JonatanWaern: Yeah I found this out too, felt pretty stupid.

py/dml/codegen.py
JonatanWaern: I think my current implementation just ends up using _identity_to_key again, just with an ObjIdentity instead.

syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9542071

syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9542206

JonatanWaern:
Testfailure seems unrelated

py/dml/codegen.py
lwaern-intel: I'm slightly skeptical about using full-fledged identities in the non-shared case -- I feel like that would cause the outermost hash table to become unnecessarily big.  Using `ObjIdentity` is certainly easier, but I feel only basing the key on indices (in this case) would be somewhat more morally correct.

mandolaerik: I vote for ObjIdentity if it makes code cleaner; we don't expect this to be anywhere close to expensive in practice. Showing a message where a human can see it costs magnitudes more than a few bytes of memory, so as long as table is only populated when messages are logged, the relative cost is negligible.

lwaern-intel:
Test is green after re-run.

lwaern-intel (approved):

PR #17, opened 2021-12-02
Add support for cancelling events posted via after -- SIMICS-17930

lwaern-intel:
This PR is _extremely heavy_ on `dmllib.h` as it needs a lot of convoluted logic to be able to both handle the old representation format (for checkpoint compatibility) and the new pseudodict format.


lib/1.4/dml-builtins.dml
```diff
@@ -375,13 +378,50 @@ template post_init {
...
+those associated with any subobject.
+*/
+template after_domain {
+    shared method remove_after() {
+        _cancel_simple_events(dev.obj, cast(this, _identity_t));
+    }
+
+    shared method posted_after() -> (bool) {
+        return _poll_simple_events(dev.obj, cast(this, _identity_t));
+    }
```
lwaern-intel, line 415: How's the naming of these methods? They currently mirror the methods of event objects, but I'd prefer `cancel_after()` instead of `remove_after()`. There's also the question if `posted_after()` should even be offered.

py/dml/c_backend.py
```diff
@@ -917,120 +925,145 @@ def generate_implements(code, device):
...
+    out('}\n', preindent=-1)
+    out('return out;\n')
+    out('}\n', preindent=-1)
+    splitting_point()
 
 def generate_simple_events(device):
     handled = set()
     done = False
 
+    # TODO PURGE
```
lwaern-intel, line 956: Refers to the quadratic nonsense with `handled` and `done`. Forgot doing this before I published the PR.

test/1.4/events/T_after_control.dml
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    method post_multi_operator_modify_unshared(
+        bool use_cycles, double t, int i, int j, int v) {
+        if (use_cycles) {
+            after cast(t, uint64) cycles: multi_operator[i][j].modify(v);
+        } else {
+            after t s: multi_operator[i][j].modify(v);
+        }
+    }
```
lwaern-intel, line 78: A lot of dumb repetition. I can't find a way around it while still keeping the `after` statements associated with the object instantiating the template.

include/simics/dmllib.h
```diff
@@ -599,6 +599,422 @@ _identity_eq(const _identity_t a, const _identity_t b) {
...
+        *out_domains = temp_out_domains;
+    } else {
+        *out_no_domains = 0;
+        *out_domains = NULL;
+    }
+    return Sim_Set_Ok;
+}
+
+UNUSED static set_error_t
+_deserialize_simple_event_data(
```
lwaern-intel, line 933: It's kind of hilarious how about 50% of the entire PR is dedicated to the code necessary to implement this function.

py/dml/c_backend.py
```diff
@@ -917,120 +925,145 @@ def generate_implements(code, device):
...
-        out(f'ASSERT(SIM_attr_list_size({data}) == 2);\n')
-        out(f'attr_value_t indices = SIM_attr_list_item({data}, 0);\n')
-        out(f'attr_value_t params = SIM_attr_list_item({data}, 1);\n')
-    elif method.dimensions > 0 or len(method.inp) > 0:
-        varname = 'indices' if method.dimensions > 0 else 'params'
-        out(f'attr_value_t {varname} = {data};\n')
-
-    if method.dimensions > 0:
-        out(f'ASSERT(SIM_attr_list_size(indices) == {method.dimensions});\n')
+def generate_simple_event_only_domains_funs():
```
lwaern-intel, line 928: The only reason these functions get generated by DMLC instead of living in `dmllib.h` is because they reference `_id_infos` and `_id_info_ht`, which `dmllib.h` do not have direct access to.

test/1.2/events/T_after_chk.dml
```diff
@@ -34,9 +42,9 @@ attribute alttest {
...
     method set(value) {
         log "info": "running alttest";
-        after (0.1) call $m1;
-        after (0.1) call $p[1][1].m1;
-        after (0.1) call $m2(1);
-        after (0.1) call $p[1][1].m2(4);
+        call $post_m1;
+        call $p[1][1].post_m1;
+        call $post_m2(1);
+        call $p[1][1].post_m2(4);
```
lwaern-intel, line 48: I move these after calls to a method of the template to more thoroughly test checkpointing of domains (with after calls directly in `set` they would all be associated with `alttest`).

py/dml/c_backend.py
```diff
@@ -917,120 +925,145 @@ def generate_implements(code, device):
...
+                out('if (error != Sim_Set_Ok) {\n', postindent=1)
+                out('out = NULL;\n')
+                out('}\n', preindent=-1)
+                out('return (lang_void *) out;\n')
                 out('}\n\n', preindent = -1)
                 splitting_point()
 
     reset_line_directive()
 
+def generate_simple_events_control_methods(device):
```
lwaern-intel, line 1044: This doesn't need the `device` parameter. Will remove.

syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9540757

doc/1.4/language.md
```diff
@@ -2823,6 +2823,20 @@ where each input parameter is of [*serializable type*](#saved-variables).
...
+that the method using the statement belongs to. It is possible to cancel or
+poll all `after` events associated with an object through that object's
+`remove_after()` or `posted_after()` method, as provided by the
+[`after_domain` template](dml-builtins.html#after_domain).
+
+<div class="note">
+
+**Note:** We plan to extend the `after` statement to allow for users to
+explicitly state what objects that events posted from the use of that statement
+are to be associated with.
```
lwaern-intel, line 2836: The formulation here is a bit awkward. Does anyone have a better suggestion?

lib/1.4/dml-builtins.dml
lwaern-intel: Also, how's the naming of the template itself? I don't dislike `after_domain` per se, but it kind of relies on leap of logic to understand why it is named like that (that objects instantiating the template are -- in a sense -- a domain for events posted via `after`). If we drop `posted_after()` we can simply name the template `removable_after` or `cancellable_after`, but if we have both methods that naming isn't appropriate.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9549468

doc/1.4/language.md
mandolaerik: Empty string? We have no obligation to advertise not-yet-implemented features. We probably won't add this until someone asks for it.

lwaern-intel: Don't you remember the debate? Leonid has already asked for it. The only reason we decided the current design was acceptable was _because_ we would support explicitly associating objects with after statements.

lib/1.4/dml-builtins.dml
mandolaerik: why not inline the template into `object`? Other supertemplates of object are motivated because it's common for shared methods to need its methods, and `object` is a big template to multiple inherit. I don't think this is a strong use case. Making methods part of `object` is conservative, a later step of extracting it into a supertemplate is 100% compatible.

mandolaerik: I agree cancel_after is a better name.

Regarding existence of posted_after: We should add it iff we can prove that it adds value. A sufficient proof is to find a real use of `event.posted()` that is better expressed as an `after` (i.e., grep for `.posted(`)

py/dml/c_backend.py
mandolaerik: Maybe easier to read if you triple-quote the function bodies?

Indeed it's hard to expose `_id_infos` in dmllib.h without unconventional patterns that make it  cumbersome to include dmllib.h standalone (which we need to do in `ctree_test`)

py/dml/c_backend.py
```diff
@@ -917,120 +925,145 @@ def generate_implements(code, device):
...
-        out(f'ASSERT(SIM_attr_list_size(indices) == {method.dimensions});\n')
+def generate_simple_event_only_domains_funs():
+    start_function_definition(
+        'attr_value_t _simple_event_only_domains_get_value('
+        + 'conf_object_t *_obj, lang_void *data)')
+    out('{\n', postindent=1)
+    out('return _serialize_simple_event_data(0, NULL, _id_infos, '
+        + '(_simple_event_data_t *)data);\n')
+    out('}\n', preindent=-1)
+    splitting_point()
```
mandolaerik, line 936: `splitting_point()` is redundant here: it should be added after generating a function that can be large or plentiful in DML 1.2. The `--split-c-file` option is a workaround for the kind of performance problems you get in 1.2, and is largely irrelevant in 1.4.

py/dml/c_backend.py
```diff
@@ -917,120 +925,145 @@ def generate_implements(code, device):
...
+                # Use emergency all-zero indices if data is NULL
+                # (happens when de/serialization fails).
+                out('const uint32 *_indices = data ? data->indices '
+                    + (': (const uint32 [%s]) { 0 };\n'
+                       % (method.dimensions,)))
+
+            if len(method.inp) > 0:
+                # Use emergency all-zeroed params if data is NULL
+                args_struct_t = info['args_type'].label + '_t'
+                out('const %s *_args = data ? data->args : &(%s) { 0 };\n'
```
mandolaerik, line 987: why % instead of f-strings?

doc/1.4/language.md
mandolaerik: "explicitly state which object to associate with the event"?

You can ponder if this is best written here or in the release note.

lib/1.4/dml-builtins.dml
lwaern-intel: > why not inline the template into `object`? Other supertemplates of object are motivated because it's common for shared methods to need its methods, and `object` is a big template to multiple inherit. I don't think this is a strong use case. Making methods part of `object` is conservative, a later step of extracting it into a supertemplate is 100% compatible.

I don't see how making methods part of `object` is more conservative. Anyway, yes, I broke it into a template since I figured people would create templates inheriting it and a `reset` template, like this:

```
template hard_reset_after_domain is (hard_reset, after_domain) {
  shared method hard_reset() default {
    hard_reset_subobjs();
    cancel_after();
  }
}
```

py/dml/c_backend.py
lwaern-intel: Noted, will remove.

py/dml/c_backend.py
lwaern-intel: Note the braces.

py/dml/c_backend.py
```diff
@@ -917,120 +925,145 @@ def generate_implements(code, device):
...
-                    code = codegen_inline(site, method, (), [], [])
-                code = mkCompound(site, declarations(scope) + [code])
-                code.toc()
-                output_dml_state_change("_dev")
+            out('_simple_event_data_t *data = '
+                + '(_simple_event_data_t *)_data;\n')
+
+            if method.dimensions > 0:
+                # Use emergency all-zero indices if data is NULL
+                # (happens when de/serialization fails).
```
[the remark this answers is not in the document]
lwaern-intel: This code was written before I knew `SIM_attribute_error` causes checkpoint restoration to fail. Event data deserializers have no other way to indicate failure, so I figured "oh no, badly deserialized events are still added to the event queue"; but that's not the case. Still I _really_ don't like the fact that `SIM_attribute_error` is the only means to fail deserialization -- I don't get why the API was designed that way.

lib/1.4/dml-builtins.dml
mandolaerik: When I say it's more conservative, I mean that if we create a template today, then we cannot remove it later without breaking compatibility. While if we don't create it today, we can extract it later without breaking compatibility. (apart from the risk for a name clash)

lwaern-intel: Ok, I see.
I personally think we're gonna want a template for this, but there's little harm in inlining it into `object` for the time being. I'll get rid of `posted_after()` at the same time, and rename `remove_after()` to `cancel_after()`

mandolaerik: Which gives you the natural template name `cancel_after`, unless `posted_after` will resurrect.

If I ever get my perfect hash stuff done, then multiple inheritance will have a zero cost, so it's OK for `hard_reset_after_domain` to inherit `object`. Also, it could happen that we want to unify coroutine domains and after domains, which also can affect naming. So this is a question of how soon we need after_domain.

doc/1.4/language.md
lwaern-intel: I think it's best written here. I've tinkered the reformulation a bit -- I want to say "associate events with objects" rather than "associate objects with events", as the objects are the "owners" of the events, in some sense.

py/dml/c_backend.py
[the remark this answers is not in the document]
lwaern-intel: Updated so that if `data` somehow manages to be `NULL`, `VT_critical_error` is invoked.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9577367

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9577366

test/1.2/events/T_after_chk.dml
lwaern-intel: Instead of simply covering domain checkpointing logic, these tests have been moved to under 1.4 and actually test that domains are checkpointed properly.

This leaves us with the question, what do we do about `T_after_chk_dup`? Its existence was already questionable. Do we remove it?

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9579330

py/dml/c_backend.py
```diff
@@ -917,120 +925,125 @@ def generate_implements(code, device):
...
 
+def generate_simple_events_control_methods():
+    start_function_definition(
+        'void _cancel_simple_events(conf_object_t *_obj, _identity_t domain)')
+    out('{\n', postindent = 1)
+    for (method, _) in list(simple_events.items()):
+        out('SIM_event_cancel_time('
+            + f'SIM_object_clock(_obj), {crep.get_evclass(method)}, _obj, '
+            + '_simple_event_predicate, (lang_void *) &domain);\n')
+    out('}\n\n', preindent = -1)
```
lwaern-intel, line 1044: I'm concerned this is _slow_. `SIM_event_cancel_time` has a bunch of overhead and is linear in the size of the event queue (_not_ just the number of events posted of the event class), and it's called for every after event class. If a user calls `cancel_after()` on multiple objects at the same time (like they'd do in reset operations), then that operation is `O(n * k * o)`, where `n`= total number of posted events, `k` = total number of after event classes, `o` = number of objects. (This is assuming `_simple_event_predicate` is `O(1)`--- although it is linear in the number of domains associated with the tested after event, that number is expected to be very low, typically just 1.)

Thing is, I don't know how we could improve upon this without either breaking checkpoint compatibility (by changing how after event classes are generated to reduce the number of them) or changing the very nice interface we currently have (by dropping `cancel_after()` being per-object in favor of a global method that can accept a list of objects.)

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9579329

lwaern-intel:
@mandolaerik @JonatanWaern Reminder that I've requested your review.

doc/1.4/language.md
```diff
@@ -2823,6 +2823,19 @@ where each input parameter is of [*serializable type*](#saved-variables).
 This means that `after` statements cannot be used with methods
 that e.g. have pointer parameters.
 
+Any events posted via an `after` statement are associated with the object
```
mandolaerik, line 2826: emphasize `*associated*`?

doc/1.4/language.md
```diff
@@ -2823,6 +2823,19 @@ where each input parameter is of [*serializable type*](#saved-variables).
...
+that the method using the statement belongs to. It is possible to cancel
```
mandolaerik, line 2827: The "Any events..." sentence is hard to parse, but indeed correct. My eyes initially parsed "the statement belongs to" as a unit. Maybe insert ", i.e., the caller," somewhere?

lwaern-intel: I don't feel like "the caller" works. After statements are not methods/functions, and do not have a caller -- but more concerning, "the caller" could be confused with "the caller of _the method_", which is very incorrect.

I was considering:
```
Any events posted via an `after` statement are associated with the object
that contain the method using the statement. 
```
But I was worried that might've caused confusion with regards to methods defined in templates, especially `shared` methods.

doc/1.4/language.md
lwaern-intel: Sure.

doc/1.4/language.md
mandolaerik: I liked the 'contains' sentence better, I don't see how it changes the risk of confusion. I like 'contain' better than 'use'; I think the repetition in 'object that contains the method that contains the statement' is fine since the containment is indeed transitive.

include/simics/dmllib.h
```diff
@@ -599,6 +599,423 @@ _identity_eq(const _identity_t a, const _identity_t b) {
...
+} _simple_event_data_format_t;
+
+UNUSED static _simple_event_data_format_t
+_simple_event_data_representation(
+    attr_value_t val, bool has_indices, bool has_params) {
+    ASSERT(SIM_attr_is_list(val) && SIM_attr_list_size(val) > 0);
+    if (SIM_attr_list_size(val) == 2 && has_indices && has_params) {
+        // Either the pseudodict format, or the [indices, params] format
+        // Determine by inspecting the first element:
+        // [integer, ...] -> Legacy representation
```
mandolaerik, line 647: I think the two formats deserve a descriptive comment somewhere. In particular, specify the format of the legacy representation, and why it exists (compatibility with checkpoints older than 6.0.x). This way, it will be visible when it's clearly irrelevant and can be dropped.

py/dml/c_backend.py
mandolaerik: I suppose the only potentially large number here is `o`; someone might run cancel_after on *all* objects. I expect both n and k to be small, and resets happen seldom.

One possible optimization would be to keep a cache of objects that have posted simple events. The cache is invalidated when event data is destroyed or deserialized. When the cache is invalid, cancel_after can rebuild it by scanning the queue once with e.g. SIM_event_find_next_time.

include/simics/dmllib.h
mandolaerik: The functions all look nice and testable. It would be nice to have some unit testing; either as a pure C file in the project, or within a header/footer block in (say) `test/1.4/misc/T_unit_test_dmllib.dml`.

In particular, this could also act as a nice "documentation by example" for the contracts of these functions.

test/1.4/methods/T_idempotent.dml
```diff
@@ -0,0 +1,156 @@
+/*
```
lwaern-intel, line 1: I have _no idea_ how this got in. Oops.

syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9699700

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9699965

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9699964

include/simics/dmllib.h
```diff
@@ -631,6 +631,27 @@ _simple_event_predicate(lang_void *_data, lang_void *match_data) {
...
+ * The modern serialized representation of after event data is a
+ * pseudo-dictionary, containing entries for method indices, arguments, and
+ * domains (associated objects for cancellation):
+ * [["indices", [index0, index1, ...]],
+ *  ["arguments", [arg0, arg1, ...]],
+ *  ["domains", [domain0, domain1, ...]]
+ * ]
+ * The entries may appear in any order, and any entry may be omitted if not
+ * applicable, (for example, a method call has no indices/arguments or
+ * associated objects). If all entries may be omitted, then Nil is used
```
mandolaerik, line 644: `s/may be/are/` ?

I think lowercase `nil` is the convention.

include/simics/dmllib.h
```diff
@@ -631,6 +631,27 @@ _simple_event_predicate(lang_void *_data, lang_void *match_data) {
...
+ * as the serialized representation.
+ *
+ * Before the introduction of domains, after event data was represented as
+ * follows:
+ * [[index0, index1, ...], [arg0, arg1, ...]]
+ * and when either list is empty, the other is represented directly; if both
+ * are empty, Nil is used.
+ * This is called the legacy representation of after event data, and is
+ * supported for backwards checkpoint compatibility.
```
mandolaerik, line 653: When documenting compatibility luggage, it's convenient for future archeologists if you mention approximate time ("early 2022"). Makes it easy to deduce if it's sufficiently ancient to be silently removed.

test/1.4/misc/T_dmllib.dml
```diff
@@ -0,0 +1,450 @@
...
+extern bool dmllib_test(char *);
+
+method test() throws {
+    local char msg_buf[512];
+    if (!dmllib_test(msg_buf)) {
+        log error: "%s", msg_buf;
+    }
+}
+
+footer %{
```
mandolaerik, line 22: nicer for editors if you extract this to a separate .c or .h file that you `#include`

test/1.4/misc/T_dmllib.dml
```diff
@@ -0,0 +1,450 @@
...
+            bool has_params  = i & 2;
+            TEST_REPR(one_entry_dict, has_indices, has_params, Pseudodict,
+                      "single-entry, indices=%d, params=%d",
+                      has_indices, has_params);
+            TEST_REPR(two_entry_dict, has_indices, has_params, Pseudodict,
+                      "double-entry, indices=%d, params=%d",
+                      has_indices, has_params);
+        }
+
+        exit:
```
mandolaerik, line 100: no need for a clean return on failure, fine to fail by FATAL_ERROR_IF or similar. Saves you the trouble of passing exit labels to macros, and propagating error message to caller. Also gives you the convenience of error messages with clickable line numbers that bring you to the failed test.

include/simics/dmllib.h
lwaern-intel: I disagree with `s/may be/are/`; "may be" ties back to the previous sentence. "A dict is like this, but an entry _may be_ omitted if not applicable, and if every entry _may be_ omitted, then the dict is nil instead.`

You're right, lowercase `nil` it is.

test/1.4/misc/T_dmllib.dml
mandolaerik: Also note that with line numbers in error messages, the message is often redundant (exceptions include loop indices and when it replaces a comment).

syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9704378

test/1.4/misc/T_dmllib.dml
lwaern-intel: Now leveraging `ASSERT_MSG` and `ASSERT_FMT`.

test/1.4/misc/T_dmllib.dml
lwaern-intel: Done.

include/simics/dmllib.h
lwaern-intel: Done.

include/simics/dmllib.h
mandolaerik: Then "if no entry is applicable" would sound more direct?

Still, I find it a strange mix of conditional and certain stuff: If one single entry is unapplicable, then it *may be* omitted, while if all entries are unapplicable, then the dict *is* replaced with nil. I'd either say "is" on both or "may be" on both (I suppose the truth is "is" on serialization and "may be" on deserialization).

mandolaerik (approved):
minor language issue in comment, no need to fix if you disagree.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9704931

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9704930

PR #18, opened 2021-12-02
map_target template

svenwest:
Add map_target template

This template makes the connect represent a Simics `map_target`, which
accepts read and writes to a given address.  This is achieved by the
use of a `map_target_t *` and `SIM_issue_transaction`.

### DIFF
--- a/lib/1.4/utility.dml
+++ b/lib/1.4/utility.dml
@@ -6,6 +6,10 @@
 dml 1.4;
 
 import "simics/devs/signal.dml";
+import "simics/devs/ram.dml";
+import "simics/devs/memory-space.dml";
+import "simics/devs/translator.dml";
+import "simics/model-iface/transaction.dml";
 
 template _reg_or_field {
     param is_register : bool;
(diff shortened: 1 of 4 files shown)


svenwest:
Verification build:
https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=9540552

lib/1.4/dml-builtins.dml
```diff
@@ -997,6 +999,48 @@ template init_as_subobj is (connect, init, _qname) {
...
+
+A `connect` object can instantiate the template `map_target`, making it only
+accept objects that are valid map targets, i.e. objects for which
+`SIM_new_map_target` returns a valid pointer, without exception. The template
+provides a session variable `map_target` of type `map_target_t *` which can be
+used to issue transactions to the connected object. The template also defines a
+method `issue(transaction_t *t, uint64 addr) -> (exception_type_t)` which can
+be used as a shorthand for `SIM_issue_transaction`.
+*/
+template map_target is connect {
```
mandolaerik, line 1013: This template is highly Simics-specific, both in name and in signatures. I want to keep dml-builtins as simulator agnostic as possible.

I propose three things:
* move to utility.dml
* add alternative shorthand simulator-agnostic methods along the lines of `memory_space.{read,write}`
* Rename to something more simulator-agnostic, like `memory`.

I have no problem keeping the `issue` method; we can view it as a simulator-specific extension.

lib/1.4/dml-builtins.dml
```diff
@@ -997,6 +999,48 @@ template init_as_subobj is (connect, init, _qname) {
...
+    session map_target_t *map_target;
+
+    method validate(conf_object_t *obj) -> (bool) default {
```
mandolaerik, line 1016: Having validate/set declared `default` is in principle API. Maybe if it should be documented (not sure). I was going to ask for use cases, but I can see pretty clear use cases myself (like a connect that combines map target with some other interface).

lib/1.4/dml-builtins.dml
```diff
@@ -997,6 +999,48 @@ template init_as_subobj is (connect, init, _qname) {
...
+        local map_target_t *tmp = SIM_new_map_target(obj, port, NULL);
```
mandolaerik, line 1017: I prefer to require that `port` is NULL. We want to deprecate interface ports, and this is a good opportunity to steer users away from using it in new code.

lib/1.4/dml-builtins.dml
svenwest: > move to utility.dml

Sure, I put it in builtins because it felt similar to `init_as_subobj` and I wanted them documented together.

> alternative shorthand simulator-agnostic

Ok, that could probably be useful

> rename to something more simulator-agnostic

I want to keep it agnostic with respect to what it targets too, so I don't really like `memory` since it could just as well target a register-bank. I think map_target _could_ be simulator agnostic, we we define it as such.

MagnusIntel: Off-topic: 
Maybe init_as_subj should be moved to utility as well? It is not done since we need to be able send params to the subobj and that may become too Simics specific as well?

mandolaerik: I agree that init_as_subobj logically would belong to utility; the reason I chose dml-builtins is that I needed some compiler hacks to make it work (because we cannot yet call methods from module init). Things that are tightly coupled with the compiler count as builtins.

svenwest: Ok, so I'll put it in utility.dml, how do we get it documented together with `init_as_subobj`?

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9549859

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9555702

lib/1.4/utility.dml
```diff
@@ -1212,3 +1216,181 @@ template bank_obj is (bank, name) {
...
+        default(obj);
+        this.map_target = obj ? SIM_new_map_target(obj, NULL, NULL) : NULL;
+    }
+
+    /**
+       <dt><small>
+       read(uint64 addr, uint64 size) -&gt; (uint64) throws
+       </small></dt>
+       <dd>
+       Reads `size` bytes starting at `addr` in the connected object. Size
```
mandolaerik, line 1274: Did you look at the resulting docs? My understanding is that you are in HTML land now, so backticks don't work. A double newline is needed to enter markdown land again (at the cost of some extra vertical space in the output, unfortunately).

Since mixing md and plain html is so error-prone, it's often better to reformulate the docs so that pure markdown constructs are sufficient; in this case you might instead express it as a dotted list of methods, or as separate subsections.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9555932

lib/1.4/utility.dml
svenwest: I tried to follow the notation in `dml-builtins.dml`, wasn't aware of the need for double newline.
Documentation will be difficult to read if different sections in the same document use different formatting for the same logical concepts. Our markdown-strategy really feels awkward.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9556215

lib/1.4/utility.dml
svenwest: Fixed.

image

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9556347

lib/1.4/dml-builtins.dml
mandolaerik: Through hyperlinks.

lib/1.4/utility.dml
mandolaerik: Looks good! I agree it's awkward. Conradi has mentioned that it *can* be solved by turning dt into `### headings` and using CSS to make them look like definition lists, but that's outside of our control. Bullet lists is probably the best we can do today. Existing dl:s are leftovers from automatic migration.

mandolaerik (approved):

PR #25, opened 2021-12-07
Lost and found patches

mandolaerik:
- Improve logging test
- Mention that sizeoftype does not count as a constant
- Move topsort to module traits
- Refactor: Move method sorting to traits.py


mandolaerik:
I found these minor patches in my stacks during migration. They are not related to each other, do not add new features, and are expected to be uncontroversial

PR #26, opened 2021-12-07
Designated initializers (SIMICS-17252); non-constant compound initializers

lwaern-intel:
This is the migrated DML part of the corresponding PR over at Simics Base.


RELEASENOTES-1.4.docu
```diff
@@ -257,4 +257,8 @@
       now by default log one <tt>spec-violation</tt> message on log level 1
       on first such write, and no logging on further such writes.
   </add-note></build-id>
+  <build-id value="next"><add-note> Added simple C99-style designated
+      initializers. Use together with <tt>struct</tt>, <tt>extern struct</tt>,
+      <tt>layout</tt>, and <tt>bitfields</tt> types are supported
+      <bug number="SIMICS-17252"/>.</add-note></build-id>
```
JonatanWaern, line 263: What is the BKM for where to put the </ad-note></build-id> to avoid merge conflicts? last line of releasenote message or new line?

mandolaerik: The BKM is to *not* have it on a new line. Otherwise, after two simultaneous notes, a merge tool will propose that the identical closing tags should be merged into one.

So, it's a minor cost, and it's fine to have closing line on a separate line now and then (e.g. if that's what you get from auto-line-wrapping).

RELEASENOTES-1.4.docu
```diff
@@ -257,4 +257,8 @@
       now by default log one <tt>spec-violation</tt> message on log level 1
       on first such write, and no logging on further such writes.
   </add-note></build-id>
+  <build-id value="next"><add-note> Added simple C99-style designated
+      initializers. Use together with <tt>struct</tt>, <tt>extern struct</tt>,
```
mandolaerik, line 261: The release notes have refmanual-level precision, which hampers readability. Better to give a simple example of what you can do with it.

RELEASENOTES.docu
```diff
@@ -1689,4 +1689,7 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
       been changed. The old serialized representation is still accepted for
       deserialization in order to preserve checkpoint compatibility.
   </add-note></build-id>
+  <build-id value="next"><add-note> Added support for non-constant
+      compound initializers in contexts where both non-constant initializers
+      and compound initializers are allowed.</add-note></build-id>
```
mandolaerik, line 1694: Hard to read, in particular for someone who doesn't know exactly the previous limitations. Replace second half of sentence with a simple example, like `local buffer_t data = {size, new uint8[size]};`?

doc/1.4/language.md
```diff
@@ -1642,8 +1642,16 @@ In the absence of explicit initializer expression, a default
 Note that the number of elements in the initializer must match with
 the number of elements or fields of the type of the *session*
 variable. This also implies that each sub-element, if itself being a
-compound data structure, must also be enclosed in braces. C99-style
-designated initializers are not supported.
+compound data structure, must also be enclosed in braces.
+
+C99-style designated initializers are supported for `struct`, `layout`, and
```
mandolaerik, line 1647: Release note also mentions `extern struct`, but the difference is not significant to an end-user so makes sense to leave it out.

However, we *should* perhaps mention somewhere how struct initializers work for `extern` struct types, in particular when some fields are unknown to DML. My understanding is that such fields are initialized to 0.

doc/1.4/language.md
```diff
@@ -1680,8 +1688,16 @@ In the absence of explicit initializer expression, a default
...
+`bitfields` types:
+```
+typedef struct { int x; struct { int i; int j; } y; } struct_t;
+saved struct_t s = { .x = 1, .y = { .i = 2, .j = 3 } }
```
mandolaerik, line 1697: Documentation should also mention that initializers are available in assignment statements (not strictly needed for this patch, but related since you duplicate information).

py/dml/types.py
```diff
@@ -386,16 +392,17 @@ def bytes(self):
...
-                    typ.const = True
-                return typ, msb, lsb
-        return None, None, None
+        return ((name, conv_const(self.const, typ), msb, lsb)
+                for (name, (typ, msb, lsb)) in self.members.items())
+
+    def get_member_qualified(self, member):
+        t = self.members.get(member)
+        if t is not None:
+            t = (conv_const(self.const, t[0]),) + t[1:]
```
mandolaerik, line 404: nitpick: consider avoiding local mutation here, e.g. `return None if t is None else ...`

py/dml/ctree.py
```diff
@@ -4166,7 +4167,7 @@ def read(self):
     def assign_to(self, dest, typ):
         '''output C statements to assign an lvalue'''
         typ = safe_realtype(typ)
-        if isinstance(typ, TExternStruct):
+        if isinstance(typ, StructType):
```
mandolaerik, line 4170: I don't understand the implications of this change. Is it covered by a test?

py/dml/types.py
lwaern-intel: Pssh, ternary operators? How about _boolean operators_?:
```
return t and (conv_const(self.const, t[0]),) + t[1:]
```

py/dml/ctree.py
lwaern-intel: `DesignatedStructInitializer` was previously used only for extern structs, but I repurposed it for any struct type for the initializers codegen:d by the use of DML designated initializers.

RELEASENOTES.docu
lwaern-intel: I don't agree that this is hard to read, and I disagree with scrapping the latter half. However, I'm ok with adding an example.

RELEASENOTES-1.4.docu
lwaern-intel: True. Examples feel redundant, but I could say "struct-like" types instead of listing all the damn types.

doc/1.4/language.md
lwaern-intel: Correct. ISO standard specifies that when designated initializers are used, fields not initialized are initialized according to the rules for static variables, i.e. they become 0.

doc/1.4/language.md
lwaern-intel: In general, we need a section in the manual for assign statements, where we can state that initializers can be used as RHS of assigns; move stuff from the "Comparison to C/C++" section to that section; and move the explanations of the various forms of initializers throughout the manual to that section. This isn't appropriate to do in this PR, however. I'll create an issue about it.

doc/1.4/language.md
lwaern-intel: I can add a blurb explaining this to the section in the DML manual which explains extern typedefs.

py/dml/types.py
mandolaerik: Depends on how we think about non-None false values.

py/dml/ctree.py
mandolaerik: Does one of your tests cover this case?

RELEASENOTES.docu
mandolaerik: Few in the audience know by heart in which contexts non-constant initializers and compound initializers are allowed. What I'm missing (apart from example) is a description of the limitation that we lift. Like, after the example, say "previously, only constant values were permitted in compound initializers". This would probably make some of the original sentence redundant.

RELEASENOTES-1.4.docu
mandolaerik: An example is a polite way to compensate that not everyone knows the term 'designated initializer'. Most of our users are HW-centric coders, not language geeks.

py/dml/ctree.py
lwaern-intel: The designated initializer tests? I don't understand the question -- this initializer is used for designated initializers of all struct-like types except bitfields, so obviously it's covered by the test that tests out the functionality of designated initializers.

py/dml/types.py
lwaern-intel: I'm fine with it, as we know they can't occur in this context.

RELEASENOTES.docu
lwaern-intel: Ok, alright. Latter half scrapped, example added.

RELEASENOTES-1.4.docu
lwaern-intel: Example added.

lwaern-intel:
The const-checking logic has been removed in favor of the work done in #35.

py/dml/ctree.py
mandolaerik: OK, good. I didn't read the full context, wanted to know if this change somehow extends DML and thus needs new tests. I'll take your word for it.

py/dml/codegen.py
```diff
@@ -943,6 +944,101 @@ def add(self, method):
...
+                                  + f"{etype}"
+                                  for field in bad_fields)
+                        + ("\nmissing initializer(s) for field(s) "
+                           + ", ".join(f"'{field}'"
+                                       for field in remaining)
+                           if remaining else '')
+                        )
+
+def loss_on_truncation(v, bits, signed):
+    if not math.isfinite(v) or int(v) != v:
```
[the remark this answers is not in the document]
lwaern-intel: `math.isfinite` returns `False` for NaNs as well. Indeed, it's a safety measure in case float constants can be +/-Infinite or NaNs. From my experimentation, I actually haven't managed to create such a constant expression from DML -- but having this additional check costs nothing.

test/1.4/errors/T_WASTRUNC.dml
```diff
@@ -0,0 +1,57 @@
...
+        -2 };
+    /// WARNING WASTRUNC
+    bf = { -4, 0, 0 };
+    bf = {
+        /// WARNING WASTRUNC
+        0.5,
+        /// WARNING WASTRUNC
+        -0.5,
+        /// WARNING WASTRUNC
+        -0.5 };
```
mandolaerik, line 49: We should add tests for NaN and inf here.

lwaern-intel: No matter what we decide re. `loss_on_truncation`, this won't happen as we can't produce NaN/inf constant expressions in DML.

py/dml/codegen.py
```diff
@@ -943,6 +944,101 @@ def add(self, method):
...
+        return True
+    # (-128).bit_length() == 8, which is not true for signed integers
+    # (bit_length() excludes sign bit).
+    # Adding one for signed negatives compensates for this
+    v = int(v) + (signed and v < 0)
+    # Don't indicate truncation loss on assignments of 1 to int1:s
+    if v == 1 and bits == 1 and signed:
+        return False
+    # Subtract one for signed numbers to compensate for the sign bit
+    return bits - signed < v.bit_length()
```
mandolaerik, line 984: Pondering if something like this would be more direct (after the uint1 check):
```
if signed:
    return (-1 << (bits - 1)) <= v < (1 << (bits - 1))
else:
    return (-1 << (bits - 1)) <= v < (1 << bits)
```
or maybe mask out the bits we lose and see if they contain information:
```
if signed:
    ...
else:
    trunc_mask = (-1 << bits)
    return (v & trunc_mask) == (trunc_mask if v < 0 else trunc_mask)
else:
```

lwaern-intel: I could be ok with the first alternative. The second one I feel is worse than the others; less readable and understandable.

py/dml/codegen_test.py
```diff
@@ -0,0 +1,110 @@
...
+            # Unsigned are treated as groups; min neg allowed for those
+            # are -(2^bits - 1)
+            min_neg = -(2**(bits-signed) - (not signed))
+            no_loss(max_pos, bits)
+            no_loss(min_neg, bits)
+            has_loss(max_pos + 1, bits)
+            has_loss(min_neg - 1, bits)
+
+    # Assuming we would ever add them, (u)int0 types are logically integer
+    # types with only one representable value: 0
```
mandolaerik, line 28: I agree that a `uint0` would be 0, but I'd say it's less clear whether an `int0` is 0 or -1. With a C99 mindset, I would say it's UB since the result depends on the nonexistent sign bit.

py/dml/codegen.py
lwaern-intel: It's stated in the comment above `loss_on_truncation`: to allow users to treat unsigned int types as the groups they are, `loss_on_truncation` also accepts any negation of representable values of the integer type. The max representable value for `uint8` is 255, so the smallest number accepted by `loss_on_truncation` is -255. -255 after truncation will become 1, which is indeed the inverse of 255. -256 is not accepted because 256 is not a representable value within `uint8`, so it doesn't make sense to talk about its inverse.

py/dml/codegen_test.py
lwaern-intel: `int0` is strange enough of a concept that I'd say its definition would be very much open to interpretation. I'd say 0 would make more sense than -1, but as you said, either way the fact the sign bit can't be represented makes it incompatible with C99.

py/dml/codegen.py
mandolaerik: That's a questionable argument. In situations when you think about an uint8 as a group in terms of arithmetic +, all negative numbers are just as much outside its range as 256 and would deserve a warning. (and note that if you use the Peano definition of 256, then it is defined as 255 + 1, which is valid also in the group ℤ<sub>256</sub>, making 256==0 as expected).

However, in cases where it is more natural to think of `uint8` as a group under the xor operator than under +, it is sometimes natural to assign a negative constant, because of the inherent signedness of small integers. Consider, for instance:
```
struct {uint2 x; uint2 y;} not_table = { {0, ~0}, {1, ~1}, {2, ~2}, {3, ~3} };
```
Here, your suggestion will warn for ~3, but not ~2.

(GCC is even more strict, `uint8 ok = ~127, warn = ~128`)

lwaern-intel: After a brief discussion, `loss_on_truncation` now revamped to accept `~v` for any representable `v` when it comes to unsigneds.

mandolaerik (approved):

PR #27, opened 2021-12-07
Ignore gitignore

mandolaerik:


lwaern-intel (approved):
The copyright test is a pain. This is twice we've been struck by "oh, this patch is so obvious it doesn't need validation" only to get told by copyright that it doesn't know the file type.

mandolaerik:
Well, it's hard to avoid if we want to be picky about copyright notices. The ignorelist is short and without clear patterns; with a less  rigid test we would likely have forgot about dmllib.h. The DMLC box runs in under 2 minutes so we should always take the time to wait for that.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9553486

PR #29, opened 2021-12-07
Provide byte order information in register_view for DML 1.4

mandolaerik:

PR #31, opened 2021-12-09
Fully qualify field names relative to register

JonatanWaern:


syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9561180

mandolaerik:
A test is needed. You may want to extend the test added by https://github.com/intel/device-modeling-language/pull/29

mandolaerik (changes requested):
_

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9564765

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9564835

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9564764

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9564834

test/1.4/misc/register_view_fields.py
```diff
@@ -4,15 +4,33 @@
...
                                           ['c', '3', 3, 3],
                                           ['ab', '12', 0, 2]])
-    expect_equal(b.register_info(2)[4], [])
+    expect_same_content(b.register_info(2)[4], [])
+    expect_same_content(b.register_info(3)[4],
+                        [['a', '', 31, 31]] +
+                        [['b.b[%d][%d]' % (i, j), '',
+
+                          30 - i * 2 - j, 30 - i * 2 - j]
+                         for i in range(0,2) for j in range(0,2)] +
```
mandolaerik, line 29: why not `range(2)`? (and if not, space after comma)

lib/1.4/dml-builtins.dml
```diff
@@ -2785,6 +2785,12 @@ template field is (object, _bitsize, _lsb, init_val, shown_desc, _get, _set) {
         set_val(init_val);
     }
 
+    shared method _full_name() -> (const char *) {
```
mandolaerik, line 2788: I'd ask for a comment on return value ownership, but I can see that we lack that on _qname as well so no need to fix now.

mandolaerik:
Approved modulo a nitpick.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9570543

lib/1.4/dml-builtins.dml
```diff
@@ -2763,7 +2763,7 @@ also used by the default implementations of hard reset, soft reset and
 power-on reset. Defaults to 0.
 </dd></dl>
 */
-template field is (object, _bitsize, _lsb, init_val, shown_desc, _get, _set) {
+template field is (object, _bitsize, _lsb, init_val, shown_desc, _get, _set, _qname) {
```
lwaern-intel, line 2766: `_qname` being inherited by `field` increases vtable size. How large of a concern is this?

JonatanWaern: We are adding one small method declaration for each field. This is not entirely trivial but we'd have to do some sort of hack to avoid this

lib/1.4/dml-builtins.dml
```diff
@@ -2785,6 +2785,12 @@ template field is (object, _bitsize, _lsb, init_val, shown_desc, _get, _set) {
         set_val(init_val);
     }
 
+    shared method _full_name() -> (const char *) {
+        local uint64 regname = strlen(reg._qname());
+        local const char *qname = _qname();
+        return qname + regname + 1;
```
lwaern-intel, line 2791: Just want to note that it's a good thing that this is 1.4 exclusive, otherwise this would be a confidentiality leak. (Hell, even instantiating the template would be). A 1.2 model importing a 1.4 module is not a concern as then the dml-builtins.dml in use will be that of DML 1.2.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9570542

lib/1.4/dml-builtins.dml
mandolaerik: It's not only vtable, it's also code generation of the _qname() implementation.  And fields are a major bottleneck. I think this is a reason why we haven't done this before.

So, I think this fix needs a more elaborate study of what it costs. Try to find the worst real example of a slow compilation with too many fields, and see how much slower it gets.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9585103

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9585102

PR #34, opened 2021-12-11
Add alternative syntax #[] for compile-time lists

mandolaerik:


syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9567113

syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9567112

mandolaerik:
The plan is to gradually migrate all known DML 1.4 heterogeneous lists to `#[ ]` syntax, and after that reqiure `[ ]` to be homogeneous. After that we can generalize it into a list value, and permit expressions like `[1,2,3][i]`.

syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9567126

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9567125

py/dml/dmlparse.py
```diff
@@ -1666,6 +1666,17 @@ def expression_paren(t):
 @prod
 def expression_constlist(t):
     'expression : LBRACKET expression_list RBRACKET'
+    s = site(t)
+    exprs = t[2]
+    if logging.show_porting:
+        if not (all(e.kind == 'int' for e in exprs)
+                or all(e.kind == 'string' for e in exprs)):
+            report(PHASHLBRACKET(site(t, 1)))
```
lwaern-intel, line 1674: This is strange to me. First of all, the kind checks feel arbitrary: why specifically do we not report lists of `int`s or `string`s?. Secondly, why have porting logging attached to this? This is not a 1.2 -> 1.4 change -- the removal of `[]` as syntax for compile-time lists would be a 1.4 -> future 1.4/1.6 change -- if anything, we should have an (opt-in) warning for all uses of `[]` today.

mandolaerik: Regarding logging, auto-porting is implemented by emitting a special kind of machine-readable messages, enabled with `--show-porting`, while compiling. These messages are interpreted by a tightly coupled text mangling script, `port_dml.py`. Messages starting with P are all related to porting.

Regarding int/string list: A tricky thing about porting messages is that I want to keep porting tag generation simple and isolated, and that I want to avoid excessive flooding of porting messages. Good coverage is desirable but not essential. In this case, I did a quick grep in real VP code (`[^]@A-Za-z0-9 ] *[[][^.<:]*[]]`) and found that most lists that look homogeneous tend to consist of either integer literals or string literals. I could have moved the check to codegen, and make it capture all lists with constant values, but I was afraid of message flooding from lists expanded in `foreach`s.

The goal here is to find a cheap way to save work. If we can change remaining 1.2->1.4 porting to also do the majority of `#[]` conversions, then conversion of the last pieces will need less work. Also, it may inspire users to start using `#[]` out of sheer superstition (I think primitive conversion rules is the reason why many modelers tend to use the pattern 'int ret; ... ret = X; return ret;` also in new code)

And this *is* a 1.2->1.4 change. I add `#[]` as an alternative syntax within 1.4. Changing [] into a run-time list within 1.4 would break compatibility, but I have a bit of hope that it can be done anyway (perhaps the costs are smaller than the benefits)

syssimics:
Verification [#9818275](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9818275): fail

syssimics:
Verification [#9818883](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9818883): pass

PR #36, opened 2021-12-15
Split transaction accesses at 8-byte address boundaries

svenwest:


syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9577415

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9577422

test/1.4/lib/T_transaction.py
```diff
@@ -35,3 +35,27 @@
...
+                   (Sim_PE_No_Exception, 32, 5))
+for x, y in [(0, 7), (7, 15), (15, 20)]:
+    stest.expect_equal(int.from_bytes(trans.data[x:y], 'little'), 4711)
+
+trans = transaction_t(read=True, size=20)
+stest.expect_equal((b.iface.transaction.issue(trans, 17), b.read_offset),
+                   (Sim_PE_No_Exception, 32))
+for x, y in [(0, 7), (7, 15), (15, 20)]:
+    stest.expect_equal(int.from_bytes(trans.data[x:y], 'little'), 4711)
+
```
mandolaerik, line 61: This doesn't test that the splits happen at the right offsets (or does it?)

lib/1.4/dml-builtins.dml
```diff
@@ -1858,6 +1861,50 @@ template bank is (object, shown_desc, _qname) {
...
+            bytes.data = buf;
+            bytes.len = size;
+            SIM_get_transaction_bytes(t, bytes);
+        }
+
+        local int sz = 8 - (offset % 8);
+        local uint64 offs = 0;
+        while (offs < size) {
+            local atom_t atoms[3] = {
+                ATOM_data(&buf[offs]),
```
mandolaerik, line 1882: `_transaction_access` only uses size, len and data; easier to extract them here and pass as args instead of building a transaction around them. At this stage, any extensions of the transaction are in the aux arg anyway.

lib/1.4/dml-builtins.dml
```diff
@@ -1171,7 +1171,10 @@ bank. The `aux` parameter is NULL by default, and is passed on to bank methods.
 In order to pass additional information on the access down to register and
 field methods, one can override `transaction_access`, decode needed information
 from the incoming transaction, and call `default` with the extracted
-information in the `aux` argument.
+information in the `aux` argument. Accesses that are bigger than 8 bytes are
+split on 8-byte address boundaries. For example, a 12-byte access to address 5
+will be split into one 3-byte access at address 5, one 8-byte access at address
+8, and one 1-byte access at address 16.
```
mandolaerik, line 1177: Maybe we want to change this later, by extending the register dispatcher to handle accesses of arbitrary length. This will not happen in the near future (your fix solves all known problems in practice), but can we reformulate the documentation to disallow users from relying on getting accesses split in the middle of registers?

Also, it can be worth mentioning something about the relation to the read/write methods: transaction_access splits accesses before passing to `read` and `write`, which may split the access further along register boundaries.

mandolaerik (changes requested):

test/1.4/lib/T_transaction.py
svenwest: Indeed, it tests that the access is split in correct sizes, and that the last access is on the correct place.
Since the test-device is quite simple the other accesses _could_ be on any random offset.

lib/1.4/dml-builtins.dml
svenwest: Hm, something like "accesses will be split in chunks of at most 8 bytes size"?

lib/1.4/dml-builtins.dml
svenwest: Ok.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9578565

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9578571

lib/1.4/dml-builtins.dml
svenwest: Actually, `_transaction_access` gets and sets the value of the transaction too. By building a smaller transaction I can wave my hand at those things and just use the transaction api-calls.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9578713

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9578737

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9578777

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9578736

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9578776

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9587301

lib/1.4/dml-builtins.dml
mandolaerik: I gave it a try myself and pushed a patch that uses a buffer.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9587300

lib/1.4/dml-builtins.dml
svenwest: Thanks, looks good!
Funny thing: I can't approve your changes, but you can - is this a sneaky way of bypassing the review-requirements?

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9593135

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9593141

mandolaerik (approved):

PR #38, opened 2021-12-30
Implement dependencies for Ninja build

adklimki:
Prerequisite for https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/1110

### DIFF
--- a/py/dml/dmlc.py
+++ b/py/dml/dmlc.py
@@ -233,6 +233,21 @@ def main(argv):
         '--dep', dest = "makedep", action = 'store',
         help = 'generate makefile dependencies')
 
+    # <dt>--no-dep-phony</dt>
+    # <dd>Avoid addition of a phony target for each dependency other than the main file.</dd>
+    optpar.add_option(
+        '--no-dep-phony', dest = "no_dep_phony", action = 'store_true',
+        help = 'do not generate phony targets for dependencies')
+
+    # <dt>--dep-target</dt>
+    # <dd>Change the target of the rule emitted by dependency generation.
+    # Specify multiple times to have multiple targets.</dd>
+    optpar.add_option(
+        '--dep-target', dest = "dep_target", action = 'append',
+        metavar = 'TARGET',
+        default = [],
+        help = 'change the target of the rule emitted by dependency generation')
+
     # <dt>-T</dt>
     # <dd>Show tags on warning messages. The tags can be used with
     # the <tt>--nowarn</tt> and <tt>--warn</tt> options.</dd>
@@ -497,16 +512,22 @@ def main(argv):
                 # dependencies
                 deplist = set(deplist) - set(future_timestamps)
             deps = ' '.join(path.replace(" ", "\\ ") for path in deplist)
+            if options.dep_target:
+                targetlist = options.dep_target
+            else:
+                targetlist = [options.makedep, f"{outputbase}.c"]
+            targets = ' '.join(path.replace(" ", "\\ ") for path in targetlist)
             if options.makedep_old:
                 f = sys.stdout
             else:
                 f = open(options.makedep, 'w')
             with f:
-                # rebuild from DML if imported file has changed
-                f.write('%s %s.c : %s\n' % (
-                    options.makedep, outputbase, deps))
-                # if imported file vanished, survive and rebuild
-                f.write('%s :\n' % (deps,))
+                f.write('%s : %s\n' % (targets, deps))
+                # By default generate phony targets similar to -MP GCC
+                # argument. Phony targets provide better experience with GNU
+                # Make when dependencies get removed.
+                if not options.no_dep_phony:
+                    f.write('%s :\n' % (deps,))
             sys.exit(0)
 
         dev = process(devname, global_defs, top_tpl, defs)
--- a/py/dml/dmlc.py
+++ b/py/dml/dmlc.py
@@ -234,14 +234,15 @@ def main(argv):
         help = 'generate makefile dependencies')
 
     # <dt>--no-dep-phony</dt>
-    # <dd>Avoid addition of a phony target for each dependency other than the main file.</dd>
+    # <dd>With --dep, avoid addition of a phony target for each dependency
+    # other than the main file.</dd>
     optpar.add_option(
         '--no-dep-phony', dest = "no_dep_phony", action = 'store_true',
         help = 'do not generate phony targets for dependencies')
 
     # <dt>--dep-target</dt>
-    # <dd>Change the target of the rule emitted by dependency generation.
-    # Specify multiple times to have multiple targets.</dd>
+    # <dd>With --dep, change the target of the rule emitted by dependency
+    # generation. Specify multiple times to have multiple targets.</dd>
     optpar.add_option(
         '--dep-target', dest = "dep_target", action = 'append',
         metavar = 'TARGET',
--- a/py/dml/dmlc.py
+++ b/py/dml/dmlc.py
@@ -238,7 +238,8 @@ def main(argv):
     # other than the main file.</dd>
     optpar.add_option(
         '--no-dep-phony', dest = "no_dep_phony", action = 'store_true',
-        help = 'do not generate phony targets for dependencies')
+        help = 'With --dep, avoid addition of a phony target for each'
+        + ' dependency other than the main file.')
 
     # <dt>--dep-target</dt>
     # <dd>With --dep, change the target of the rule emitted by dependency
@@ -247,7 +248,9 @@ def main(argv):
         '--dep-target', dest = "dep_target", action = 'append',
         metavar = 'TARGET',
         default = [],
-        help = 'change the target of the rule emitted by dependency generation')
+        help = 'With --dep, change the target of the rule emitted by'
+        + ' dependency generation. Specify multiple times to have multiple'
+        + ' targets.')
 
     # <dt>-T</dt>
     # <dd>Show tags on warning messages. The tags can be used with
--- a/test/tests.py
+++ b/test/tests.py
@@ -20,6 +20,7 @@ from simicsutils.internal import package_path
 import testparams
 from testparams import simics_root_path
 import traceback
+from depfile import parse_depfile
 
 sys.path.append(join(simics_root_path(), "scripts", "build"))
 import module_id
@@ -123,6 +124,16 @@ class BaseTestCase(object):
         self.output = None
     def pr(self, msg):
         self.output.call(testparams.pr, msg)
+
+    def expect_equal_sets(self, a, b):
+        assert isinstance(a, set) and isinstance(b, set)
+        if a != b:
+            import traceback
+            for line in traceback.format_stack():
+                self.pr(line.rstrip())
+            self.pr('missing elements: %r' % ((b - a),))
+            self.pr('unexpected elements: %r' % ((a - b),))
+            raise TestFail('difference found')
 
     def run_test(self, output_handler):
         self.output = output_handler
@@ -873,8 +884,8 @@ class DebuggableCheck(BaseTestCase):
 
 all_tests.append(DebuggableCheck('debuggable-check'))
 
-class DmlDep(CTestCase):
-    '''Stability test, '''
+class DmlDepBase(CTestCase):
+    '''Base class for DML dependency test cases.'''
     __slots__ = ()
     def run_cc(self, cc_extraargs):
         assert not cc_extraargs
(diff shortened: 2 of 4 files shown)


py/dml/dmlc.py
```diff
@@ -502,11 +508,13 @@ def show_illegal_attributes(option, opt, value, parser):
...
                 f = open(options.makedep, 'w')
             with f:
-                # rebuild from DML if imported file has changed
-                f.write('%s %s.c : %s\n' % (
-                    options.makedep, outputbase, deps))
-                # if imported file vanished, survive and rebuild
-                f.write('%s :\n' % (deps,))
+                    # rebuild from DML if imported file has changed
+                if options.depninja:
+                    f.write('%s.c : %s\n' % (outputbase, deps))
```
adklimki, line 513: Following original code here. If `pathlib.Path(..).write_text(..)` is preferable, it should be a separate PR.

py/dml/dmlc.py
```diff
@@ -502,11 +508,13 @@ def show_illegal_attributes(option, opt, value, parser):
...
+                else:
+                    f.write('%s %s.c : %s\n' % (options.makedep, outputbase, deps))
+                    # if imported file vanished, survive and rebuild
+                    f.write('%s :\n' % (deps,))
```
adklimki, line 517: I did manual tests for both Ninja and Unix Makefiles generators in CMake. When using Makefiles, `--dep` works ok. Ninja generators complain about the extra target for .dmldep file and extra targets for individual files.

test/tests.py
```diff
@@ -990,6 +993,46 @@ def test(self):
...
+                   if not p.startswith(project_host_path())}
+        # Main .c target depends on main DML file and all DML imports.
+        # We test that all non-library DML files are included, as well as
+        # one library file
+        self.expect_equal_sets(non_lib, {
+            os.path.normpath(join(testdir, '1.2', 'misc', f + '.dml'))
+            for f in ['T_import_rel_1', 'rel/a/x', 'rel/a/y',
+                      'rel/z', 'rel/misc/rel/z']})
+
+        assert list(target_prereqs.keys()) == [c_target], "Ninja deps should have only .c target"
```
adklimki, line 1029: I added a test here. There is a minimal test for CMake build `test/t1019_project_cmake`. However, we are missing some critical pieces to add more. In particular `project-setup` arguments to generate devices need to understand `--cmake` option to create `CMakeLists.txt` file instead of `Makefile`. Once this is done, we can add integrations tests for CMake build to cover:
* Ninja and Unix Makefiles generators
* Build for DML, C, and other devices

adklimki:
@mandolaerik @andreashedstrom-intel

test/tests.py
```diff
@@ -503,6 +513,32 @@ def verify_dmlc_messages(self, stderr, expected_msgs):
             self.expect_messages(
                 kind, actual_msgs.get(kind, []), expected_msgs.get(kind, []))
 
+    @staticmethod
+    def parse_depfile(f):
```
mandolaerik, line 517: Doesn't feel natural to have this as a method on DMLFileTestCase, the method is too specific. Its scope is highly specific and limited to dependency file tests. If you must extract it to a common place, then maybe a top-level function makes most sense. (But even better to not duplicate the DmlDep test)

test/tests.py
```diff
@@ -990,6 +993,46 @@ def test(self):
                         dmlc_extraargs = ['--dep', 'T_dmldep.dmldep'],
                         api_version=latest_api_version))
 
+class DmlDepNinja(DMLFileTestCase):
```
mandolaerik, line 996: This looks like a modified copy of DmlDep. How much pain would it be to extend DmlDep with parameters to make that class cover both flags?

py/dml/dmlc.py
mandolaerik: Agreed (it's preferrable, and a separate PR)

py/dml/dmlc.py
mandolaerik: I would prefer to make the args less tied to particular build systems, instead of more. The current --dep option is tailored for one particular build system (`module.mk` interpreted by GNU make). If we need something more customized, then perhaps we should extract separate options similar to gcc's `-MP` and `-MT`.

For instance:
* add `--no-phony-dep`, corresponding to -MP. (perhaps we'll later add --phony-dep and invert the default)
* add `--dep-target` for -MT, with gcc-like semantics (preserving the current default of two targets).

test/tests.py
mandolaerik: @andreashedstrom-intel any comments on roadmap for t1019 or `project-setup --cmake`?

py/dml/dmlc.py
```diff
@@ -502,11 +508,13 @@ def show_illegal_attributes(option, opt, value, parser):
             else:
                 f = open(options.makedep, 'w')
             with f:
-                # rebuild from DML if imported file has changed
-                f.write('%s %s.c : %s\n' % (
-                    options.makedep, outputbase, deps))
-                # if imported file vanished, survive and rebuild
-                f.write('%s :\n' % (deps,))
+                    # rebuild from DML if imported file has changed
```
andreashedstrom-intel, line 511: This line looks wrong w.r.t number of spaces. I think python will choke on this. I expect old line 505 and new line 511 to be identical.

test/tests.py
andreashedstrom-intel: I don't understand what @adklimki is referring to here; the t1019 test *does* use --cmake to create the project with top-level CMakeLists.txt for building with CMake instead of GNU Make.

The project-setup --copy-module can be used to copy modules into the project, but as t1019 currently builds *all* converted modules distributed by Simics Base there is (currently) no need to add this. If we add new dummy devices that we don't want to distribute with Simics Base these devices can be manually copied/symlinked to the project directly from the outer repo.

t1019 currently assumes we run with -G Ninja but can easily be updated to run with -G Unix Makefiles (I'll add this as a separate PR; we should probably have this to explicitly cover the case where we want to support both generators). 

As we build all converted modules, we already build DML, C, C++ and python based modules in t1019.

mandolaerik: Since we don't have a specification of Ninja's depfile requirements, it would be great with some kind of complementary smoke test that DML build deps work.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9618914

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9618913

test/tests.py
adklimki: @andreashedstrom-intel I mean this command creates a Makefile, not CMakeLists.txt:
```
> ./bin/project-setup --cmake --dml-device XXX
Skeleton for module 'XXX' successfully created.
> ls -la modules/XXX
total 24
drwxr-x--- 3 adklimki users 4096 Jan  4 12:08 ./
drwxr-x--- 3 adklimki users 4096 Jan  4 12:08 ../
-rw-r----- 1 adklimki users  402 Jan  4 12:08 Makefile
-rw-r----- 1 adklimki users 1724 Jan  4 12:08 XXX.dml
-rw-r----- 1 adklimki users  468 Jan  4 12:08 module_load.py
drwxr-x--- 2 adklimki users 4096 Jan  4 12:08 test/
```

I see there are tests that use these templates to test the build scripts.

andreashedstrom-intel: ah, ok, yes now *that* is indeed missing. Current flow just supports building already converted modules. And converting a skeleton device is straight forward and easily done and should work out-of-the-box by invoking cmake/gmake-to-cmake.py on the modules/XXX/Makefile to produce the modules/XXX/CMakeLists.txt.

But yes, I agree, this should be done by project-setup to save the user from this. Not sure if we need a special --cmake flag here. We could simply generate both CMakeLists.txt and Makefile and let the user choose which one to keep. What do you think @mandolaerik ?

The t1019 now builds both with Ninja and Unix Makefiles and as mentioned it builds all converted modules so it's both C, C++, DML and Python. It does not run any tests though. We probably want a dedicated suite for smoking out problems with the DML deps as Erik suggested. We already know C/C++ deps work, as that works out-of-the-box in CMake.

py/dml/dmlc.py
andreashedstrom-intel: Indeed, this is something for @mandolaerik to dig into

test/tests.py
mandolaerik: It should be easy to write a smoke test that either writes a hardcoded CMakeLists.txt, or runs a conversion script to produce one. I assume cmake is still in tech preview so we should not yet create it by default from project-setup. We should eventually add this, and then we can remove the explicit generation from the smoke test.

py/dml/dmlc.py
mandolaerik: Python ignores the indentation of all blank lines, with or without comments.

adklimki:
@mandolaerik I see this test fails on the stable `6` branch:
```
/scratch/adklimki/simics-base/logs/test/linux64/src/tools/dmlc/test/test.log:375: test copyright in src/tools/dmlc/test failed (*** failed (missing copyrights) ***)
```
The log file `logs/test/linux64/src/tools/dmlc/test/test.log`:
```
/scratch/adklimki/simics-base/src/tools/dmlc/Makefile:1: error: copyright year too old
...
```

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9630885

py/dml/dmlc.py
```diff
@@ -233,6 +233,21 @@ def main(argv):
         '--dep', dest = "makedep", action = 'store',
         help = 'generate makefile dependencies')
 
+    # <dt>--no-dep-phony</dt>
+    # <dd>Avoid addition of a phony target for each dependency other than the main file.</dd>
```
mandolaerik, line 237: Some context, e.g. "With `--dep`, ..." (docs + help)

mandolaerik:
Looks great, just a doc nitpick

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9630884

py/dml/dmlc.py
```diff
@@ -233,6 +233,22 @@ def main(argv):
         '--dep', dest = "makedep", action = 'store',
         help = 'generate makefile dependencies')
 
+    # <dt>--no-dep-phony</dt>
+    # <dd>With --dep, avoid addition of a phony target for each dependency
+    # other than the main file.</dd>
+    optpar.add_option(
+        '--no-dep-phony', dest = "no_dep_phony", action = 'store_true',
+        help = 'do not generate phony targets for dependencies')
```
mandolaerik, line 241: help text should match doc (both flags)

mandolaerik (approved):

mandolaerik:
either of @lwaern-intel @JonatanWaern, please review the design of the new dep flags

py/dml/dmlc.py
```diff
@@ -233,6 +233,25 @@ def main(argv):
         '--dep', dest = "makedep", action = 'store',
         help = 'generate makefile dependencies')
 
+    # <dt>--no-dep-phony</dt>
```
lwaern-intel, line 236: `--no-dep-phony` or `--no-phony-dep`?

py/dml/dmlc.py
```diff
@@ -233,6 +233,25 @@ def main(argv):
...
+    # <dd>With --dep, avoid addition of a phony target for each dependency
+    # other than the main file.</dd>
+    optpar.add_option(
+        '--no-dep-phony', dest = "no_dep_phony", action = 'store_true',
+        help = 'With --dep, avoid addition of a phony target for each'
+        + ' dependency other than the main file.')
+
+    # <dt>--dep-target</dt>
+    # <dd>With --dep, change the target of the rule emitted by dependency
+    # generation. Specify multiple times to have multiple targets.</dd>
```
lwaern-intel, line 246: I'd prefer if it were possible to specify multiple targets with only one specification of `--dep-target`. Is it?

lwaern-intel (commented):
The flags seem good to me except for some minor questions I have.

py/dml/dmlc.py
andreashedstrom-intel: The user of this flag will be our CMake (or possibly some GNU Make?) code where it doesn't matter what we do here. I think having explicit --dep-target x --dep-target y --dep-target z is better and more explicit than --dep-target x y z

Also, --warn down below works exactly like this option, making it consistent.

py/dml/dmlc.py
andreashedstrom-intel: I'd prefer --no-phony-dep or possibly --dep-no-phony (as it's a sub-option of --dep)

lwaern-intel: @mandolaerik Opinions?

mandolaerik: The reasoning here is that we can inverse the defaults eventually: make `no` the default and let gmake pass `--dep-phony`. The `--no-dep-phony` name then comes from the convention that the inverse of `--X` is named `--no-X`.

`--no-phony-dep` would be consistent with this. I welcome suggestions for better names, though; "phony dep" is not overwhelmingly descriptive.

adklimki: I second @mandolaerik here. `--dep-phony` is related to `--dep` and would add phony targets. `--no-dep-phony` is a reversal.

andreashedstrom-intel: fair enough, issue resolved I guess

PR #42, opened 2022-01-13
Fix python warnings (from fisketur)

mandolaerik:


syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9645788

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9645787

lwaern-intel (approved):
Confirmed warnings get generated without this patch, and are not with it.

PR #43, opened 2022-01-19
Multiple simultaneous variable declarations (SIMICS-7027); initializer syntax for return statements; improved error messages for method calls

lwaern-intel:


syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9660914

syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9660913

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9663820

syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9663818

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9673769

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9675673

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9675627

test/1.4/statements/T_return.dml
```diff
@@ -0,0 +1,40 @@
...
+  © 2022 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+dml 1.4;
+device test;
+import "testing.dml";
+
+typedef struct {
+    int i;
+    bool b;
```
mandolaerik, line 11: make one field `const` for more fun!

test/1.4/syntax/T_assign.dml
```diff
@@ -53,4 +53,9 @@ method test() throws {
     i = 3;
     c[++j] += ++i;
     expect(c[1] == 4 && j == 1 && i == 4, "+= operands evaluated twice");
+
+    i = 1;
+    j = 0;
+    (i, j) = (j, i);
+    expect(i == 0 && j == 1, "multiple simultaneous assignment");
```
mandolaerik, line 60: I prefer `assert` over `expect` also in surroundings where `expect` is the dominant function. If you prefer to follow the surrounding pattern then I'm fine with that.

py/dml/codegen.py
```diff
@@ -1624,10 +1733,57 @@ def stmt_return_dml12(stmt, location, scope):
...
+                return ([sym_declaration(sym) for sym in outarg_syms]
+                        + [method_invocation,
+                           codegen_exit(stmt.site, outargs)])
+
+    if len(inits) != len(outp_typs):
+        report(ERETARGS(stmt.site, len(outp_typs), len(inits)))
+        return [mkAssert(stmt.site, mkBoolConstant(stmt.site, False))]
+
+    return [codegen_exit(stmt.site,
+                         [initast_to_expr(stmt.site, init, typ, location,
```
mandolaerik, line 1784: I'm somewhat surprised that this is sufficient -- I had expected that we had to initialize an intermediate local variable. Is this a piece of joy brought by your const fixes?

mandolaerik (commented):
Looks overall good. I haven't examined all parts of codegen in detail.

py/dml/codegen.py
```diff
@@ -221,6 +217,14 @@ def declarations(scope):
 
     return decls
 
+def initast_to_expr(site, init, typ, location, scope):
+    init = eval_initializer(site, typ, init, location, scope, False)
+    if isinstance(init, ExpressionInitializer):
+        return init.expr
+    else:
+        return mkLit(site, f'({typ}) {init.read()}', typ)
```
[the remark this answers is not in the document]
lwaern-intel: _Oops._ Will fix that and add a test for it.
I'll look into `as_expr()`. If that doesn't work out, I'll add the assert.

test/1.4/syntax/T_multidecls.dml
```diff
@@ -0,0 +1,54 @@
...
+method wrapper() -> (bool, int) {
+    return (false, 4);
+}
+
+method test() throws {
+    cast(b, t).q = 9;
+    b.r = 10;
+    b.s = 11;
+    b.t = 12;
+    local (bool x, int y) = (false, 4);
```
mandolaerik, line 32: It struck me that an alternative syntax that we haven't discussed is `local (bool, int) (x, y);`. With the thinking that `(bool, int)` is a type (ok not yet but maybe later), and `(x, y)` deconstructs this tuple type.

I think the current syntax looks nicer at first glance, but the idea has some merit and warrants a discussion before dismissed.

test/1.4/statements/T_throwing_initializers.dml
```diff
@@ -0,0 +1,35 @@
...
+        throw;
+    }
+    return (3, 5);
+}
+
+method test() throws {
+    local int x = 1;
+    // Throwing methods may be used as initializers in DML 1.4
+    try {
+        local int i = demand_a(true);
```
mandolaerik, line 27: how can initializers throw, if they can also always be expressed as expressions?

py/dml/codegen.py
mandolaerik: does initast_to_expr mean that it's really easy to permit initializers in method args?

test/1.4/statements/T_return.dml
lwaern-intel: Hey, that reveals a bug in how `codegen_exit` works! Not exclusive to this patch! I'll create a JIRA.

lwaern-intel: Actually, wait, it _is_ exclusive to this patch! That wasn't obvious at first. Neat!

lwaern-intel: I fixed this by banning methods from having (partially) const outputs, period. This needs to be tested against VP.

syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9764953

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9766584

syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9766583

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9766921

py/dml/structure.py
```diff
@@ -2592,6 +2592,11 @@ def mkmethod(site, rbrace_site, location, parent_obj, name, inp_ast,
             if t.is_int and t.is_endian:
                 raise EEARG(site)
 
+    for (n, t) in outp:
+        if t and deep_const(t):
+            raise ICE(site,
```
mandolaerik, line 2597: ICEs that are known to be reachable should be accompanied with a jira-referencing comment. A failing test is good but not strictly necessary.

py/dml/codegen.py
```diff
@@ -1620,58 +1609,55 @@ def try_codegen_invocation(site, init_ast, outargs, location, scope):
...
-    elif tgts_kind == 'tuple':
-        method_tgts = tgts
     else:
-        raise ICE(site, 'unknown LHS kind')
+        method_tgts = tgts
 
     # TODO support multiple assign sources. It should be generalized.
     method_invocation = try_codegen_invocation(site, src_asts, method_tgts,
                                                location, scope)
     if method_invocation:
```
mandolaerik, line 1634: Below this line, there's about 2 lines of code in this function that isn't under a tgt_ast.kind condition. I think control flow will be clearer if you unify those ifs into one, swallowing the cost of duplicating those two lines (avoids flag programming). You might consider breaking out two functions for that. (if this turns out nice, then you could consider one further step: split stmt_assign in two, assign_chain/assign_tuple. Cleaner AST at the cost of a few more lines of dup)

lwaern-intel: Those lines are the following:
```
    lscope = Symtab(scope)
    stmts = []
    ...
    return stmts
```
I'm OK with duplicating `lscope`, but I disagree that duplicating `stmts` and `return stmts` makes the control flow clearer.

syssimics:
Verification [#9797761](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9797761): pass

py/dml/codegen.py
```diff
@@ -1534,60 +1609,87 @@ def try_codegen_invocation(site, apply_ast, outargs, location, scope):
...
+        for (i, tgt) in enumerate(reversed(tgts[1:])):
+            name = 'tmp%d' % (i,)
+            sym = lscope.add_variable(
+                name, type=tgt.ctype(), site=tgt.site, init=init, static=False,
+                stmt=True)
+            init = ExpressionInitializer(mkLocalVariable(tgt.site, sym))
+            stmts.extend([sym_declaration(sym),
+                          mkAssignStatement(tgt.site, tgt, init)])
+        return stmts + [mkAssignStatement(tgts[0].site, tgts[0], init)]
+    elif tgt_ast.kind == 'assign_target_tuple':
```
mandolaerik, line 1660: Please protect from implicit return, e.g.
```
else:
  assert ast.kind == '...'
  ...
```

lwaern-intel: Nice catch.

syssimics:
Verification [#9816218](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9816218): pass

py/dml/messages.py
```diff
@@ -633,6 +633,24 @@ def __init__(self, fun, ftype = None):
...
+    be called either as a standalone statement, as the RHS of an assignment, or
+    as the argument of a `return` statement.
+    """
+    fmt = "call to method '%s' in unsupported context\n%s"
+    def __init__(self, site, fun):
+        if site.dml_version() == (1, 2):
+            suggestion = ("use the 'call' or 'inline' statements to call "
+                          + "this method")
+        else:
+            suggestion = ("perform this method call either as a standalone "
```
mandolaerik, line 650: For multi-line, please add site to the second line. The current convention is:
```
    def log(self):
        DMLError.log(self)
        self.print_site_message(self.site, "...")
```
If you find this too verbose, then I wouldn't mind polishing the idiom.

test/1.4/syntax/T_multidecls.dml
mandolaerik: Discussed and dismissed.

py/dml/messages.py
lwaern-intel: That's not at all the convention. Multi-line error messages use `\n` all over the place, `self.print_site_message` is typically used only when there are multiple sites; in fact, there is only one case where it's used with `self.site`, namely `EINC`.

mandolaerik: ok, sorry for the noise

mandolaerik (approved):
Approved after fixing the last nitpick

syssimics:
Verification [#9827926](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9827926): fail

mandolaerik:
Interesting, build error on something like:
```
method m() -> (uint32) { return 4; }
method foo() {
  // illegal function application of m of type uint32
  local uint32 m = m();
}

**Comment by lwaern-intel** [comment]
For some reason, the comment failed to be created, but verification passed: https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9831164

**Comment by lwaern-intel** [comment]
> When you use add_variable here, it means that the `_tmp_name` variable is added to DML scope and can in principle be referenced by DML code. Not a problem in practice, but you could consider a more direct approach where the declaration is created with `mkDeclaration` and the reference is created with `mkLit`.

`add_variable` is done with the newly created `local_scope`, which is not used otherwise. That means the variables are impossible to access from within DML, so there's no issue at all.

File: py/dml/codegen.py [hunk 11]
```diff
@@ -1244,87 +1239,168 @@ def check_varname(site, name):
...
+        rt = safe_realtype_shallow(etype)
+        if isinstance(rt, TArray) and not rt.size.constant and deep_const(rt):
+            raise EVLACONST(stmt.site)
+        check_shadowing(scope, name, stmt.site)
+        return (name, etype)
+
+    decls = list(map(convert_decl, decls))
+
+    def add_var(scope, name, typ, init = None,
+                make_unique=not dml.globals.debuggable):
```
mandolaerik, line 1272: unused arg

py/dml/codegen.py
```diff
@@ -1244,87 +1239,168 @@ def check_varname(site, name):
...
+        local_scope = Symtab(scope)
+        late_sym_adds = []
+        late_declares = []
+        tgts = []
+
+        for (name, typ) in decls:
+            if typ.const:
+                nonconst_typ = typ.clone()
+                nonconst_typ.const = False
+                sym = add_var(local_scope, '_tmp_' + name, nonconst_typ,
```
mandolaerik, line 1289: why don't you just create a LocalSymbol like in the else clause?  This way you can remove local_scope. You can still enclose that in a function add_var if you like.

py/dml/codegen.py
```diff
@@ -1244,87 +1239,168 @@ def check_varname(site, name):
...
+        return sym
+
+    if method_call_init:
+        local_scope = Symtab(scope)
+        late_sym_adds = []
+        late_declares = []
+        tgts = []
+
+        for (name, typ) in decls:
+            if typ.const:
```
mandolaerik, line 1286: Perhaps you need realtype(typ) first, what if typ is a `typedef const int const_int_t;`? Also, what about deep_const?

lwaern-intel: `safe_realtype_shallow` makes sense. The partially const situation has no good solution, which is why (partially) const method outputs are now banned.

syssimics:
Verification [#9834337](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9834337): pass

mandolaerik (approved):

PR #44, opened 2022-01-20
Add support for `break` in `#foreach` statements

lwaern-intel:
As requested in HSD-ES ticket 1309451301.


syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9664422

RELEASENOTES-1.4.docu
```diff
@@ -280,4 +280,6 @@
       local size_t size = get_size();
       local buffer_t buf = {.len = size, .data = new uint8[size]};
       </pre></add-note></build-id>
+  <build-id value="next"><add-note> <tt>break</tt> may now be used within
```
mandolaerik, line 283: hsd-es reference?

lwaern-intel: True, forgot.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9664421

py/dml/ctree.py
```diff
@@ -379,6 +391,12 @@ def control_flow(self):
 
 mkThrow = Throw
 
+class UnrolledBreak(Goto):
+    def control_flow(self):
+        return ControlFlow(fallthrough=True)
```
mandolaerik, line 396: should be `br=True`? Test needed (see T_ENORET)

mandolaerik (changes requested):

py/dml/codegen.py
```diff
@@ -2005,6 +2021,8 @@ def foreach_constant_list(site, itername, lst, statement, location, scope):
                     stmt)
             spec.append(mkCompound(site, decls + [stmt]))
 
+        if site.dml_version() != (1, 2):
```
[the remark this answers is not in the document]
lwaern-intel: I didn't want to support 1.2 because it would mean documenting it for 1.2, but I suppose that's not sufficient reason.

py/dml/codegen.py
```diff
@@ -87,6 +87,21 @@ def gensym(prefix = '_gensym'):
...
+    pass
+
+class MethodLoop(Loop):
+    pass
+
+class UnrolledLoop(Loop):
+    count = 0
+    def __init__(self):
+        self.id = UnrolledLoop.count
+        UnrolledLoop.count += 1
```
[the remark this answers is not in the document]
lwaern-intel: Sounds good.

py/dml/codegen.py
```diff
@@ -87,6 +87,21 @@ def gensym(prefix = '_gensym'):
 # Saved variables in methods, method->list of symbols
 saved_method_variables = {}
 
+class Loop(object):
```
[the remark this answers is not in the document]
lwaern-intel: I was wondering about it. Now I know.

py/dml/codegen.py
```diff
@@ -2005,6 +2021,8 @@ def foreach_constant_list(site, itername, lst, statement, location, scope):
                     stmt)
             spec.append(mkCompound(site, decls + [stmt]))
 
+        if site.dml_version() != (1, 2):
+            spec.append(mkUnrolledLoopLabel(site, unroll.id))
```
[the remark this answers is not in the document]
lwaern-intel: Selectively generating the label is actually easier than I first thought; I can have `stmt_break` mutate a `used` variable of the `UnrolledLoopContext` item in the `loop_stack`. So it's straight-forward to eliminate the label for `#foreach` loops that don't use `break` at all.
I don't want to drop the `UNUSED` annotation though, because I'm concerned some code elimination shenanigans might drop a break statement, even though `stmt_break` is called.

py/dml/ctree.py
lwaern-intel: I had tested with `br=True`, but that resulted in broken generated C. Same thing with `fallthrough=False`.
I agree, test is warranted.

lwaern-intel: Indeed, this incorrectl y reports `ENORET`:
```
method m10() -> (int) {
    try {
        #foreach b in ([dev]) {
            break;
            // this statement is dead...
            throw;
        }
        return 0;
    } catch {
        // .. so this block is also dead ..
    }
    // .. so finally, this point is unreachable and we should get no error
}
```
The obvious way to fix this requires the addition of a DML statement class corresponding to `#foreach` statements.

lwaern-intel: Aforementioned fix implemented.

py/dml/codegen.py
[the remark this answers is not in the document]
lwaern-intel: Done.

py/dml/codegen.py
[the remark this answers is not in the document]
lwaern-intel: Done.

py/dml/codegen.py
```diff
@@ -2087,19 +2110,24 @@ def stmt_switch(stmt, location, scope):
...
 @statement_dispatcher
 def stmt_break(stmt, location, scope):
-    if not loop_stack or loop_stack[-1] == 'method':
+    if not loop_stack or isinstance(loop_stack[-1], NoLoopContext):
         raise EBREAK(stmt.site)
-    elif loop_stack[-1] == 'c':
+    elif isinstance(loop_stack[-1], CLoopContext):
         return [mkBreak(stmt.site)]
+    elif (isinstance(loop_stack[-1], UnrolledLoopContext)
+          and stmt.site.dml_version() != (1, 2)):
```
lwaern-intel, line 2127: forgot to remove this 1.2 gate.

py/dml/messages.py
```diff
@@ -849,7 +849,7 @@ class EBREAK(DMLError):
 class EBREAKU(DMLError):
     """
     A `break` statement cannot be used in a `foreach`
-    or `select` statement.
+    or `select` statement in DML 1.2.
```
lwaern-intel, line 852: Only `select` statement, now.

py/dml/codegen.py
```diff
@@ -2016,7 +2039,7 @@ def stmt_while(stmt, location, scope):
     if stmt.site.dml_version() == (1, 2) and cond.constant and not cond.value:
         return [mkNull(stmt.site)]
     else:
-        loop_stack.append('c')
+        loop_stack.append(CLoopContext())
         try:
```
mandolaerik, line 2043: Perhaps a context manager would make sense here, i.e. `with CLoopContext():`, like we do with `codegen.Failure`.

lwaern-intel: That does sound reasonable.

py/dml/codegen.py
```diff
@@ -2087,19 +2110,24 @@ def stmt_switch(stmt, location, scope):
...
 @statement_dispatcher
 def stmt_continue(stmt, location, scope):
-    if not loop_stack or loop_stack[-1] == 'method':
+    if not loop_stack or isinstance(loop_stack[-1], NoLoopContext):
         raise ECONT(stmt.site)
-    elif loop_stack[-1] == 'c':
+    elif isinstance(loop_stack[-1], CLoopContext):
         return [mkContinue(stmt.site)]
     else:
         raise ECONTU(stmt.site)
```
mandolaerik, line 2118: It would make sense to permit `continue` as well, for symmetry. Not strictly needed to be part of this patch though.

lwaern-intel: `continue` would clutter up generated code with a lot of `goto`s and labels; and besides, unlike `break`, there are easy usages of `if`-`else` that can be used instead of `continue`. For those reasons, I'm not as keen on supporting `continue`.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9668002

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9668001

py/dml/codegen.py
```diff
@@ -1975,10 +2001,10 @@ def stmt_hashforeach(stmt, location, scope):
 def foreach_constant_list(site, itername, lst, statement, location, scope):
     assert isinstance(lst, AbstractList)
     spec = []
-    try:
-        loop_stack.append('unroll')
+    unroll = UnrolledLoopContext()
+    with unroll:
```
mandolaerik, line 2005: You could rephrase as `with UnrolledLoopContext() as unroll:`, if you let `__enter__` return `self`. I don't know if that's better or worse.

py/dml/codegen.py
```diff
@@ -2034,74 +2055,75 @@ def stmt_for(stmt, location, scope):
...
+                              "duplicate default label")
+            if defaults:
+                for sub in stmts[defaults[0]:]:
+                    if isinstance(sub, ctree.Case):
+                        raise ESWITCH(sub.site,
+                                      "case label after default label")
+            body = ctree.Compound(body_ast.site, stmts)
+        else:
+            body = codegen_statement(body_ast, location, scope)
+
```
mandolaerik, line 2098: Nice, this looks like a bugfix, ESWITCH could push the loop_stack out of sync.

py/dml/codegen.py
mandolaerik: Fair enough. I don't have a problem with procrastinating `continue` until someone asks for it.

py/dml/codegen.py
```diff
@@ -2034,74 +2055,75 @@ def stmt_for(stmt, location, scope):
...
+    if (LoopContext.current is None
+        or isinstance(LoopContext.current, NoLoopContext)):
         raise EBREAK(stmt.site)
-    elif loop_stack[-1] == 'c':
+    elif isinstance(LoopContext.current, CLoopContext):
         return [mkBreak(stmt.site)]
+    elif isinstance(LoopContext.current, UnrolledLoopContext):
+        LoopContext.current.used = True
+        return [mkUnrolledBreak(stmt.site,
+                                f'_unrolledbreak{LoopContext.current.id}')]
```
mandolaerik, line 2122: We could consider making this a method on LoopContext, `return LoopContext.current.break_()`

py/dml/ctree.py
```diff
@@ -356,6 +358,28 @@ def toc(self):
...
+
+    def toc_inline(self):
+        for substatement in self.substatements:
+            substatement.toc()
+        if self.break_id is not None:
+            out(f'_unrolledbreak{self.break_id}: UNUSED;\n')
+
+    def control_flow(self):
+        bodyflow = mkCompound(self.site, self.substatements).control_flow()
+        return bodyflow.replace(fallthrough=True, br=False)
```
mandolaerik, line 379: how do we know that `fallthrough=True`? Consider e.g.: 
```
method m() -> (int) {
  #foreach x in ([0]) {
    return x;
  }
}
```
Wouldn't a more correct value be `fallthrough=bodyflow.fallthrough or bodyflow.br`?

py/dml/ctree.py
```diff
@@ -356,6 +358,28 @@ def toc(self):
...
+    def toc(self):
+        out('{\n', postindent = 1)
+        self.toc_inline()
+        out('}\n', preindent = -1)
+
+    def toc_inline(self):
+        for substatement in self.substatements:
+            substatement.toc()
+        if self.break_id is not None:
+            out(f'_unrolledbreak{self.break_id}: UNUSED;\n')
```
mandolaerik, line 375: is UNUSED needed? (I'm just curious; if you are unsure and just added this defensively then that's ok)

py/dml/codegen.py
lwaern-intel: I'd prefer not to.

py/dml/codegen.py
lwaern-intel: Yeah, I figured the old `try`-`finally` was weird.

py/dml/codegen.py
lwaern-intel: I very slightly prefer keeping the code in `stmt_break`, but I don't have strong feelings about it.

py/dml/ctree.py
lwaern-intel: Oops. You're quite right.

py/dml/codegen.py
mandolaerik: fair enough

py/dml/ctree.py
lwaern-intel: I mentioned it in an earlier comment, but yeah, it's done defensively. I'm slightly concerned some code elimination shenanigans might drop a break statement, even though `stmt_break` is called.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9672848

py/dml/codegen.py
mandolaerik: It would eliminate the `else` clause.

Also, would make sense to store the full label `_unrolledbreak{id}` instead of the ID in LoopContext?

lwaern-intel: > It would eliminate the else clause.

I suppose that's a good enough argument to go with that route.

> Also, would make sense to store the full label _unrolledbreak{id} instead of the ID in UnrolledLoopContext?

I'd much sooner prefer that `UnrolledBreak.toc()` adds `_unrolledbreak` and `mkUnrolledBreak` just gets passed the ID.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9672847

py/dml/codegen.py
mandolaerik: The identifier is written in two places (goto and label); it would be good if we didn't duplicate this string in two classes. (ok to repeat it twice within one class)

lwaern-intel: Ah, yes, that makes sense.

py/dml/ctree.py
```diff
@@ -356,6 +358,29 @@ def toc(self):
...
+
+    def control_flow(self):
+        bodyflow = mkCompound(self.site, self.substatements).control_flow()
+        return bodyflow.replace(fallthrough=(bodyflow.fallthrough
```
mandolaerik, line 379: test?

lwaern-intel: Sure.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9673039

mandolaerik (approved):

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9673038

PR #45, opened 2022-01-22
If the preferred GCC path does not exist on Linux, fall back

zhang-peng-j:
If build-deps does not exist, use the default path of GCC. 

Otherwise, a lot of unit tests would end up with the following error message:

FileNotFoundError: [Errno 2] No such file or directory: '/nfs/simics/build-deps/linux64/6.18/packages/gcc_6.4.0/bin/gcc': '/nfs/simics/build-deps/linux64/6.18/packages/gcc_6.4.0/bin/gcc'

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -95,10 +95,10 @@ if is_windows():
     if not cc:
         try:
             cc = testparams.mingw_cc()
-            cc_not_found = not os.path.exists(cc)
+            cc_found = os.path.exists(cc)
         except Exception:
-            cc_not_found = True
-        if cc_not_found:
+            cc_found = False
+        if not cc_found:
             raise TestFail('gcc not found(specify gcc by env var DMLC_CC)')
         ldflags = [f"-L{testparams.mingw_root()}/lib/gcc/x86_64-w64-mingw32/lib"]
     else:
--- a/test/tests.py
+++ b/test/tests.py
@@ -90,29 +90,29 @@ def dmlc_reaper_args(exitcode_file, timeout_multiplier = 1):
 
 common_cflags = ["-O2", "-std=gnu99", '-Wall', '-Werror', '-Wpointer-arith',
                  '-Wwrite-strings', '-Wformat-nonliteral',]
-cc = [os.environ.get('DMLC_CC')]
+cc = os.environ.get('DMLC_CC')
 if is_windows():
-    if not cc[0]:
+    if not cc:
         try:
-            cc = [testparams.mingw_cc()]
-            cc_not_found = not os.path.exists(cc[0])
-        except FileNotFoundError:
+            cc = testparams.mingw_cc()
+            cc_not_found = not os.path.exists(cc)
+        except Exception:
             cc_not_found = True
         if cc_not_found:
             raise TestFail('gcc not found(specify gcc by env var DMLC_CC)')
-    cflags = common_cflags + [
-        "-DUSE_MODULE_HOST_CONFIG", "-D__USE_MINGW_ANSI_STDIO=1"]
-    if not cc[0]:
         ldflags = [f"-L{testparams.mingw_root()}/lib/gcc/x86_64-w64-mingw32/lib"]
     else:
         ldflags = []
+    cflags = common_cflags + [
+        "-DUSE_MODULE_HOST_CONFIG", "-D__USE_MINGW_ANSI_STDIO=1"]
 else:
-    if not cc[0]:
-        cc = [join(package_path(), "gcc_6.4.0", "bin", "gcc")]
-        if not os.path.exists(cc[0]):
+    if not cc:
+        cc = join(package_path(), "gcc_6.4.0", "bin", "gcc")
+        if not os.path.exists(cc):
             raise TestFail('gcc not found(specify gcc by env var DMLC_CC)')
     cflags = ['-g', '-fPIC', '-Wundef'] + common_cflags
     ldflags = []
+cc = [cc]
 cflags_shared = ["-shared"]
 
 os.environ['DMLC_DEBUG'] = 't'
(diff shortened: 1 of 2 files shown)


mandolaerik (changes requested):
I think you can control package_path() by setting the SIMICS_BUILD_DEPS_ROOT variable.

There is a point that the test runs against a specific gcc version (in case generated code might depend on specific gcc feature). There is also a point that gcc can be overridden (in particular, would be great to run with different `-O` flags depending on speed requirements). But silently falling back to whatever default `gcc` is problematic since it can cause the test to get intricate test failures depending on which machine it is being run.

Maybe add an env variable DMLC_CC, to explicitly override gcc, and hint the user about this variable if the default path is missing?

zhang-peng-j:
Ok. I will introduce DMLC_CC. The changes passed on Linux. But seems it exposed some issue on Windows. I am investigating the errors.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9672507

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9672506

mandolaerik (approved):

zhang-peng-j:
> DMLC_CFLAGS

zhang-peng-j:
> Would be nice to also have an env variable to select -O0 vs -O2. For DMLC_PYTHON we solved this by expressing the env var using a json representation, `DMLC_PYTHON='["/usr/bin/python", "-X", "utf8"]'`. We could do the same for DMLC_CC, or alternatively we could add a separate space separated variable `DMLC_CFLAGS`. If we choose the latter, then it can be done later as a separate patch.

I prefer DMLC_CFLAGS. Will do it in a separate patch.

zhang-peng-j:
> Ok. I will introduce DMLC_CC. The changes passed on Linux. But seems it exposed some issue on Windows. I am investigating the errors.

The unit tests failure on Windows host are caused by the fact that win64\\sys\\win64\\mingw-root-native is not packaged and thus does not exist in Simics Base installation directory. However, it is referenced in dotest_impl.py.

I workaround the issue. Please review the changes. Thank you.

test/tests.py
```diff
@@ -85,24 +90,33 @@ def dmlc_reaper_args(exitcode_file, timeout_multiplier = 1):
...
 common_cflags = ["-O2", "-std=gnu99", '-Wall', '-Werror', '-Wpointer-arith',
                  '-Wwrite-strings', '-Wformat-nonliteral',]
+cc = [os.environ.get('DMLC_CC')]
 if is_windows():
-    cc = [testparams.mingw_cc()]
+    if not cc[0]:
+        try:
+            cc = [testparams.mingw_cc()]
+            cc_not_found = not os.path.exists(cc[0])
+        except FileNotFoundError:
```
mandolaerik, line 99: insufficient. Looking at the implementation, seems that mingw_cc can throw a bare Exception. And that it cannot return None.

test/tests.py
```diff
@@ -85,24 +90,33 @@ def dmlc_reaper_args(exitcode_file, timeout_multiplier = 1):
 
 common_cflags = ["-O2", "-std=gnu99", '-Wall', '-Werror', '-Wpointer-arith',
                  '-Wwrite-strings', '-Wformat-nonliteral',]
+cc = [os.environ.get('DMLC_CC')]
```
mandolaerik, line 93: easier to not embrace cc in a list until we are done inspecting the string

zhang-peng-j:
I hope it looks better now :-)

test/tests.py
```diff
@@ -85,24 +90,33 @@ def dmlc_reaper_args(exitcode_file, timeout_multiplier = 1):
 
 common_cflags = ["-O2", "-std=gnu99", '-Wall', '-Werror', '-Wpointer-arith',
                  '-Wwrite-strings', '-Wformat-nonliteral',]
+cc = os.environ.get('DMLC_CC')
 if is_windows():
-    cc = [testparams.mingw_cc()]
+    if not cc:
+        try:
+            cc = testparams.mingw_cc()
+            cc_not_found = not os.path.exists(cc)
```
mandolaerik, line 98: Please normalize logic to avoid double negations, i.e. `cc_found = os.path.exists(cc)`

test/tests.py
```diff
@@ -85,24 +90,33 @@ def dmlc_reaper_args(exitcode_file, timeout_multiplier = 1):
...
+        except Exception:
+            cc_not_found = True
+        if cc_not_found:
+            raise TestFail('gcc not found(specify gcc by env var DMLC_CC)')
+        ldflags = [f"-L{testparams.mingw_root()}/lib/gcc/x86_64-w64-mingw32/lib"]
+    else:
+        ldflags = []
```
mandolaerik, line 105: Looks strange that the -L is needed in one branch but not the other, but if it works without the -L flag in your env, then I'm happy for now.

test/tests.py
zhang-peng-j: Ok. Fixed.

test/tests.py
zhang-peng-j: I don't need -L in my env. Let's leave ldflags as [] for now. In long term, I may need to parameterize ldflags, as an env var,  for test purpose.

mandolaerik (approved):

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9677131

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9677130

PR #46, opened 2022-01-28
Allocate trait vtable instances on the heap -- quick-fix for SIMICS-18498, SIMICS-18640, SIMICS-18596

lwaern-intel:
SSM has encountered issues multiple times relating to .bss being too large due to large arrays of trait vtable instances. Properly addressing this issue requires optimizing the generation and use of vtables such that vtable arrays will no longer be necessary, but the difficulty of implementing that solution justifies implementing a simpler fix in the meantime -- simply allocate trait vtable instances on the heap.

I'm unsure if t126 exhaustively covers every case this change would affect, so this should probably be tested against VP.

### DIFF
--- a/py/dml/c_backend.py
+++ b/py/dml/c_backend.py
@@ -1629,6 +1629,7 @@ def generate_init(device, initcode, outprefix):
             ');\n')
         out('}\n', preindent = -1)
 
+    out('_allocate_traits();\n')
     out('_initialize_traits();\n')
     out('_initialize_identity_ht();\n')
     out('_register_events(class);\n')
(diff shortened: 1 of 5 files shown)


syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9685223

py/dml/c_backend.py
```diff
@@ -1691,26 +1693,38 @@ def eom_linemark(site):
...
+            offset = 'sizeof(struct _%s)' % (cident(ancestry_path[0].name),)
+            items.append("{(uintptr_t)%s, %d, %s, %d}\n"
+                         % (base, num, offset, uniq))
+        add_variable_declaration(f'const _vtable_list_t *{ident}')
+        tables.append((ident, items))
+
+    start_function_definition('void _initialize_each_in_tables(void)')
+    out('{\n', postindent=1)
+    for (ident, items) in tables:
+        out(f'{ident} = MM_MALLOC({len(items)}, _vtable_list_t);\n')
```
mandolaerik, line 1722: Does this really solve a problem? If you allocate mem and fill it with constants, then those constants still has to live somewhere; it doesn't get smaller than if you just put the whole initializer in bss.

And if huge size is an issue here, then the generated function will be huge, which is a problem for compile speed. We have previously seen that the time consumption of some of gcc's optimizations tend to be superlinear in function size (this is why initialize_traits was split into initialize_traitsN). Parsing initializers is instead extremely fast.

lwaern-intel: The changes to the `each in` tables was because they now need to be initialized dynamically as the vtables memory addresses are no longer constant -- and at that point, I figured "why not also allocate them on the heap?"

I don't think the each in tables are a bottleneck, but if you think its best, I can implement code splitting up initialization into multiple functions just like we've done with trait vtables.

mandolaerik: Right, at least it's not subject to array explosion. Would it make sense to represent it as an array literal (wrapped in a struct), something like
```
void init_each_in_tables() {
  memcpy(x, &(const struct { _vtable_list_t _[N]; }){{{...}, {...}, ...}}, sizeof(x));
}
```
? or would this be too heavy on the stack?

lwaern-intel: I'd say so. Naively, my current version should be as bad, as compound literals have the lifespan of the block, although I'm fairly sure GCC optimizes my current solution away. I shouldn't take the risk, though, and instead wrap each assignment in a block.

lwaern-intel: Assignments updated to be wrapped in blocks. Question still remains if initialization should be split into multiple functions, from a discussion with Erik, we decided that should be decided from profiling.

lwaern-intel: Profiling shows heavy regression on GCC compilation times on dml models with many large each-in tables due to dynamic initialization. Alternate solution implemented which allows each-in tables to keep being initialized statically.

mandolaerik (approved):

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9692381

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9692380

PR #50, opened 2022-02-07
Rearrange device struct: propagate array dimensions to leaf members

mandolaerik:


lwaern-intel (approved):
Should only be merged after https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/1261.

PR #52, opened 2022-02-10
Independent methods -- SIMICS-6181, SIMICS-6182, SIMICS-17632

lwaern-intel:
Work remaining:

- [x] Implement logic to ban independent method calls in definitions of typed params
- [x] Documentation and release notes
- [x] Figure out way to clean up various hacks -- in particular, try to find a more elegant solution to SIMICS-6182
- [x] Touch-up pass (e.g. remove stray obsolete TODOs, etc)
- [x] More thorough idempotency tests (test `throws`)
- [x] Test against VP


syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9718969

lwaern-intel:
Jenkins job: https://cbjenkins-fm01.devtools.intel.com/teams-sre-vp/job/staging/job/testing/213/
No indication that the ESTATICVIOL check breaks anything. Will need to be tested again later once SSM corrects the code broken by #50.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9744632

syssimics:
TeamCity build completed (FAILURE): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9744631

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9744942

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9744941

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9782119

mandolaerik:
We discussed a rename of `idempotent` to `memoized`.

lwaern-intel:
> We discussed a rename of `idempotent` to `memoized`.

In progress.

lwaern-intel:
Did another VP test, however turned out `LAST_GOOD` is no good -- it ended up being as outdated as the last test, and filled with errors due to newer changes. I'll run yet another one on latest.

syssimics:
Verification [#9794325](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9794325): pass

syssimics:
Verification [#9794926](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9794926): fail

lwaern-intel:
After testing against VP again, still no indication of forbidding partially const outputs or `EINDEPENDENTVIOL` breaking anything. I'm ticking the box.

lib/1.4/dml-builtins.dml
```diff
@@ -492,6 +515,8 @@ The `device` template contains the following parameters:
 */
 template device {
     is object;
+    // _model_init is automatically called on model load
```
mandolaerik, line 518: move comment to `_model_init` definition?

lib/1.4/dml-builtins.dml
```diff
@@ -519,8 +544,10 @@ template device {
     // this carried semantics in DML 1.2; deprecated in 1.4
     param _confidentiality          = undefined;
 
+    independent method _model_init() { _rec_model_init(); }
     method _init() { _rec_init(); }
     method _post_init() { _rec_post_init(); }
+    independent method model_init() default {}
     method init() default {}
     method post_init() default {}
```
mandolaerik, line 552: these should be `shared`?

lib/1.4/dml-builtins.dml
lwaern-intel: I commented this here for symmetry with the comments re. `_init` and `_post_init` immediately below. Although you could argue they should be moved as well.

doc/1.4/language.md
```diff
@@ -1615,6 +1609,55 @@ and inline methods remain mainly for compatibility reasons.
...
+Memoized methods have the restriction that they may not take any input
+parameters. If a memoized method call throws, then subsequent calls will throw
+without executing the body.
+
+Results are cached per device instance, except for memoized methods that are
+also declared independent, in which case result caching is shared across all
+device instances. This mechanism can be used to compute device-independent data
+which is then shared across all device instances of the model.
+
+For `shared` memoized methods, result caching is done per-instance, and is not
```
mandolaerik, line 1654: You say independent memoized are shared across instances, whereas shared memoized are not. Then what about shared independent memoized methods? I would assume "shared across instances" but I can't derive that clearly from the docs.

test/1.4/errors/T_ESYNTAX_method_qualifiers.dml
```diff
@@ -0,0 +1,36 @@
...
+
+/// ERROR ESYNTAX
+independent independent method c() {}
+/// ERROR ESYNTAX
+memoized memoized method d() {}
+/// ERROR ESYNTAX
+memoized independent memoized method e() {}
+/// ERROR ESYNTAX
+independent memoized independent method f() {}
+
```
mandolaerik, line 20: Add cases for `inline independent method` and `inline memoized method` ? I expect those to fail (all nullary inline methods are banned already, but that happens post-parse so the ESYNTAX should win).

test/1.4/methods/T_idempotent.dml
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    #foreach m in (two_params_nt) {
+        try {
+            (i, j) = m();
+        } catch {
+            assert false;
+        }
+        assert fired && i == 4 && j == 7.2;
+        i = j = 0;
```
mandolaerik, line 247: if you say `i=j=1` instead, then you also test that they aren't accidentally cleared.

test/1.4/methods/T_independent.dml
```diff
@@ -0,0 +1,109 @@
+/*
+  © 2022 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+dml 1.4;
+device test;
+import "testing.dml";
+
+extern int sprintf(char *str, const char *format, ...);
```
mandolaerik, line 9: You can reduce pain a bit by using `strbuf_t`, which is already DML-wrapped. (or you could abandon strings and just use integers everywhere, like `return default() + 100;`)

test/1.4/methods/T_independent.dml
```diff
@@ -0,0 +1,109 @@
...
+}
+
+is u;
+
+
+method test() throws {
+    local char buf[50];
+    M(buf);
+    /// GREP \[obj info\] overridden M;default M
+    log info: "%s", buf;
```
mandolaerik, line 92: I'd say `assert strcmp(...) == 0` is more concise. `/// GREP` greps all of stdout.

test/1.4/methods/T_idempotent.dml
```diff
@@ -0,0 +1,257 @@
...
+dml 1.4;
+device test;
+import "testing.dml";
+
+header %{
+    static bool fired = false;
+%}
+extern bool fired;
+
+memoized method idemp_0() {
```
mandolaerik, line 14: `s/idemp/memo/` in this file?

doc/1.4/language.md
lwaern-intel: This is a case of two different meanings of "instance"; independent memoized are shared across _device_ instances. shared memoized are never shared across _template_ instances.
I'll try to make this more clear in the docs.

test/1.4/methods/T_idempotent.dml
lwaern-intel: yup

test/1.4/methods/T_independent.dml
lwaern-intel: The GREP thing was stolen from another file with very similar structure. `strcmp` is the better choice, though, I agree.

test/1.4/methods/T_idempotent.dml
lwaern-intel: also rename the file from idempotent to memoized

lib/1.4/dml-builtins.dml
lwaern-intel: i dunno, I'm not the one who wrote it. It really doesn't matter, it's not like the device template will be instantiated by more than one object.

test/1.4/methods/T_independent.dml
mandolaerik: I think most GREP tests are ancient. I use it with caution in newer tests, but sometimes it does provide a convenient way to check that a hook is called.

lib/1.4/dml-builtins.dml
mandolaerik: There is also a slight semantic difference: `template x is device { shared independent method model_init() {} }` is permitted iff the method has the `shared` annotation. I agree that this does not make a difference in practice.

py/dml/crep.py
```diff
@@ -19,8 +19,52 @@
...
+
+class TypedParamContext(StaticContext):
+    active = False
+
+    def __enter__(self):
+        StaticContext.__enter__(self)
+        self.prev_typed_context = TypedParamContext.active
+        TypedParamContext.active = True
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        StaticContext.__exit__(self, exc_type, exc_val, exc_tb)
```
mandolaerik, line 48: nitpick: `__enter__` and `__exit__` have return values so `return super().__exit__(...)` is idiomatic (would have made a difference if the superclass was further away than ten lines, but it's not so ok to ignore)

py/dml/crep.py
```diff
@@ -19,8 +19,52 @@
...
+def dev(site):
+    require_dev(site)
+    return "_dev"
+
+def require_dev(site):
+    if StaticContext.active:
+        raise EINDEPENDENTVIOL(site)
+
+def maybe_dev(independent):
+    return [] if independent else ['_dev']
```
mandolaerik, line 60: `else [dev(site)]`? The function name suggests that responsibilities are similar to `dev()`.

py/dml/crep.py
```diff
@@ -19,8 +19,52 @@
...
+
+def maybe_dev_arg(independent):
+    return ([] if independent
+            else [('_dev', TDevice(structtype(dml.globals.device)))])
```
mandolaerik, line 64: Similar to maybe_dev, unprotected `'_dev'`.

py/dml/crep.py
```diff
@@ -19,8 +19,52 @@
...
+    'require_dev',
+    'maybe_dev',
+    'maybe_dev_arg',
+    'StaticContext',
+    'TypedParamContext',
     )
 
+
+class StaticContext:
+    active = False
```
mandolaerik, line 32: I would find it more natural to invert the default, i.e. we evaluate in static context by default, and enter "instance context" when evaluating a method body. I.e., if we enter a context where an instance is known, then that makes `dev()` return a reference to that instance. Perhaps even softcode the name, `with InstanceContext('_dev')`, to eliminate the need for `maybe_dev_arg`.

py/dml/crep.py
lwaern-intel: No, it's different. The name should be changed, not a `require_dev` check -- the latter would break it.

lwaern-intel: Actually, I have a mind to scrap this function and replace calls to it with `['_dev'] * independent` -- it's about as clean as a function call.
Still, that can't be done with `maybe_dev_arg`, and I can't for the life of me figure out how to rename it.

py/dml/crep.py
lwaern-intel: That'll be a pain. I'll look into it, but I'd like to keep the static context approach if it proves too difficult.

test/1.4/errors/T_ESYNTAX_method_qualifiers.dml
lwaern-intel: `inline independent method` and `inline memoized method` wouldn't parse. So at the very least, they'd need their own dedicated test, and I don't see why they should if e.g. `shared inline method` doesn't have one.

test/1.4/methods/T_independent.dml
lwaern-intel: I don't see the need; I'll keep this as-is.

test/1.4/methods/T_idempotent.dml
lwaern-intel: Good idea.

include/simics/dmllib.h
```diff
@@ -2386,6 +2387,27 @@ UNUSED static void *_get_shared_memoized_outs(
...
+    ASSERT(error == 0);
+    error = pthread_mutexattr_settype(&attributes, PTHREAD_MUTEX_RECURSIVE);
+    ASSERT(error == 0);
+    error = pthread_mutex_init(lock, &attributes);
+    ASSERT(error == 0);
+    error = pthread_mutexattr_destroy(&attributes);
+    ASSERT(error == 0);
+}
+
+UNUSED static void _independent_memoized_lock(
```
mandolaerik, line 2402: Threading is hard -- please document invariants thoroughly.

Looks like this one either returns with *ran==1 and no lock taken, or with *ran!=1 and the lock taken? Would make sense to return `*ran==1`?

py/dml/codegen.py
```diff
@@ -242,35 +242,46 @@ def codegen_exit(self, site, retvals):
         assert retvals is not None, 'dml 1.2/1.4 mixup'
         return codegen_return(site, self.outp, self.throws, retvals)
 
+
```
mandolaerik, line 245: double newline

py/dml/crep.py
lwaern-intel: Partly done. I've done a straight-up inversion (`StaticContext` to `DeviceInstanceContext´), but nothing else, yet -- I'm confused about the rest of your comment, and want to talk about it. In the meantime, the conversion is messy and likely fragile, so it warrants an additional test against VP.

syssimics:
DML [#9835856](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9835856): fail

syssimics:
Verification [#9835855](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9835855): fail

syssimics:
DML [#9837813](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9837813): fail

syssimics:
Verification [#9837812](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9837812): fail

syssimics:
Verification [#9838031](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9838031): fail

syssimics:
Verification [#9842789](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9842789): pass

test/tests.py
```diff
@@ -648,7 +648,7 @@ class options(object):
                 self.cfilename + ".o",
                 module_id_base + ".o",
                 "-L"+join(host_path(), "bin", "py3")] + ldflags + \
-                ["-lsimics-common", "-lvtutils"]
+                ["-lsimics-common", "-lvtutils", "-lpthread"]
```
mandolaerik, line 651: I realized that `-lpthread` poses some problems: looks like the build system doesn't pass -lpthread by default (added to our local simics-base build env by `config/env.mk:35`). I don't understand why this doesn't cause t1000 to fail (perhaps the linker resolves pthread as a transitive dependency?).
Unsure what to do about this dependency; perhaps we could let -lvtutils or -lsimics-common re-export trampolines to pthread? or just add -lpthread as an explicit dep to all modules?

In any case, if we add a -lpthread dep, then it's a source of errors, so it deserves a releasenote entry even if we solve it in standard makefiles.

doc/1.4/language.md
```diff
@@ -1644,6 +1638,58 @@ and inline methods remain mainly for compatibility reasons.
...
+be declared `independent`:
+```
+independent method m(...) -> (...) {...}
+```
+[Exported](#export-declarations) independent methods do not have the input
+parameter corresponding to the device instance, allowing them to be called
+in greater number of contexts. The body of independent methods may not contain
+statements or expressions that rely on the device instance in any way; for
+example, `session` or `saved` variables may not be referenced, `after` and `log`
+statements may not be used, and non-`independent` may not be called.
```
lwaern-intel, line 1661: `and non-`independent` may not be called.` should be `and non-`independent` methods may not be called.`

syssimics:
DML [#9884430](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9884430): fail

syssimics:
Verification [#9884429](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9884429): pass

lwaern-intel:
Moved to #81

PR #54, opened 2022-02-15
Update docs re. `in each` in #if; true/false in experimental #if -- SIMICS-18702

lwaern-intel:
HSD-ES ticket 18020519096

### DIFF
--- a/doc/1.4/language.md
+++ b/doc/1.4/language.md
@@ -1982,14 +1982,12 @@ The *conditional* is an expression with a constant boolean value.  It
 may reference parameters declared at the same level in the object
 hierarchy, or in parent levels.
 
-The *object declarations* are any number of declarations of
-objects, session variables, saved variables, methods,
-`in each` statements or other `#if` statements, but not
-parameters or `is` statements.  When the conditional is `true`
-(or if it's the else branch of a false conditional), the object
-declarations are treated as if they had appeared without any
-surrounding *#if*.  So the two following declarations are
-equivalent:
+The *object declarations* are any number of declarations of objects, session
+variables, saved variables, methods, or other `#if` statements, but not
+parameters, `is` statements, or `in each` statements . When the conditional is
+`true` (or if it's the else branch of a false conditional), the object
+declarations are treated as if they had appeared without any surrounding *#if*.
+So the two following declarations are equivalent:
 
 ```
 #if (true) {
--- a/py/dml/toplevel.py
+++ b/py/dml/toplevel.py
@@ -140,11 +140,18 @@ def scan_statements(filename, site, stmts):
         elif s.kind == 'toplevel_if':
             [cond, tbranch, fbranch] = s.args
             scope = symtab.Symtab()
-            bsite = SimpleSite('<builtin>', dml_version=dml.globals.dml_version)
-            scope.add(ctree.ExpressionSymbol(
+            bsite = SimpleSite('<builtin>',
+                               dml_version=dml.globals.dml_version)
+            # HACK Add constants to scope typically defined by dml-builtins,
+            # which is not accessible here
+            def add_constant(name, expr):
+                scope.add(ctree.ExpressionSymbol(name, expr, bsite))
+
+            add_constant(
                 'dml_1_2',
-                ctree.BoolConstant(bsite, dml.globals.dml_version == (1, 2)),
-                bsite))
+                ctree.BoolConstant(bsite, dml.globals.dml_version == (1, 2)))
+            add_constant('true', ctree.BoolConstant(bsite, True))
+            add_constant('false', ctree.BoolConstant(bsite, False))
             try:
                 expr = ctree.as_bool(codegen.codegen_expression(
                     cond, None, scope))
--- a/test/1.4/errors/T_ECONDINEACH.dml
+++ b/test/1.4/errors/T_ECONDINEACH.dml
@@ -21,3 +21,10 @@ port p {
         in each x {}
     }
 }
+
+/// WARNING WEXPERIMENTAL
+#if (false && true && dml_1_4) {
+    // Experimental support for complex top-level conditionals allows in-each
+    // to be used
+    in each x {}
+}
--- a/py/dml/dmlparse.py
+++ b/py/dml/dmlparse.py
@@ -249,7 +249,8 @@ def toplevel_if(t):
     if all(stmt.kind in allowed_in_hashif for stmt in t[6] + t[8]):
         t[0] = ast.hashif(site(t), t[3], t[6], t[8])
     else:
-        report(WEXPERIMENTAL(site(t), "'if' body with top-level statements"))
+        report(WEXPERIMENTAL(site(t), ("top-level 'if' body with unsupported "
+                                       + "statements")))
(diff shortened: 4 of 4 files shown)


lwaern-intel:
@JonatanWaern With Erik on vacation, you will be the primary target of being pestered for reviews!

py/dml/toplevel.py
```diff
@@ -140,11 +140,24 @@ def scan_statements(filename, site, stmts):
         elif s.kind == 'toplevel_if':
             [cond, tbranch, fbranch] = s.args
             scope = symtab.Symtab()
-            bsite = SimpleSite('<builtin>', dml_version=dml.globals.dml_version)
-            scope.add(ctree.ExpressionSymbol(
+            bsite = SimpleSite('<builtin>',
+                               dml_version=dml.globals.dml_version)
+            # HACK Add constants to scope typically defined by dml-builtins,
```
JonatanWaern, line 145: Ew, but not much to do about this I assume?

JonatanWaern (approved):

py/dml/toplevel.py
lwaern-intel: Well, it's probably possible to eliminate the hack, but it would require more intricate changes which I didn't want to waste my time looking into as the current implementation of experimental `#if`s is very unlikely to remain, anyway.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9741638

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9741637

py/dml/toplevel.py
mandolaerik: i dont think these added constants have a use case. isn't it easier to read/remember/explain 'dml_1_2 is the only permitted symbol' instead of 'the only allowed symbols are dml_1_2 plus some random useless symbols'?

test/1.4/errors/T_ECONDINEACH.dml
```diff
@@ -21,3 +21,10 @@ port p {
         in each x {}
     }
 }
+
+/// WARNING WEXPERIMENTAL
+#if (false) {
```
mandolaerik, line 26: `!dml_1_2`?

doc/1.4/language.md
```diff
@@ -1982,14 +1982,12 @@ The *conditional* is an expression with a constant boolean value.  It
...
-objects, session variables, saved variables, methods,
-`in each` statements or other `#if` statements, but not
-parameters or `is` statements.  When the conditional is `true`
-(or if it's the else branch of a false conditional), the object
-declarations are treated as if they had appeared without any
-surrounding *#if*.  So the two following declarations are
-equivalent:
+The *object declarations* are any number of declarations of objects, session
+variables, saved variables, methods, or other `#if` statements, but not
+parameters, `is` statements, or `in each` statements . When the conditional is
```
mandolaerik, line 1987: s/is/evaluates to?

doc/1.4/language.md
```diff
@@ -1982,14 +1982,12 @@ The *conditional* is an expression with a constant boolean value.  It
...
+`true` (or if it's the else branch of a false conditional), the object
+declarations are treated as if they had appeared without any surrounding *#if*.
```
mandolaerik, line 1989: ...exceot for the subtle difference for session/saved in a template. this should be mentioned somewhere, would it be too detailed to say it here?

py/dml/toplevel.py
lwaern-intel: `unmapped_offset` I can understand, but I certainly would not call `true` and `false` "random useless symbols" -- people will get understandably confused if they see an error message telling them that "false" is an unknown identifier. Adding `dml_1_4` is nice for symmetry if we already support `dml_1_2`.

doc/1.4/language.md
lwaern-intel: In my mind people will grok that "is" means "evaluates to", but the latter wording is better. I suppose I could change that too, while I'm at it.

doc/1.4/language.md
lwaern-intel: I say we talk about it in the template-as-types section, perhaps in a new subsection. We could present the `#if (true)` workaround and explain the cases where it is necessary.

test/1.4/errors/T_ECONDINEACH.dml
lwaern-intel: I see no reason why?

doc/1.4/language.md
lwaern-intel: However, that is all unrelated to this PR.

py/dml/toplevel.py
mandolaerik: what's the use case for `#if (false)`?
dml_1_4 is meant for distinguishing between 1.4 and 1.5 in code shared between the two versions; this use case doesn't exist yet, we just reserved the ident.

lwaern-intel: If nothing else, `#if (false)` is a nice way to comment out code; and for that reason alone people _will_ reach for it, in which case we don't want to say "unknown identifier: `false`", because that will only serve to confuse people. Just see the HSD-ES ticket.

lwaern-intel: Oh right, sorry, you're on vacation; I keep forgetting this repo is public :)

mandolaerik: right, that's a fair use of `false`. however, what would truly reduce confusion is an error message that clarifies stuff, like 'only [false and] dml_1_2 allowed as conditions of #if blocks containing %s'. hard to write that short with a long list of idents.

doc/1.4/language.md
mandolaerik: fair enough

py/dml/toplevel.py
lwaern-intel: We don't write that down regardless. Besides, despite `false` and `true` being identifiers in DML 1.4 people typically view them as literals instead, so perhaps we don't need to explicitly say they may be used. I'm more than fine with scrapping `unmapped_offset` (I only added it because it was a `constant`, rather than a `param`) and while I won't protest to scrapping `dml_1_4` as well, doing so feels asymmetrical and weird.

mandolaerik: the report mentions 'if false' but that's likely just a minimal reproducer. the original use case was likely a param. I don't think their debug confusion would be any smaller if the behaviour further changed when moving from a false param to false. Just changes the reproducer.

hyrum requires you to write test cases for all allowed idents and those test cases will be weird. (dml_1_2 is naturally smoke tested because it has strong use cases).

the dml_1_2 hack is hideous and by nature asymmetrical; to me, extending it with other arbitrary idents seems to increase that asymmetry. the current pattern is that any idents absolutely needed for survival are included.

mandolaerik: that said, i'd say the change is neither significantly harmful nor useful, so i don't object to merge if you insist (but test is needed for all idents). just wanted to explain my reasoning.

lwaern-intel: I insist at least for `true` and `false`. `dml_1_4` can be scrapped.

PR #55, opened 2022-02-15
Support break inside DML 1.4 foreach statements -- SIMICS-18719

lwaern-intel:


py/dml/codegen.py
```diff
@@ -121,7 +121,21 @@ def __init__(self):
 
     def break_(self, site):
         self.used = True
-        return [mkUnrolledBreak(site, self.label)]
+        return [mkGotoBreak(site, self.label)]
+
+class SequenceLoopContext(LoopContext):
```
mandolaerik, line 126: unify context classes to, say, GotoLoopContext?

lwaern-intel: They are so similar, in fact, I oopsied when I wrote `__init__` of `SequenceLoopContext`. I agree, they should be unified, but I don't really like the name `GotoLoopContext`.

lwaern-intel: Pushed up a commit unifying them with the name `AbstractLoopContext` -- it's placeholder, I don't like it much more either. It's supposed to be "abstract" in the sense "a loop not corresponding to a C loop", but the name doesn't really communicate that.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9744625

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9744624

py/dml/codegen.py
mandolaerik: `Abstract*` is usually an abstract class.

loop context is a thing that offers break. this class does that by gotoing to a label that the owner is responsible to provide. so i vote for a name with goto in.

lwaern-intel: Alright. `GotoLoopContext` it is.

mandolaerik (approved):

PR #59, opened 2022-02-24
Defensive deserialization of values of serializable types -- SIMICS-18522

lwaern-intel:


syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9766346

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9766345

include/simics/dmllib.h
```diff
@@ -2113,8 +2113,10 @@ _set_device_member(attr_value_t val,
                             &dimension_strides[1],
                             remaining_dimensions-1,
                             setter);
-                        // currently, generated deserialized cannot fail
-                        ASSERT(status == Sim_Set_Ok);
+                        // If deserialization fails, bail out.
+                        if (unlikely(status != Sim_Set_Ok)) {
```
mandolaerik, line 2117: Ideally, a failed setter should have no effect, i.e. we should split _set_device_member in two stages, a side-effect free `is_legal()` plus a never failing `set()`. However, this is not a strict requirement since we already break this rule in arrays of attributes or connects.

include/simics/dmllib.h
```diff
@@ -2113,8 +2113,10 @@ _set_device_member(attr_value_t val,
...
+                            return status;
```
mandolaerik, line 2118: this path should be covered by a test?

mandolaerik (approved):

PR #61, opened 2022-03-01
Restore the [i < ...] array syntax.

lwaern-intel:
This reverts commit fa5e26926729d035f54675b4beeb1d2a8f8a1591.

The discussion has been settled: the conclusion is that the workaround for Konstantin's use-case is acceptable. This is due to the rarity of the use-case, together with fact that other solutions to do not significantly improve that use-case, and that those other improvements would be very vulnerable to changes in the model.


mandolaerik (approved):

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9778482

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9778481

PR #63, opened 2022-03-02
Add environment variable DMLC_DUMP_INPUT_FILES

mandolaerik:
- Repair and run test
- Provide a sensible default for api_version
- Simplify profiling test case
- Add environment variable DMLC_DUMP_INPUT_FILES

### DIFF
--- a/py/dml/dmlc.py
+++ b/py/dml/dmlc.py
@@ -172,10 +172,13 @@ def dump_input_files(outputbase, imported):
     max_dotdot = max(dotdot_depth(p)
                      for (_, paths) in imported.items()
                      for p in paths)
+    # HACK: if any file is imported via ../, then store all DML files
+    # sufficiently deep in a directory hierarchy _/_/.../, so that the
+    # .. path can be resolved within the archive.
     prefix = '/'.join(['_'] * max_dotdot)
     basenames = set()
     with tarfile.open(outputbase + ".tar.xz", 'w:xz') as tf:
-        # if two different files foo.dml are imported, where one is
+        # HACK: if two different files foo.dml are imported, where one is
         # imported as "foo.dml" or "./foo.dml" and the other is only
         # imported using a qualified path such as "bar/foo.dml", then
         # the rename hack below has a better chance to work if we process
@@ -184,7 +187,7 @@ def dump_input_files(outputbase, imported):
             paths = imported[f]
             base = os.path.basename(f)
             while base in basenames:
-                # all DML files end up in the same directory; if two
+                # HACK: all DML files end up in the same directory; if two
                 # have the same name, then dodge that by renaming. In
                 # this case we may need to change some import directive
                 # manually in order for all files to be imported.
@@ -201,7 +204,8 @@ def dump_input_files(outputbase, imported):
                 if os.path.dirname(symlink) == os.path.normpath(prefix):
                     # import path resolves to file already present in tarfile
                     continue
-                # file imported through relative path, create symlink there
+                # file imported through relative path, create symlink where
+                # the relative path is resolved
                 ti = tarfile.TarInfo(symlink)
                 ti.type = tarfile.SYMTYPE
                 ti.linkname = os.path.normpath(os.path.join(
@@ -560,7 +564,7 @@ def main(argv):
             print_cdep(outputbase, headers, footers)
             if logging.failure:
                 sys.exit(2)
-            deplist = [inputfilename] + list(imported)
+            deplist = [inputfilename] + list(imported.keys())
             now = time.time()
             future_timestamps = [path for path in deplist
                                  if os.stat(path).st_mtime > now]
@@ -604,7 +608,7 @@ def main(argv):
 
         if output_c:
             dml.c_backend.generate(dev, headers, footers, outputbase,
-                                   [inputfilename] + list(imported),
+                                   [inputfilename] + list(imported.keys()),
                                    options.full_module)
             logtime("c")
             structure.check_unused_and_warn(dev)
--- a/py/dml/dmlc.py
+++ b/py/dml/dmlc.py
@@ -20,6 +20,7 @@ import dml.dmlparse
 from .logging import *
 from .messages import *
 from .env import api_versions, default_api_version
+import tarfile
 
 def prerr(msg):
     sys.stderr.write(msg + "\n")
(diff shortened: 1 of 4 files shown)


py/dml/dmlc.py
```diff
@@ -153,6 +154,60 @@ def print_cdep(outputbase, headers, footers):
     f.close()
     f.commit()
 
+def dotdot_depth(path):
+    '''return how many levels of parent directories are required to resolve a
+    relative path'''
+    return os.path.normpath(path).count('..' + os.path.sep)
+
+assert dotdot_depth('./../../foo/../../../bar.dml') == 4
```
lwaern-intel, line 162: This is weird. Is this really the right place for this test?

py/dml/dmlc.py
```diff
@@ -153,6 +154,60 @@ def print_cdep(outputbase, headers, footers):
...
+                     for p in paths)
+    prefix = '/'.join(['_'] * max_dotdot)
+    basenames = set()
+    with tarfile.open(outputbase + ".tar.xz", 'w:xz') as tf:
+        # if two different files foo.dml are imported, where one is
+        # imported as "foo.dml" or "./foo.dml" and the other is only
+        # imported using a qualified path such as "bar/foo.dml", then
+        # the rename hack below has a better chance to work if we process
+        # the unqualified paths first.
+        for f in sorted(imported, key=lambda f: min(map(len, imported[f]))):
```
lwaern-intel, line 183: This is a very hacky hack. I'd like to see a better approach. If you can't think of one, add a `HACK` annotation.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9780808

py/dml/dmlc.py
```diff
@@ -498,11 +553,14 @@ def show_illegal_attributes(option, opt, value, parser):
...
 
+        if 'DMLC_DUMP_INPUT_FILES' in os.environ:
+            dump_input_files(outputbase, dict(
+                imported, **{inputfilename: [os.path.basename(inputfilename)]}))
         if options.makedep:
             print_cdep(outputbase, headers, footers)
             if logging.failure:
                 sys.exit(2)
-            deplist = [inputfilename] + imported
+            deplist = [inputfilename] + list(imported)
```
lwaern-intel, line 563: `imported` is a dict, right? I'd do `list(imported.keys())` for clarity.

py/dml/dmlc.py
```diff
@@ -546,7 +604,7 @@ def show_illegal_attributes(option, opt, value, parser):
 
         if output_c:
             dml.c_backend.generate(dev, headers, footers, outputbase,
-                                   [inputfilename] + imported,
+                                   [inputfilename] + list(imported),
```
lwaern-intel, line 607: Same here.

test/tests.py
```diff
@@ -544,6 +543,19 @@ def verify_dmlc_messages(self, stderr, expected_msgs):
             self.expect_messages(
                 kind, actual_msgs.get(kind, []), expected_msgs.get(kind, []))
 
+def _unittest_parse_messages():
```
lwaern-intel, line 546: Why does it have a `_` prefix to begin with?

test/tests.py
```diff
@@ -544,6 +543,19 @@ def verify_dmlc_messages(self, stderr, expected_msgs):
...
+/tmp/f.dml:3:20: In template abc
+/tmp/g.dml:4:9: error EDPARAM: blah blah
+/tmp/f.dml:8:12: conflicting definition
+'''.splitlines()
+    assert (list(DMLFileTestCase.parse_messages(stderr))
+            == [('error', 'EDPARAM',
+                 [('g.dml', 4, "/tmp/g.dml:4:9: error EDPARAM: blah blah"),
+                  ('f.dml', 3, "/tmp/f.dml:3:20: In template abc"),
+                  ('f.dml', 8, "/tmp/f.dml:8:12: conflicting definition")])])
+_unittest_parse_messages()
```
lwaern-intel, line 557: Why not a proper test case?

lwaern-intel (changes requested):

py/dml/dmlc.py
```diff
@@ -153,6 +154,60 @@ def print_cdep(outputbase, headers, footers):
...
+        # the unqualified paths first.
+        for f in sorted(imported, key=lambda f: min(map(len, imported[f]))):
+            paths = imported[f]
+            base = os.path.basename(f)
+            while base in basenames:
+                # all DML files end up in the same directory; if two
+                # have the same name, then dodge that by renaming. In
+                # this case we may need to change some import directive
+                # manually in order for all files to be imported.
+                base = '_' + base
```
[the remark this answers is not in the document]
mandolaerik: Note also that it's unsafe and will give a non-compilable archive if `import "./foo.dml"` happens in two different directories. The idea is that this likely is uncommon, so we can handle it manually for now.
Solving that problem seems rather hard, and probably requires that you either pass extra -I flags when building, or let this python code edit imports while saving files. Likely not worth the complexity.

test/tests.py
mandolaerik: Test cases are for testing DMLC. This one tests the test.

test/tests.py
mandolaerik: I have no opinion, I can remove it if you prefer.

py/dml/dmlc.py
mandolaerik: My approach is inherently unsafe, so there is no natural clean solution, just a risk-minimizing heuristics. So I'll go for `# HACK` (the code probably deserves more than one such annotation).

py/dml/dmlc.py
mandolaerik: I sometimes keep extremely lightweight tests of pure functions as top-level assertions to avoid boilerplate. In particular, I think this is a good way to make top-level `re.compile()` calls more readable. And this function has similar weight as a regexp.

lwaern-intel: Hm, alright. I suppose it makes no difference, even if perhaps it's not "proper."

test/tests.py
lwaern-intel: Ah.

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9781146

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9781145

test/tests.py
mandolaerik: Done.

lwaern-intel (approved):
The hacky stuff really is very hacky, but I can live with it.

PR #64, opened 2022-03-02
Use nullcontext to avoid the overly complex ExitStack

mandolaerik:


syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9781896

py/dml/codegen.py
```diff
@@ -2375,11 +2375,10 @@ def codegen_inline(site, meth_node, indices, inargs, outargs,
     meth_node.refcount += 1        # regard method as used
 
     # Open the scope
-    with contextlib.ExitStack() as contexts:
-        contexts.enter_context(RecursiveInlineGuard(site, meth_node))
-        contexts.enter_context(ErrorContext(meth_node, site))
-        if not meth_node.throws:
-            contexts.enter_context(NoFailure(site))
+    # TODO: in python 3.10 we can use parentheses instead of \
```
lwaern-intel, line 2378: Remind me -- what python version _are_ we using?

lwaern-intel (approved):

syssimics:
TeamCity build completed (SUCCESS): https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9781895

py/dml/codegen.py
mandolaerik:
```
$ ls core/linux64/lib/
images  python3.7  python-py3
```
Henrik is working on migration to 3.9. Versioning is driven by [Intel Python](https://www.intel.com/content/www/us/en/developer/tools/oneapi/distribution-for-python.html) releases. For DMLC, we have the additional constraint that we want it to be pypy compatible, which doesn't work with newer versions than 3.8.

PR #71, opened 2022-03-10
Make map_target.issue 'default'

svenwest:
This allows implementers to override it, and add custom atoms to all transactions

### DIFF
--- a/lib/1.4/utility.dml
+++ b/lib/1.4/utility.dml
@@ -1365,7 +1365,7 @@ template map_target is (connect, _qname) {
        * `issue(transaction_t *t, uint64 addr) -> (exception_type_t)`
 
          Provides a shorthand to the API function `SIM_issue_transaction`.
-         This method is called by the read/write methods in this template, it
+         This method is called by the read/write methods in this template. It
          can be overridden, e.g. to add additional atoms to the transactions,
          while still allowing the ease-of-use from the simpler methods.
     */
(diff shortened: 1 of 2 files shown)


lib/1.4/utility.dml
```diff
@@ -1367,7 +1367,7 @@ template map_target is (connect, _qname) {
          Provides a shorthand to the API function `SIM_issue_transaction`.
```
mandolaerik, line 1367: I suppose the motivation is that you may want to override the method to add extra atoms or whatever, and still be able to access stuff conveniently through read/write methods.

In order for users to understand this, you need to update documentation, maybe say that it's called by read/write methods and can be overridden.

svenwest: Ok, but if users are going to override it they'll need to read the source - documentation won't be enough.

mandolaerik: Unless you document it properly. This is a documented library, not just common code, so as soon as it's released, the responsibilities of the method are fixed, whether or not you document them (due to Hyrum's law). It's best to document those responsibilities, to make them explicit both for us and for users.

svenwest: I don't fully agree. If someone is likely to read the source anyway, duplicating information easily digested from the source in documentation seems unnecessary and inexact.

Anyway, I'll improve the docs :-)

lib/1.4/utility.dml
```diff
@@ -1365,9 +1365,12 @@ template map_target is (connect, _qname) {
        * `issue(transaction_t *t, uint64 addr) -> (exception_type_t)`
 
          Provides a shorthand to the API function `SIM_issue_transaction`.
+         This method is called by the read/write methods in this template, it
```
svenwest, line 1368:
```suggestion
         This method is called by the read/write methods in this template. It
```

mandolaerik (approved):

PR #74, opened 2022-03-24
Fix bug in naming when init_as_subobj is instantiated in a bank or port

mandolaerik:
Reported on IM by Sven.

### DIFF
--- a/lib/1.4/dml-builtins.dml
+++ b/lib/1.4/dml-builtins.dml
@@ -1008,10 +1008,14 @@ template init_as_subobj is (connect, init, _qname) {
     param classname;
     param configuration default "none";
     method init() default {
-        local conf_object_t *o = SIM_object_descendant(dev.obj, this._qname());
+        local int portname_len = this._nongroup_parent.objtype == "device"
+            #? 0 #: strlen(this._nongroup_parent._qname()) + 1;
+        local conf_object_t *o = SIM_object_descendant(
+            this._nongroup_parent.obj, this._qname() + portname_len);
         if (o == NULL) {
             // should not happen
-            log critical: "subobject %s not found", this._qname();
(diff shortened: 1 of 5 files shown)


svenwest:
I haven't digested all of the python-yielding and recursion, but it looks good from far away :-)

syssimics:
Verification [#9838919](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9838919): pass

lib/1.4/dml-builtins.dml
```diff
@@ -1008,10 +1008,14 @@ template init_as_subobj is (connect, init, _qname) {
     param classname;
     param configuration default "none";
     method init() default {
-        local conf_object_t *o = SIM_object_descendant(dev.obj, this._qname());
+        local int portname_len = this._nongroup_parent.objtype == "device"
+            #? 0 #: strlen(this._nongroup_parent._qname()) + 1;
+        local conf_object_t *o = SIM_object_descendant(
+            this._nongroup_parent.obj, this._qname() + portname_len);
```
lwaern-intel, line 1014: `_qname()` won't be the same as `logname_anonymized`, which means this logic won't work if anonymization applies -- but I expect it won't as this is 1.4 exclusive. I just want to point that out since it's a subtle detail which might matter if I've missed something else.

lwaern-intel (approved):
Looks good.

lib/1.4/dml-builtins.dml
mandolaerik: There is no plan to support this is in 1.2 so this should be fine (and we make similar assumptions elsewhere)

syssimics:
Verification [#9841108](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9841108): pass

PR #76, opened 2022-03-29
Provide initiator in bank instrumentation

jhbaarnh:

PR #78, opened 2022-04-05
Pass -r flag to gcc directly

mandolaerik:
This fixes problems on Ubuntu 20.4, where gcc passes -pie to the linker
by default, which is incompatible with -r. Passing -r to gcc suppresses
this default, whereas -Wl,-r does not.

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -925,7 +925,7 @@ class SplitTestCase(CTestCase):
                 return result
         self.cc_stdout = join(self.scratchdir, self.shortname + '.cc_stdout')
         self.cc_stderr = join(self.scratchdir, self.shortname + '.cc_stderr')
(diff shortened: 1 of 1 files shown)


syssimics:
DML [#9875842](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9875842): fail

syssimics:
Verification [#9875841](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9875841): pass

lwaern-intel (approved):

mandolaerik:
@lwaern-intel what's the error in the DML build? false negative?

lwaern-intel:
> @lwaern-intel what's the error in the DML build? false negative?

Same error as the one in #76. We still haven't cut a release since then!

PR #80, opened 2022-04-07
Remove element restrictions of compile-time list expressions -- SIMICS-13113

lwaern-intel:


mandolaerik:
This is welcome, but I'd prefer to do it after #34 if possible, and only allow it for `#[]` lists, to avoid compatibility issues.

mandolaerik:
Blocked by #34

lwaern-intel:
@mandolaerik Now rebased to modern day. I want to try to _finally_ get this in.

syssimics:
Verification [#14350182](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14350182): ✅ pass

mandolaerik (approved):
We could also add a test of a non-constant element in a list, but I don't insist (would not increase coverage IIUC).

lwaern-intel:
> We could also add a test of a non-constant element in a list, but I don't insist (would not increase coverage IIUC).

Indeed it wouldn't, so I won't bother.

PR #81, opened 2022-04-11
Independent methods -- SIMICS-6181, SIMICS-6182, SIMICS-17632

lwaern-intel:
Continued from #52

### DIFF
--- a/include/simics/dmllib.h
+++ b/include/simics/dmllib.h
@@ -181,7 +181,13 @@ DML_eq(uint64 a, uint64 b)
                                                       __VA_ARGS__);})
 #define CALL_TRAIT_METHOD0(type, method, dev, traitref)                 \
         ({_traitref_t __tref = traitref;                                \
-          ((struct _ ## type *) __tref.trait)->method(dev, __tref);})   \
+          ((struct _ ## type *) __tref.trait)->method(dev, __tref);})
+#define CALL_INDEPENDENT_TRAIT_METHOD(type, method, traitref, ...)       \
+        ({_traitref_t __tref = traitref;                                      \
+          ((struct _ ## type *) __tref.trait)->method(__tref, __VA_ARGS__);})
+#define CALL_INDEPENDENT_TRAIT_METHOD0(type, method, traitref)     \
+    ({_traitref_t __tref = traitref;                                    \
+      ((struct _ ## type *) __tref.trait)->method(__tref);})
 
 #define _raw_load_uint8_be_t   UNALIGNED_LOAD_BE8
 #define _raw_load_uint16_be_t  UNALIGNED_LOAD_BE16
@@ -2411,6 +2417,14 @@ UNUSED static int _free_sub_table(ht_int_table_t *table,
 UNUSED static void _free_table(ht_int_table_t *table) {
         ht_for_each_entry_int(table, _free_sub_table, NULL);
         ht_clear_int_table(table, true);
+}
+
+UNUSED static void _memoized_recursion(const char *name) {
+    char msg[512];
+    snprintf(msg, 512,
+             "Recursive call to memoized method %s. "
+             "This is considered undefined behavior.", name);
+    VT_critical_error(msg, msg);
 }
 
 #endif
--- a/py/dml/c_backend.py
+++ b/py/dml/c_backend.py
@@ -5,9 +5,10 @@ import sys, os
 import itertools
 import operator
 import re
+from contextlib import nullcontext
 from functools import reduce
 
-from . import objects, logging, crep, output, ctree, serialize
+from . import objects, logging, crep, output, ctree, serialize, structure
 from . import traits
 import dml.globals
 from .logging import *
@@ -243,7 +244,8 @@ def generate_structfile(device, filename, outprefix):
             out("extern %s;\n"
                 % func.rettype.declaration(
                     "%s(%s)" % (name,
-                                ", ".join(["conf_object_t *_obj"]
+                                ", ".join((["conf_object_t *_obj"]
+                                           * (not func.independent))
                                           + [t.declaration(n)
                                              for (n, t) in func.cparams[1:]]))))
     emit_guard_end(filename)
@@ -319,6 +321,9 @@ def generate_hfile(device, headers, filename):
     out('\n')
 
     for (name, t) in list(dml.globals.traits.items()):
+        for method in t.method_impls.values():
+            if method.memoized:
+                method.memo_outs_struct.print_struct_definition()
         t.type().print_vtable_struct_declaration()
     out('\n')
 
@@ -401,7 +406,7 @@ def generate_attr_setter(fname, node, port, dimsizes, cprefix, loopvars,
         out('attr_value_t attr%d = %s;\n' % (dim, list_item))
         valuevar = 'attr%d' % (dim,)
 
-    with NoFailure(node.site):
+    with NoFailure(node.site), crep.DeviceInstanceContext():
         setcode = [
             codegen_inline_byname(
                 node, port_indices + loopvars,
@@ -464,7 +469,7 @@ def generate_attr_getter(fname, node, port, dimsizes, cprefix, loopvars):
         out('attr_value_t %s;\n' % (next_valuevar.read()))
         valuevar = next_valuevar
 
-    with NoFailure(node.site):
+    with NoFailure(node.site), crep.DeviceInstanceContext():
         getcode = codegen_inline_byname(
             node, port_indices + loopvars,
             '_get_attribute' if dml.globals.dml_version == (1, 2)
@@ -705,18 +710,19 @@ def wrap_method(meth, wrapper_name, indices=()):
         assert meth.dimensions == len(indices)
         out(devstruct+' *_dev UNUSED = ('+devstruct+'*)_obj;\n')
         indices = tuple(mkIntegerLiteral(meth.site, i) for i in indices)
-    if retvar:
-        decl = mkDeclaration(meth.site, retvar, rettype,
-                             init = get_initializer(meth.site, rettype,
-                                                    None, None, None))
-        decl.toc()
+    with crep.DeviceInstanceContext():
+        if retvar:
+            decl = mkDeclaration(meth.site, retvar, rettype,
+                                 init = get_initializer(meth.site, rettype,
+                                                        None, None, None))
+            decl.toc()
 
-    with LogFailure(meth.site, meth, indices):
-        inargs = [mkLit(meth.site, v, t) for v, t in meth.inp]
-        outargs = [mkLit(meth.site, v, t) for v, t in meth.outp]
-        codegen_call(meth.site, meth,
-                     indices,
-                     inargs, outargs).toc()
+        with LogFailure(meth.site, meth, indices):
+            inargs = [mkLit(meth.site, v, t) for v, t in meth.inp]
+            outargs = [mkLit(meth.site, v, t) for v, t in meth.outp]
+            codegen_call(meth.site, meth,
+                         indices,
+                         inargs, outargs).toc()
     output_dml_state_change("_dev")
 
     reset_line_directive()
@@ -1026,7 +1032,7 @@ def generate_simple_events(device):
                         for i in range(method.dimensions))
         params = tuple(mkLit(site, f'_args->{pname}', ptype)
                        for (pname, ptype) in method.inp)
-        with LogFailure(site, method, indices):
+        with LogFailure(site, method, indices), crep.DeviceInstanceContext():
             code = codegen_call(site, method, indices, params, ())
         code = mkCompound(site, [code])
         code.toc()
@@ -1127,19 +1133,19 @@ def generate_register_events(device):
             for indices in event.all_indices():
                 out('%s%s = SIM_register_event("%s", class, 0, %s);\n'
                     % (crep.get_evclass(event),
-                       ''.join('[' + str(i) + ']' for i in indices),
-                       event.logname_anonymized(indices),
-                       ', '.join(
-                           cname
-                           for (_, cname) in event_callbacks(event, indices))))
+                    ''.join('[' + str(i) + ']' for i in indices),
+                    event.logname_anonymized(indices),
+                    ', '.join(
+                        cname
+                        for (_, cname) in event_callbacks(event,
+                                                          indices))))
         for (method, info) in list(simple_events.items()):
             out(('%s = SIM_register_event("%s", class, 0, %s, %s, %s, %s, '
-                 + 'NULL);\n')
+                + 'NULL);\n')
                 % ((crep.get_evclass(method),
                     method.logname_anonymized())
-                   + tuple(info[fun] for fun in ('callback', 'destroy',
-                                                 'get_value', 'set_value'))))
-
+                + tuple(info[fun] for fun in ('callback', 'destroy',
+                                              'get_value', 'set_value'))))
     out('}\n\n', preindent = -1)
     splitting_point()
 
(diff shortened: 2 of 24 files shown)


syssimics:
DML [#9891185](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9891185): fail

syssimics:
Verification [#9891184](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9891184): fail

syssimics:
Verification [#9893779](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9893779): pass

py/dml/c_backend.py
```diff
@@ -401,7 +406,10 @@ def generate_attr_setter(fname, node, port, dimsizes, cprefix, loopvars,
         out('attr_value_t attr%d = %s;\n' % (dim, list_item))
         valuevar = 'attr%d' % (dim,)
 
-    with NoFailure(node.site):
+    with ExitStack() as stack:
```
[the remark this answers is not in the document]
lwaern-intel: Isn't that syntax very new? What version of python are we working with, again?

py/dml/c_backend.py
```diff
@@ -1713,21 +1746,24 @@ def generate_init(device, initcode, outprefix):
...
-    out("%s = (%s)_obj;\n" % (typ.declaration(name), typ.declaration("")))
+    if not func.independent:
+        (name, typ) = func.cparams[0]
+        out("%s = (%s)_obj;\n" % (typ.declaration(name), typ.declaration("")))
     out("%s%s(%s);\n" % ("" if func.rettype.void
                          else func.rettype.declaration("result") + " = ",
                          func.get_cname(),
                          ", ".join(n for (n, t) in func.cparams)))
-    output_dml_state_change(name)
+    if not func.independent:
```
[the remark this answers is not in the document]
lwaern-intel: One scenario you outlined is an exported independent method as a callback to some SIM api function, with aux data being a pointer to a device-specific stateful data (session/saved) that is then modified inside the independent method.
As you've suggested, we could mention in the documentation for #82, next to the point were we detail how `&` could be used to call dependents from independents, that a dependent must be called when mucking about with device state in order to guarantee that the state-change notifier being notified.

py/dml/c_backend.py
```diff
@@ -2258,10 +2307,118 @@ def generate_alloc_trait_vtables(node):
...
+def generate_startup_calls(devnode):
+    start_function_definition('void _startup_calls(void)')
+    out('{\n', postindent=1)
+
+    def startup_calls(startup_methods):
+        by_dims = {}
+        for (method_data_tuple, dims) in startup_methods:
+            by_dims.setdefault(dims, []).append(method_data_tuple)
+        all_dims = sorted(by_dims)
+        prev_dims = ()
```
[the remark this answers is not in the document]
lwaern-intel: You're right that the code is copy-pasted, but I already stripped out the loop-splitting parts from it. `prev_dims` is part of the mechanism that allow loops to be merged, i.e. what allows this:
```
group g1[i < 3][j < 5] {
  group c[k < 7] {
    independent startup method m() { ... }
  }
  independent startup method n() { ... }
}

group g2[i < 3] {
  independent startup method o();
}
```
to be codegenned as this:
```
for (int i0 = 0; i0 < 3; ++i0) {
  for (int i1 = 0; i1 < 5; ++i1) {
    for (int i2 = 0; i2 < 7; ++i2) {
      m(i0, i1, i2);
    }
    n(i0, i1);
  }
  o(i0);
}
```
instead of genning a whole set of nested loops for each individual independent startup. I can very much believe such boilerplate would add up with an increasing number of startup methods across unique nodes, and while it's unlikely that number will be very large, seeing how we plan to have startup methods in templates in dml-builtins, I'd rather preemptively have this optimization in place.

py/dml/c_backend.py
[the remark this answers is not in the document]
lwaern-intel: Ah, the syntax is actually from 3.1; what I was thinking of is that syntax _with parentheses_, needed for it be multiline -- that was only introduced in 3.10!

py/dml/c_backend.py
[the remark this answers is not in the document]
mandolaerik: There will be a massive number of things done through startup calls, but for non-memoized I expect that we'll mostly do this from top level, like
```
independent startup method _register_attributes() {
  foreach a in (each conf_attribute in this) {
    register_attribute(..., a);
  }
}
```
Also, *if* this is a preformance problem, then there are better ways to optimize, e.g. group methods by number of dimensions and do a single table driven loop for each number of dimensions. Tables with constants are cheap.

py/dml/c_backend.py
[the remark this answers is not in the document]
mandolaerik: Agree, doc-only and belongs to #82.

py/dml/c_backend.py
```diff
@@ -2258,10 +2307,118 @@ def generate_alloc_trait_vtables(node):
...
+    def generate_regular_call(method, idxvars):
+        site = method.site
+        indices = tuple(mkLit(site, idx, TInt(32, False)) for idx in idxvars)
+        outargs = [mkLit(site,
+                         ('*((%s) {0})'
+                          % ((TArray(t, mkIntegerLiteral(site, 1))
+                              .declaration('')),)),
+                         t)
+                   for (_, t) in method.outp]
+        with IgnoreFailure(site):
```
mandolaerik, line 2356: why IgnoreFailure? Shouldn't startup be nonthrowing, so that NoFailure is sufficient? (same goes for generate_trait_calls)

lwaern-intel: It makes sense when you view throwing as an output parameter. Throws are ignored during startup, but an independent startup memoized can be called in order to query whether or not it threw during startup.
Non-memoized independent startups are not allowed to throw.

mandolaerik: Ah, I see. Perhaps a comment to explain that?

(I find memoized throws a bit odd to start with, but I can't deny that they add some symmetry, so I choose not to complain about them)

py/dml/codegen.py
```diff
@@ -238,6 +245,135 @@ def codegen_exit(self, site, retvals):
...
+            stmts.append(mkCopyData(site, mkBoolConstant(site, False), threw))
+        targets = []
+        for ((name, typ), val) in zip(self.outp, retvals):
+            target = self.mkRef(f'p_{name}', typ)
+            stmts.append(mkCopyData(site, val, target))
+            targets.append(target)
+        stmts.append(codegen_return(site, self.outp, self.throws, targets))
+        return mkCompound(site, stmts)
+
+class Memoization:
```
mandolaerik, line 295: consider `abc.ABC` / `abstractmethod` for clarity

py/dml/codegen.py
```diff
@@ -238,6 +245,135 @@ def codegen_exit(self, site, retvals):
...
+    def fail_handler(self):
+        pass
+
+class IndependentMemoized(Memoization):
+    def __init__(self, func):
+        assert func.method.independent
+        self.func = func
+        self.method = func.method
+
+    def mkRef(self, ref, typ):
```
mandolaerik, line 309: PEP8 violation, albeit somewhat understandable.

lwaern-intel: You'll need to clarify, I don't understand what specifically you're referring to.

py/dml/codegen.py
```diff
@@ -238,6 +245,135 @@ def codegen_exit(self, site, retvals):
...
+    def fail_handler(self):
+        return (MemoizedReturnFailure(self.method.rbrace_site, self.mkRef)
+                if self.method.throws else NoFailure(self.method.site))
+
+def memoization_common_prelude(name, site, outp, throws, mkRef):
+    stmts = []
+    if throws:
+        stmts.append(
+            mkIf(site, mkRef('threw', TBool()),
+                 mkReturn(site, mkBoolConstant(site, True))))
```
mandolaerik, line 360: would `fail(site)` be an equivalent way to express this?

py/dml/crep.py
```diff
@@ -19,8 +19,46 @@
...
     'cloggroup',
+    'dev',
+    'require_dev',
+    'maybe_dev_arg',
+    'DeviceInstanceContext',
+    'TypedParamContext',
     )
 
+
+class DeviceInstanceContext:
```
mandolaerik, line 30: missing comment/docstring, what invariants do entry guard?

py/dml/crep.py
```diff
@@ -19,8 +19,46 @@
...
+    active = False
+
+    def __enter__(self):
+        self.prev_context = DeviceInstanceContext.active
+        DeviceInstanceContext.active = True
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        DeviceInstanceContext.active = self.prev_context
+
+class TypedParamContext:
```
mandolaerik, line 39: missing comment/docstring, what invariants does entry guard?

py/dml/codegen.py
lwaern-intel: Seems so. Replaced the `mkReturn` with `Failure.fail_stack[-1].fail(site)`, and tests still pass.

py/dml/codegen.py
```diff
@@ -238,6 +245,135 @@ def codegen_exit(self, site, retvals):
...
+            self.method.name, self.method.site, self.method.outp,
+            self.method.throws, self.mkRef)
+    def exit_handler(self):
+        return MemoizedReturnExit(self.method.site, self.method.outp,
+                                  self.method.throws, self.mkRef)
+    def fail_handler(self):
+        return (MemoizedReturnFailure(self.method.rbrace_site, self.mkRef)
+                if self.method.throws else NoFailure(self.method.site))
+
+def memoization_common_prelude(name, site, outp, throws, mkRef):
```
mandolaerik, line 355: Would be nice with a comment that explains `ran` and `threw`, and how they solve memo problems.

py/dml/codegen.py
lwaern-intel: Aha! Looking at generated code, since MemoizedFailHandler is used, it redundantly sets `ran` to 1.
```
    switch ((int64 )(((struct _t *) _t.trait)->_s_memo_0_t_outs.ran)) {
    ...
    case 1LL: ;
        if (((struct _t *) _t.trait)->_s_memo_0_t_outs.threw) {
            ((struct _t *) _t.trait)->_s_memo_0_t_outs.ran = (int8 )1LL;
            ((struct _t *) _t.trait)->_s_memo_0_t_outs.threw = 1;
            return 1;
        }
        return 0;
        ...
```
So using `mkReturn` directly allows us to avoid having that line be generated.

py/dml/traits.py
```diff
@@ -600,8 +653,15 @@ def scope(self, global_scope):
         s = Symtab(global_scope)
         selfref = mkLit(self.site, '_' + cident(self.name), self.type())
         for name in self.members():
-            s.add(ExpressionSymbol(
-                name, mkSubRef(self.site, selfref, name, '.'), self.site))
+            # HACK the dumbness of this speaks for itself.
+            # One way to address this hack is to add a symbol class with a
+            # deferred-calculated expression.
```
lwaern-intel, line 658: Do you have a suggestion for eliminating this hack?

py/dml/c_backend.py
```diff
@@ -2258,10 +2307,118 @@ def generate_alloc_trait_vtables(node):
...
-                out(f'{name} = MM_MALLOC(1, typeof(*{name}));\n')
+                out(f'{name} = MM_ZALLOC(1, typeof(*{name}));\n')
     out('}\n', preindent=-1)
     splitting_point()
 
+def generate_startup_calls(devnode):
+    start_function_definition('void _startup_calls(void)')
+    out('{\n', postindent=1)
+
+    def startup_calls(startup_methods):
```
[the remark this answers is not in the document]
lwaern-intel: Did this for `startup_calls`, didn't deem the other helpers large enough.

py/dml/codegen.py
lwaern-intel: It was about dromedary case! Doing `mkRef` -> `make_ref` is simple enough.

py/dml/traits.py
lwaern-intel: After some discussion, we decided this wasn't so bad, after all.

syssimics:
Verification [#9927932](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9927932): pass

include/simics/dmllib.h
```diff
@@ -2402,14 +2408,42 @@ UNUSED static uint64 _select_log_level(ht_int_table_t *ht,
         return subsequent;
 }
 
+// The internal format for ht is:
+// dict(trait_identifier -> dict(method_idx -> outs_struct))
+UNUSED static void *_get_shared_idempotent_outs(
```
lwaern-intel: I thought I removed this. Apparently not.

py/dml/ctree.py
```diff
@@ -196,7 +196,6 @@ def expr(self, site):
         expr.incref()
         return expr
 
-# HACK this exists to make a hack in traits.Trait.scope possible
 class InvalidExpressionSymbol(symtab.Symbol):
     """A symbol that corresponds to an invalid expression"""
     def __init__(self, name, mkError, site):
```
mandolaerik, line 201: `make_error`

mandolaerik: also, PEP8 asks you to say `site=site`, without spaces, in the kw args of the `super` call.

py/dml/traits.py
```diff
@@ -653,15 +653,14 @@ def scope(self, global_scope):
...
+            # This is very hacky, but works well
             try:
-                s.add(ExpressionSymbol(
-                    name, mkSubRef(self.site, selfref, name, '.'), self.site))
+                symbol = ExpressionSymbol(
+                    name, mkSubRef(self.site, selfref, name, '.'), self.site)
             except EINDEPENDENTVIOL:
-                s.add(InvalidExpressionSymbol(name, EINDEPENDENTVIOL,
-                                              self.site))
+                symbol = InvalidExpressionSymbol(name, EINDEPENDENTVIOL,
```
mandolaerik, line 661: Might be more conservative to use an ExpressionSymbol with a non-value, because there's a precedent for that: That `Symbol.expr()` may throw an exception is something new, while we already do protect the invariant that standalone non-values will give errors.

py/dml/c_backend.py
```diff
@@ -2258,10 +2293,110 @@ def generate_alloc_trait_vtables(node):
...
+            gen_call(data, idxvars)
+
+        for i in range(len(dims)):
+            out('}\n', preindent=-1)
+
+def generate_startup_calls_entry_function(devnode):
+    start_function_definition('void _startup_calls(void)')
+    out('{\n', postindent=1)
+
+    def generate_regular_call(method, idxvars):
```
mandolaerik, line 2321: you could move out this one (and `generate_trait_calls`) to global scope as well

py/dml/ctree.py
```diff
@@ -3603,7 +3595,7 @@ class NoallocNodeRef(PlainNodeRef):
     @auto_init
     def __init__(self, site, node, indices, node_type): pass
     def exc(self):
-        return ENALLOC(self.site, self.node)
+        return ENALLOC(self.site, elf.node)
```
lwaern-intel, line 3598: WHAT

lwaern-intel: _this never happened._

syssimics:
Verification [#9934828](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9934828): pass

PR #83, opened 2022-04-13
Whitelist based type checking for casts -- SIMICS-17955

lwaern-intel:
~Turns out this simple, obvious change actually passes t126, so we'll need to run verification and test against VP to see what kind of issues this would pose.~ Actually, hold on, I might have made a mistake.


lwaern-intel:
Yeah, I made a mistake in testing.

PR #84, opened 2022-04-13
Whitelist based type checking for casts -- SIMICS-17955

lwaern-intel:
This needs to be tested against VP.

### DIFF
--- a/py/dml/ctree.py
+++ b/py/dml/ctree.py
@@ -3998,13 +3998,38 @@ def mkCast(site, expr, new_type):
         old_type = safe_realtype(expr.ctype())
         if real.cmp(old_type) == 0:
             return expr
-        else:
+        elif old_type.is_arith or isinstance(old_type, TPtr):
             return mkApply(
                 expr.site,
                 mkLit(expr.site, *real.get_store_fun()),
                 (mkCast(expr.site, expr, TInt(64, False)),))
+        else:
+            raise ECAST(site, expr, new_type)
+    if ((real.is_arith or isinstance(real, TBool))
+         and (old_type.is_arith or isinstance(old_type, TBool))):
+        assert (not (real.is_int and real.is_endian)
+                and not (old_type.is_int and old_type.is_endian))
+        return Cast(site, expr, new_type)
+    if ((isinstance(real, TPtr) or real.is_int or isinstance(real, TBool))
+        and (isinstance(old_type, (TPtr, TArray, TFunction))
+             or old_type.is_int or isinstance(old_type, TBool))):
+        assert (not (real.is_int and real.is_endian)
+                and not (old_type.is_int and old_type.is_endian))
+        return Cast(site, expr, new_type)
 
-    return Cast(site, expr, new_type)
+    # Allow unsafe casts from method references to function pointers in DML 1.2
+    # for compatibility reasons
+    if (dml.globals.dml_version == (1, 2) and isinstance(expr, NodeRef)
+        and expr.get_ref()[0].objtype == 'method'
+        and isinstance(real, TPtr) and isinstance(real.base, TFunction)):
+        return Cast(site, expr, new_type)
+    # Allow casts from dev to pointer types in DML 1.2 for compatibility
+    # reasons
+    if (dml.globals.dml_version == (1, 2) and isinstance(expr, NodeRef)
+        and expr.get_ref()[0].objtype == 'device' and isinstance(real, TPtr)):
+        return Cast(site, expr, new_type)
+
+    raise ECAST(site, expr, new_type)
 
 class RValue(Expression):
     '''Wraps an lvalue to prohibit write. Useful when a composite
--- a/test/1.4/errors/T_ECAST.dml
+++ b/test/1.4/errors/T_ECAST.dml
@@ -45,6 +45,7 @@ method init() {
     cast(l, uint32);
     // no error!
     cast(a, uint32);
+    /// ERROR ECAST
     cast(a, double);
     /// ERROR ECAST
     cast(i, a_t);
--- a/test/1.2/operators/T_cast.dml
+++ b/test/1.2/operators/T_cast.dml
@@ -12,7 +12,9 @@ method test() -> (bool ok) {
     local int b = cast(127, int8);
     local int c = cast(128, int8);
     local int d = cast(-1, uint8);
-    log info: "%d %d %d %d", a, b, c, d;
+    local bool e = cast(d, bool);
+    local void *f = cast($dev, void *);
+    log info: "%d %d %d %d %s %p", a, b, c, d, e ? "true" : "false", f;
 
-    ok = a == -24 && b == 127 && c == -128 && d == 255;
+    ok = a == -24 && b == 127 && c == -128 && d == 255 && e && f == $dev.obj;
 }


syssimics:
Verification [#9897114](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9897114): pass

test/1.4/errors/T_ECAST.dml
```diff
@@ -45,6 +45,7 @@ method init() {
     cast(l, uint32);
     // no error!
     cast(a, uint32);
+    /// ERROR ECAST
```
mandolaerik, line 48: Also need error tests for casts between ptr and endian int

mandolaerik (changes requested):
Consider to also improve docs on cast.

Disallowing casts between endian int and ptr is   not an obvious decision -- i recall seeing (highly unidiomatic) ssm code that did this. Perhaps that code was smoked out earlier?
Supporting it would be straightforward and consistent,  but otoh an explicit error would be helpful to divert users from bad language use.

test/1.4/errors/T_ECAST.dml
lwaern-intel: No, cause they don't actually cause errors! The checks are misleading, I'll create another comment thread about that.

py/dml/ctree.py
```diff
@@ -3958,8 +3958,16 @@ def mkCast(site, expr, new_type):
                 expr.site,
                 mkLit(expr.site, *real.get_store_fun()),
                 (mkCast(expr.site, expr, TInt(64, False)),))
+    if real.is_arith and old_type.is_arith:
+        assert (not (real.is_int and real.is_endian)
+                and not (old_type.is_int and old_type.is_endian))
+        return Cast(site, expr, new_type)
+    if ((isinstance(real, TPtr) or (real.is_int and not real.is_endian))
+        or (isinstance(old_type, TPtr)
+            or (old_type.is_int and not old_type.is_endian))):
```
lwaern-intel, line 3967: So this check is misleading -- endian ints already undergo conversion to regular ints earlier in the code. In fact, this check is morally incorrect because of this.
This should have the same asserts as the check before it, so you could combine the two: but that's messy.
```
    if ((real.is_arith and old_type.is_arith)
        or ((isinstance(real, TPtr) or real.is_int)
            and (isinstance(old_type, TPtr) or old_type.is_int))):
        assert (not (real.is_int and real.is_endian)
                and not (old_type.is_int and old_type.is_endian))
        return Cast(site, expr, new_type)
```
Any suggestions?

py/dml/ctree.py
```diff
@@ -3958,8 +3958,16 @@ def mkCast(site, expr, new_type):
                 expr.site,
```
lwaern-intel, line 3958: Oup, this path is unprotected! Needs checks here and a test to sniff it out.

lwaern-intel: Actually, it's not unprotected, it just has a recursive call to `mkCast`. Still, it's good to have an explicit check to give better error messages (since the recursive call will have `uint64` as target type).

py/dml/ctree.py
lwaern-intel: The old check was incorrect in more than one way! (A mistaken `or` instead of `and`). When corrected, a couple of tests fail, all fixable without issue except for one: `1.2/misc/destroy`. This test does the following heinous crime:
```
method del_notify_cb(conf_object_t *subscriber, void *data) { ... }

typedef void (*notify_cb_t)(conf_object_t *subscriber, void *data);

method post_init {
  ...
    SIM_add_global_notifier(Sim_Global_Notify_Object_Delete, $dev.obj,
                            cast($del_notify_cb, notify_cb_t), NULL);
}
```
This is very much bad and wrong for multiple reasons -- for one, the generated function for `del_notify_cb` _has incorrect signature_ for the type `notify_cb_t` and so this test actually has undefined behavior.
This raises the very uncomfortable question if SSM abuses this, too. This just further proves we really need to test this against VP.

py/dml/ctree.py
```diff
@@ -3953,13 +3953,28 @@ def mkCast(site, expr, new_type):
...
+        assert (not (real.is_int and real.is_endian)
+                and not (old_type.is_int and old_type.is_endian))
+        return Cast(site, expr, new_type)
+
+    # Allow unsafe casts from method references to function pointers in DML 1.2
+    # for compatibility reasons
+    if (dml.globals.dml_version == (1, 2) and isinstance(expr, NodeRef)
+        and expr.get_ref()[0].objtype == 'method'
+        and isinstance(real, TPtr) and isinstance(real.base, TFunction)):
+        return Cast(site, expr, new_type)
```
lwaern-intel, line 3975: I wish to be rid of this if possible. We must test against VP to make sure that's safe to do, and `1.2/misc/destroy` will need to be rewritten.

py/dml/ctree.py
```diff
@@ -3953,13 +3953,30 @@ def mkCast(site, expr, new_type):
...
                 expr.site,
                 mkLit(expr.site, *real.get_store_fun()),
                 (mkCast(expr.site, expr, TInt(64, False)),))
+        else:
+            raise ECAST(site, expr, new_type)
+    if (((real.is_arith or isinstance(real, TBool))
+         and (old_type.is_arith or isinstance(old_type, TBool)))
+        or ((isinstance(real, TPtr) or real.is_int or isinstance(real, TBool))
+            and (isinstance(old_type, (TPtr, TArray)) or old_type.is_int
+                 or isinstance(old_type, TBool)))):
```
lwaern-intel, line 3967: This is horrifically ugly and _desperately_ needs a rewrite.

syssimics:
Verification [#9911605](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9911605): fail

syssimics:
Verification [#9911774](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9911774): pass

py/dml/ctree.py
mandolaerik: why bother? it's within a 1.2 guard, and reflects a (sad) 1.2 idiom. IIRC I have seen it used for hap callbacks. You could consider restricting the 1.2 guard to check site.file_info.dml_version instead.

py/dml/ctree.py
```diff
@@ -3953,13 +3953,30 @@ def mkCast(site, expr, new_type):
...
+        else:
+            raise ECAST(site, expr, new_type)
+    if (((real.is_arith or isinstance(real, TBool))
+         and (old_type.is_arith or isinstance(old_type, TBool)))
+        or ((isinstance(real, TPtr) or real.is_int or isinstance(real, TBool))
+            and (isinstance(old_type, (TPtr, TArray, TFunction))
+                 or old_type.is_int or isinstance(old_type, TBool)))):
+        assert (not (real.is_int and real.is_endian)
+                and not (old_type.is_int and old_type.is_endian))
+        return Cast(site, expr, new_type)
```
mandolaerik, line 3970: A fairly exhaustive test in ctree_test could be appropriate?

Rather complex condition, perhaps worthwhile to split in two and duplicate `return Cast`.

Unsure if TFunction is right, if you have `local void (*p)(void);`, then isn't `cast(*p, int)` illegal?

mandolaerik (commented):
did you check if the documentation has some sloppy wording that should be strictened up accordingly?

py/dml/ctree.py
mandolaerik: Also, cast between pointers and non-`uintptr_t` integers is questionable, I think they give GCC warnings. We need a stance on this.
A fairly common use case is to store an integer in a `void *`, as is seen in `uint64_event`. It can make sense to convert uint8 into such pointer, and later convert the pointer back into uint8. But I find it hard to constuct a use case where conversion from TArray into any other integer than uintptr_t would make sense (but OTOH, compatibility rules should be transitive so perhaps we must keep it a warning, like GCC does).
One option is to create a separate case for pointer-to-int conversions, produce our own warnings, and generate a double-cast via uintptr_t to silence GCC warnings.

lwaern-intel:
Not yet, I'll take a look.

py/dml/ctree.py
lwaern-intel: > A fairly exhaustive test in ctree_test could be appropriate?
> 
> Rather complex condition, perhaps worthwhile to split in two and duplicate `return Cast`.
> 
> Unsure if TFunction is right, if you have `local void (*p)(void);`, then isn't `cast(*p, int)` illegal?

My intent with allowing TFunction as old_type is so that casting a function reference to a function pointer type won't error out. As for `cast(*p, int)`, surprisingly enough, `(int) (*p)` is legal in C; the dereferenced function pointer gets automatically converted back into a pointer again. (Generates a warning though since `int` is smaller than pointer width).

lwaern-intel:
After some brief discussion, we determined documentation is fine.

mandolaerik (approved):

lwaern-intel:
Shouldn't be merged until we have the opportunity to test against VP.

syssimics:
Verification [#10095518](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10095518): fail

syssimics:
Verification [#10100204](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10100204): pass

lwaern-intel:
With that last patch, this passes testing against VP (modulo unrelated failures)

mandolaerik (approved):

PR #85, opened 2022-04-25
1.4: Utility: map_target: Make read_bytes/write_bytes overrideable

gwikland:


lib/1.4/utility.dml
```diff
@@ -1300,7 +1300,7 @@ template map_target is (connect, _qname) {
          Reads `size` bytes into `bytes`, starting at `addr` in the connected
          object. Throws an exception if the read fails.
     */
-    shared method read_bytes(uint64 addr, uint64 size, uint8 *bytes) throws {
+    shared method read_bytes(uint64 addr, uint64 size, uint8 *bytes) throws default{
```
mandolaerik, line 1303: space before `{`

lib/1.4/utility.dml
```diff
@@ -1348,7 +1348,7 @@ template map_target is (connect, _qname) {
          object. Throws an exception if the write fails.
     */
     shared method write_bytes(uint64 addr, uint64 size,
-                              const uint8 *bytes) throws {
+                              const uint8 *bytes) throws default {
```
mandolaerik, line 1351: What's the reason for overriding these? My impression is that they are convenience methods; if they don't suite you, then you are free to write a convenience method with a different name.

Note that if you want to add some statically known atoms to the transaction, then `issue` is the method to override.

gwikland: When these throw I want to know whether it was a read or write. The catch is way up in the hierarchy.
Of course I can create my own method, however adding default seems more appropriate. I am just adding logic on top of the default method without affecting the behavior.

gwikland:
```
connect address_space is (map_target) "DMAC address space" {
    session bool failed_write;
    session bool failed_read;

    method write_bytes(uint64 addr, uint64 size,
                       const uint8 *bytes) throws {
        try {
            default(addr, size, bytes);
        } catch {
            failed_write = true;
            throw;
        }
    }
    method read_bytes(uint64 addr, uint64 size, uint8 *bytes) throws {
        try {
            default(addr, size, bytes);
        } catch {
            failed_read = true;
            throw;
        }
    }
}
```

svenwest: You can still do this by overriding the issue method, though:

```
connect address_space is (map_target) "DMAC address space" {
    session bool failed_write;
    session bool failed_read;

    method issue(transaction_t *t, uint64 addr) -> (exception_t) {
        local exception_t exc = default(t, addr);
        if (exc != Sim_PE_No_Exception) {
            if (SIM_transaction_is_read(t))
                failed_read = true;
            else
                failed_write = true;
        }
}
```

gwikland: Neat, I will use that :)

svenwest: Oops, forgot the return-statement in my code above, updated :-)

svenwest: Also note that using 'device-global' state like this isn't reentrant, i.e. if a read/write operation going out of this code results in another call into this code, in the same call-chain, you might run into problems.

gwikland: Sure, In the catch statement which is several levels above I abort the transfer and clear these flags. The preferred way would be to add expection types to the catch statement, like c++ supports. Then no state would have to be saved. I do not want to call top level code in leaf code in a long callstack, which would be the other solution.

PR #87, opened 2022-05-02
Fix field-array names in register view

svenwest:


mandolaerik (approved):

lib/1.4/dml-builtins.dml
```diff
@@ -2885,7 +2886,8 @@ also used by the default implementations of hard reset, soft reset and
 power-on reset. Defaults to 0.
 </dd></dl>
 */
-template field is (object, _bitsize, _lsb, init_val, shown_desc, _get, _set) {
+template field is (object, init_val, shown_desc,
+                   _bitsize, _lsb, _get, _set, _qname) {
```
mandolaerik, line 2890: Unfortunately, this is currently expensive. Inheriting qname means that DMLC will generate one qname function for every field, which shows up in the compile time profile even when we only do it on registers.

svenwest: Sounds like a serious limitation of DML 1.4 then.
We can't have register-view without presenting proper names for registers and fields, but doing that is prohibitively expensive in terms of compile time?
Remind me again why we got rid of the device-info xml?

syssimics:
Verification [#9943930](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9943930): pass

lib/1.4/dml-builtins.dml
mandolaerik: I think @lwaern-intel 's work on attribute accessor trampolines may provide an economic mechanism that is less broken than what we have today, where a field `reg.mygroup[2].f[3]` would be named `"f[2][3]"` by reginfo, which likely is unique but still different from the expected `"mygroup[2].f[3]"`

lwaern-intel:
I've fixed this since leveraging my work on free `_qname()`.

PR #88, opened 2022-05-03
Suggest `sizeoftype` when erroring on `sizeof` -- SIMICS-17991

lwaern-intel:


RELEASENOTES.docu
```diff
@@ -1706,4 +1706,7 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
   <build-id value="6143"><add-note> Implicitly initialized <tt>local</tt>
       variables of (partially) const-qualified struct type no longer result in
       invalid generated C <bug number="SIMICS-10197"/>.</add-note></build-id>
+  <build-id value="next"><add-note> Any error encountered when evaluating the
+      operand to <tt>sizeof</tt> will now cause DMLC to suggest
+      <tt>sizeoftype</tt> <bug number="SIMICS-17991"/>.</add-note></build-id>
```
lwaern-intel, line 1711: Is this releasenote warranted? On one hand, it has no bearing outside of error messages; on the other hand, it does close an issue.

mandolaerik: It's warranted if it closes an *externally reported* issue.

lwaern-intel: It's not; internally reported by Jakob. I'm removing this note, then.

py/dml/messages.py
```diff
@@ -1511,6 +1511,21 @@ class EVLACONST(DMLError):
     fmt = ("variable length array declared with (partially) const-qualified "
            + "type")
 
+class ESIZEOFOP(DMLError):
```
lwaern-intel, line 1514: Verification revealed an interesting problem with this error wrapping: control flow paths that rely on catching errors that are not terminal for the program, in particular, the eager evaluation of params in `structure.py` does `except EIDXVAR` to detect when static indices are used. Do we want to ad-hoc patch such `except`s, or is this indicative that this entire approach is wrong?

syssimics:
Verification [#9947563](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9947563): fail

py/dml/codegen.py
```diff
@@ -543,7 +543,10 @@ def expr_unop(tree, location, scope):
             report(WSIZEOFTYPE(tree.site))
             return codegen_sizeof(
                 tree.site, mkLit(tree.site, cident(var), None))
-    rh = codegen_expression_maybe_nonvalue(rh_ast, location, scope)
+    try:
+        rh = codegen_expression_maybe_nonvalue(rh_ast, location, scope)
+    except EIDENT as e:
+        raise ESIZEOFOP(e.site, e.identifier) if op == 'sizeof' else e
```
mandolaerik, line 549: unsure, but `else None` *might* be better; subtle differences between `raise` and `raise e`

lwaern-intel: Huh! Didn't know about that feature of python.

lwaern-intel: Oh, it's not. `raise None` is invalid, what you mean is `raise`: it would be rewritten as:
```
if op == 'sizeof':
    raise ESIZEOFOP(e.site, e.identifier)
raise
```

mandolaerik: OK. I recall seeing messages from Python talking about None being raised when referring to a naked `raise` statement. Maybe it was a py2 thing. Your suggestion is clearer in any case.

py/dml/messages.py
```diff
@@ -1511,6 +1515,14 @@ class EVLACONST(DMLError):
...
+    """
+    The operand of 'sizeof' contains an unknown identifier.
+    This often occurs because `sizeof` was used when `sizeoftype` was intended.
```
mandolaerik, line 1521: This text appears in documentation and describes the error message kind, not the error. Current is OK, but you can consider smth like "This is a variant of the EIDENT message that..."

lwaern-intel: Ok. Also renamed it to `EIDENTSIZEOF` to really rub it in.

py/dml/codegen.py
```diff
@@ -543,7 +543,12 @@ def expr_unop(tree, location, scope):
             report(WSIZEOFTYPE(tree.site))
             return codegen_sizeof(
                 tree.site, mkLit(tree.site, cident(var), None))
-    rh = codegen_expression_maybe_nonvalue(rh_ast, location, scope)
+    try:
+        rh = codegen_expression_maybe_nonvalue(rh_ast, location, scope)
+    except EIDENT as e:
+        if op == 'sizeof':
```
mandolaerik, line 549: Perhaps add a precondition that e.ident is the name of a type?

lwaern-intel: Are you suggesting running `eval_type`? Because `typedefs` wouldn't cover e.g. `uint32`. I'd really rather not.

mandolaerik: I would suggest `types.parse_type`. I think it does exactly what you want, but perhaps I missed some corner.

lwaern-intel: `parse_type` alone isn't enough, but when _combined_ with typedefs... maybe. I'm not 100% confident it covers all cases, but I feel like any remaining corner would be ill-advised in the first place.

mandolaerik (approved):

syssimics:
Verification [#9953254](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9953254): pass

PR #90, opened 2022-05-11
Allow independent, startup and memoized as idents

mandolaerik:


syssimics:
Verification [#9971381](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9971381): pass

lwaern-intel:
Oh good, you preempted me.

PR #93, opened 2022-05-16
port-dml: substitute unmapped_offset for undefined -- SIMICS-19191

lwaern-intel:


py/dml/dmlparse.py
```diff
@@ -1144,6 +1144,9 @@ def sizespec_empty(t):
 @prod
 def offsetspec(t):
     'offsetspec : AT expression'
+    if (t[2].kind == 'undefined' and site(t).dml_version() == (1, 2)
+        and logging.show_porting):
+        report(PUNDEFOFFS(site(t, 2)))
```
lwaern-intel, line 1149: Instead of doing it here, we could check it in `reginfo.explode_register()` to cover `param offset = undefined` for a small additional complexity cost. Yes or no?

py/dml/dmlparse.py
```diff
@@ -1144,6 +1144,9 @@ def sizespec_empty(t):
 @prod
 def offsetspec(t):
     'offsetspec : AT expression'
+    if (t[2].kind == 'undefined' and site(t).dml_version() == (1, 2)
```
mandolaerik, line 1147: I'd reorder the conjunction with the version check first to highlight that this is dying code.

mandolaerik (commented):
Externally reported, so deserves a releasenote.

py/dml/dmlparse.py
mandolaerik: that's nice, in particular if you can hide it in an existing version==1.2 clause. But not strictly needed.

py/dml/dmlparse.py
lwaern-intel: > Externally reported, so deserves a releasenote.

Which `RELEASENOTES` should that be in?

lwaern-intel:
> Externally reported, so deserves a releasenote.

Which `RELEASENOTES` should that be in?

mandolaerik:
Yes, use SIMINT as that's the ticket seen by the customer. It's specific to current users of 1.2, so I'd say RELEASENOTES-1.2.

syssimics:
Verification [#9985642](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=9985642): pass

py/dml/messages.py
```diff
@@ -2204,3 +2204,9 @@ class PRENAME_TEMPLATE(PortingMessage):
     `unimpl`.
     """
     fmt = ""
+
+class PUNDEFOFFS(PortingMessage):
+    """`undefined` is no longer a valid offset for registers; `unmapped_offset`
+    should be used instead.
+    """
+    fmt = "Use 'unmapped_offset' instead of 'undefined'"
```
mandolaerik, line 2212: fmt is redundant for porting messages (haven't cleaned it up yet elsewhere)

mandolaerik (approved):

PR #95, opened 2022-05-23
Various doc improvements -- SIMICS-18539, -18138, -18373. -18381

lwaern-intel:


doc/1.4/language.md
```diff
@@ -987,22 +987,42 @@ Templates are imported into an object declaration body using
...
-hierarchy is used to deduce which method overrides the other: If one
-template *B* instantiates another template *A*, directly or indirectly,
-then methods from *B* override methods from *A*. Note, however, that
-overrides can only happen on methods and parameters that are
-declared `default`. Example:
+These can be used in any context where an object declaration may be written, and
+has the effect of expanding the body of the template at the point of the `is`.
+Using `is` together with object declarations is typically more idiomatic than
+the standalone `is` object statement, but the latter is very useful in
+conjunction with [`in each` declarations](#in-each-declarations); for example:
```
mandolaerik, line 1007: Another use case is top level, where standalone `is` is the only way to instantiate templates.

mandolaerik: Also, the `in each { is }` example is a bit weird; we had a discussion on unifying syntaxes for object-like declarations to allow `in each field is A;`, `template a is b;`; the conclusion was that this would not be wrong in principle, but the use cases are so far too weak (together with irregularities such as `desc` string shorthand only making sense on concrete object declarations).

If we think that the in each use case is a too weak use case for adding an `in each field is A { ... }`, then it is inconsistent to present it as the primary motivation for `in each field { is A; }`; we could just as well add the former form (unless we think the latter looks better).

doc/1.4/language.md
```diff
@@ -1435,6 +1455,19 @@ local uint32 v = bf;
...
 
+### Serializable types
+_Serializable types_ are types that the DML compiler knows how to serialize and
+deserialize for the purposes of checkpointing. This is important for the use of
+[`saved` variables](#saved-variables) and the [`after`
+statement](#after-statements).
+
+All primitive non-pointer data types (integers, floating-point types, booleans,
+etc.) are considered serializable, as is any struct, layout, or array type
+consisting entirely of serializable types. Any pointer type is not considered
```
mandolaerik, line 1466: I'm genuinely unsure if the plural/singular mix "*are* considered, as *is*" is good English.

lwaern-intel: Couldn't find anything about it from cursory research, but I see no problem with it. E.g. "They are wrong, as is he."

doc/1.4/language.md
```diff
@@ -2821,28 +2854,84 @@ used in expression-statements.
...
+```
+```
+(a, i) = m();
+```
+
+Targets are updated simultaneously, meaning it's possible to e.g. swap the
+contents of variables through the following:
+```
+(a, b) = (b, a)
+```
```
mandolaerik, line 2890: You could consider a note with a cross-reference to the syntactically related multi-value return statements and variable declarations, if that can be done without disturbing the flow.

doc/1.4/language.md
mandolaerik: I agree that there's nothing *wrong* with it. Question is if e.g. s/All..are/Any..is/ would be *better*. Don't know.

doc/1.4/language.md
lwaern-intel: No, I can't make it fit. I knit them together in new changes to the method calls section, though.

doc/1.4/language.md
```diff
@@ -987,22 +987,43 @@ Templates are imported into an object declaration body using
...
+has the effect of expanding the body of the template at the point of the `is`.
+Using `is` together with object declarations is typically more idiomatic than
+the standalone `is` object statement; however, the latter is useful in order
+to instantiate templates for the top-level device object, and also for use in
+conjunction with [`in each` declarations](#in-each-declarations); for example:
+
+```
+register r {
+    in each field {
+        is A;
```
mandolaerik, line 1013: doesn't hurt to concretize, e.g. `is write_1_clears;` turns it into a potentially real use case

doc/1.4/language.md
```diff
@@ -987,22 +987,43 @@ Templates are imported into an object declaration body using
...
-hierarchy is used to deduce which method overrides the other: If one
-template *B* instantiates another template *A*, directly or indirectly,
-then methods from *B* override methods from *A*. Note, however, that
-overrides can only happen on methods and parameters that are
-declared `default`. Example:
+These can be used in any context where an object declaration may be written, and
+has the effect of expanding the body of the template at the point of the `is`.
+Using `is` together with object declarations is typically more idiomatic than
+the standalone `is` object statement; however, the latter is useful in order
+to instantiate templates for the top-level device object, and also for use in
```
mandolaerik, line 1007: I'd say instantiate "on" or "in" the device object.

mandolaerik (approved):

doc/1.4/language.md
```diff
@@ -2935,6 +2939,36 @@ scope.
...
+* Performing a method call and propagating the return values:
+```
+method n(int i) -> (bool, int) {
+    ...
+}
+
+method m() -> (bool, int) {
+    return n(4);
+}
+```
```
mandolaerik, line 2970: you could make this an extension of the previous example and flip the relation, `method n() -> (bool, int) { return m(); }`.

doc/1.4/language.md
```diff
@@ -3004,12 +3038,21 @@ method m() -> (int) { ... }
...
+as an initializer in any context that accepts non-constant initializers; i.e.,
+in [assignment statements](#assignment-statements) (as shown above), [local
+variable declarations](#local-statements), and [return
+statements](#return-statements). For example:
 ```
 // declare multiple variables, and
 // initialize them from one method call
 local (int i, uint8 j) = m(e1);
+
+
```
mandolaerik, line 3051: double newline

doc/1.4/running-dmlc.md
```diff
@@ -17,8 +17,9 @@ The DML compiler and its libraries are available as part of the
 The `dmlc` compiler can be build locally. This requires an
 installation of the Simics 6 base package.
 
-In order to build the compiler, you can copy source directory `dmlc`
-into the `modules` subdirectory of your Simics project. The compiler
+In order to build the compiler, checkout [the DML repository
+](https://github.com/intel/device-modeling-language) into the
```
mandolaerik, line 21: I would remove the whitespace before `]`, though I don't think this is enforced by a style guide

doc/1.4/language.md
```diff
@@ -3004,12 +3038,21 @@ method m() -> (int) { ... }
...
+// Propagate all return values from a
+// method call as the return values of
+// the caller.
```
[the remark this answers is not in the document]
lwaern-intel: To follow existing code. No idea. Figured it looked better in the docs or something.

syssimics:
Verification [#10004578](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10004578): fail

doc/1.4/running-dmlc.md
```diff
@@ -17,8 +17,8 @@ The DML compiler and its libraries are available as part of the
 The `dmlc` compiler can be build locally. This requires an
 installation of the Simics 6 base package.
 
-In order to build the compiler, checkout [the DML repository](
-https://github.com/intel/device-modeling-language) into the
+In order to build the compiler, checkout [the DML repository
+](https://github.com/intel/device-modeling-language) into the
```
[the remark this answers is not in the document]
lwaern-intel: I don't really consider that any prettier, only perhaps less strange.

mandolaerik (approved):

PR #97, opened 2022-06-07
Softcode attribute registration

lwaern-intel:
Currently only works for top-level, non-array attribute objects.

### DIFF
--- a/include/simics/dmllib.h
+++ b/include/simics/dmllib.h
@@ -1333,18 +1333,19 @@ typedef struct {
         const char *attrname;
         get_attr_t get;
         set_attr_t set;
+        void *getset_data;
 } _port_attr_t;
 static attr_value_t
-_getattr_from_portobj(lang_void *ptr, conf_object_t *obj, attr_value_t *idx)
+_get_legacy_proxy_attr(lang_void *ptr, conf_object_t *obj, attr_value_t *idx)
 {
         ASSERT(SIM_attr_is_nil(*idx));
         _port_attr_t *port = (_port_attr_t *)ptr;
         conf_object_t *portobj = *(conf_object_t **)(
                 (uintptr_t)obj + port->port_obj_offset);
-        return port->get(NULL, portobj, NULL);
+        return port->get(port->getset_data, portobj, NULL);
 }
 static set_error_t
-_setattr_from_portobj(lang_void *ptr, conf_object_t *obj, attr_value_t *val,
+_set_legacy_proxy_attr(lang_void *ptr, conf_object_t *obj, attr_value_t *val,
                       attr_value_t *idx)
 {
         ASSERT(SIM_attr_is_nil(*idx));
@@ -1352,7 +1353,7 @@ _setattr_from_portobj(lang_void *ptr, conf_object_t *obj, attr_value_t *val,
         conf_object_t *portobj = *(conf_object_t **)(
                 (uintptr_t)obj + port->port_obj_offset);
         if (port->attrname == NULL || SIM_object_is_configured(portobj)) {
-                return port->set(NULL, portobj, val, NULL);
+                return port->set(port->getset_data, portobj, val, NULL);
         } else {
                 // port attribute is registered as required; need to propagate
                 // value through API call to fulfil requirement
@@ -1373,14 +1374,13 @@ _register_port_attr_no_aux(conf_class_t *portcls, const char *attrname,
 // port_obj_offset is the offset within the device struct of a pointer to the
 // port object.
 UNUSED static void
-_register_port_attr(conf_class_t *devcls, conf_class_t *portcls,
-                    ptrdiff_t port_obj_offset, bool is_bank,
-                    const char *portname, const char *attrname,
-                    get_attr_t getter, set_attr_t setter,
-                    attr_attr_t attr, const char *type, const char *desc)
+_register_port_legacy_proxy_attr(conf_class_t *devcls, conf_class_t *portcls,
+                                 ptrdiff_t port_obj_offset, bool is_bank,
+                                 const char *portname, const char *attrname,
+                                 get_attr_t getter, set_attr_t setter,
+                                 attr_attr_t attr, const char *type,
+                                 const char *desc, void *getset_data)
 {
-        _register_port_attr_no_aux(portcls, attrname,
-                getter, setter, attr, type, desc);
         _port_attr_t *data = MM_MALLOC(1, _port_attr_t);
         data->port_obj_offset = port_obj_offset;
         if ((attr & Sim_Attr_Flag_Mask) == Sim_Attr_Required) {
@@ -1391,6 +1391,7 @@ _register_port_attr(conf_class_t *devcls, conf_class_t *portcls,
         }
         data->get = getter;
         data->set = setter;
+        data->getset_data = getset_data;
         char name[strlen(portname) + strlen(attrname) + 2];
         sprintf(name, "%s_%s", portname, attrname);
         strbuf_t proxy_desc = sb_newf(
@@ -1398,22 +1399,36 @@ _register_port_attr(conf_class_t *devcls, conf_class_t *portcls,
                 is_bank ? "bank" : "port", portname, attrname);
         SIM_register_typed_attribute(
                 devcls, name,
-                getter ? _getattr_from_portobj : NULL, data,
-                setter ? _setattr_from_portobj : NULL, data,
+                getter ? _get_legacy_proxy_attr : NULL, data,
+                setter ? _set_legacy_proxy_attr : NULL, data,
                 (attr_attr_t)(Sim_Attr_Pseudo | Sim_Attr_Internal),
                 type, NULL, sb_str(&proxy_desc));
         sb_free(&proxy_desc);
+}
+
+UNUSED static void
+_register_port_attr(conf_class_t *devcls, conf_class_t *portcls,
+                    ptrdiff_t port_obj_offset, bool is_bank,
+                    const char *portname, const char *attrname,
+                    get_attr_t getter, set_attr_t setter,
+                    attr_attr_t attr, const char *type, const char *desc) {
+    _register_port_attr_no_aux(portcls, attrname, getter, setter, attr, type,
+                               desc);
+    _register_port_legacy_proxy_attr(devcls, portcls, port_obj_offset, is_bank,
+                                     portname, attrname, getter, setter, attr,
+                                     type, desc, NULL);
 }
 
 typedef struct {
         ptrdiff_t port_obj_base_offset;
         get_attr_t get;
         set_attr_t set;
+        void *getset_data;
         uint32 array_size;
 } _port_array_attr_t;
 static attr_value_t
-_getattr_from_portobj_array(lang_void *ptr, conf_object_t *obj,
-                            attr_value_t *idx)
+_get_legacy_proxy_array_attr(lang_void *ptr, conf_object_t *obj,
+                             attr_value_t *idx)
 {
         ASSERT(SIM_attr_is_nil(*idx));
         _port_array_attr_t *port = (_port_array_attr_t *)ptr;
@@ -1422,7 +1437,8 @@ _getattr_from_portobj_array(lang_void *ptr, conf_object_t *obj,
         attr_value_t vals = SIM_alloc_attr_list(port->array_size);
         for (uint32 i = 0; i < port->array_size; i++) {
                 conf_object_t *port_obj = port_obj_ptr_base[i];
-                attr_value_t val = port->get(NULL, port_obj, NULL);
+                attr_value_t val = port->get(port->getset_data, port_obj,
+                                             NULL);
                 if (SIM_attr_is_invalid(val)) {
                         SIM_attr_free(&vals);
                         return val;
@@ -1432,8 +1448,8 @@ _getattr_from_portobj_array(lang_void *ptr, conf_object_t *obj,
         return vals;
 }
 static set_error_t
-_setattr_from_portobj_array(lang_void *ptr, conf_object_t *obj,
-                            attr_value_t *vals, attr_value_t *idx)
+_set_legacy_proxy_array_attr(lang_void *ptr, conf_object_t *obj,
+                             attr_value_t *vals, attr_value_t *idx)
 {
         ASSERT(SIM_attr_is_nil(*idx));
         _port_array_attr_t *port = (_port_array_attr_t *)ptr;
@@ -1442,7 +1458,8 @@ _setattr_from_portobj_array(lang_void *ptr, conf_object_t *obj,
         for (uint32 i = 0; i < port->array_size; i++) {
                 conf_object_t *port_obj = port_obj_ptr_base[i];
                 attr_value_t val = SIM_attr_list_item(*vals, i);
-                set_error_t err = port->set(NULL, port_obj, &val, NULL);
+                set_error_t err = port->set(port->getset_data, port_obj, &val,
+                                            NULL);
                 if (err != Sim_Set_Ok) {
                         return err;
                 }
@@ -1452,20 +1469,17 @@ _setattr_from_portobj_array(lang_void *ptr, conf_object_t *obj,
 // port_obj_offset is the offset within the device struct of the first pointer
 // to a port object. Remaining port objects are stored consecutively in memory.
 UNUSED static void
-_register_port_array_attr(conf_class_t *devcls, conf_class_t *portcls,
-                          ptrdiff_t port_obj_offset,
-                          uint32 array_size, bool is_bank,
-                          const char *portname, const char *attrname,
-                          get_attr_t getter, set_attr_t setter,
-                          attr_attr_t attr, const char *type, const char *desc)
-{
-        _register_port_attr_no_aux(portcls, attrname,
-                getter, setter, attr, type, desc);
+_register_port_array_legacy_proxy_attr(
+    conf_class_t *devcls, conf_class_t *portcls, ptrdiff_t port_obj_offset,
+    uint32 array_size, bool is_bank, const char *portname,
+    const char *attrname, get_attr_t getter, set_attr_t setter,
+    attr_attr_t attr, const char *type, const char *desc, void *getset_data) {
         _port_array_attr_t *data = MM_MALLOC(1, _port_array_attr_t);
         data->port_obj_base_offset = port_obj_offset;
         data->get = getter;
         data->set = setter;
         data->array_size = array_size;
+        data->getset_data = getset_data;
         char name[strlen(portname) + strlen(attrname) + 2];
         sprintf(name, "%s_%s", portname, attrname);
         strbuf_t proxy_desc = sb_newf("Proxy attribute for %s.%s[].%s",
@@ -1474,13 +1488,29 @@ _register_port_array_attr(conf_class_t *devcls, conf_class_t *portcls,
         strbuf_t proxy_type = sb_newf("[%s{%d}]", type, array_size);
         SIM_register_typed_attribute(
                 devcls, name,
-                getter ? _getattr_from_portobj_array : NULL, data,
-                setter ? _setattr_from_portobj_array : NULL, data,
+                getter ? _get_legacy_proxy_array_attr : NULL, data,
+                setter ? _set_legacy_proxy_array_attr : NULL, data,
                 (attr_attr_t)(Sim_Attr_Pseudo | Sim_Attr_Internal),
                 sb_str(&proxy_type), NULL, sb_str(&proxy_desc));
         sb_free(&proxy_type);
         sb_free(&proxy_desc);
 }
+
+UNUSED static void
+_register_port_array_attr(conf_class_t *devcls, conf_class_t *portcls,
+                          ptrdiff_t port_obj_offset, uint32 array_size,
+                          bool is_bank, const char *portname,
+                          const char *attrname, get_attr_t getter,
+                          set_attr_t setter, attr_attr_t attr,
+                          const char *type, const char *desc) {
+    _register_port_attr_no_aux(portcls, attrname, getter, setter, attr, type,
+                               desc);
+    _register_port_array_legacy_proxy_attr(devcls, portcls, port_obj_offset,
+                                           array_size, is_bank, portname,
+                                           attrname, getter, setter, attr,
+                                           type, desc, NULL);
+}
+
 
 typedef struct {
         char *bufs[4];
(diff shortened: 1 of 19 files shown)


syssimics:
Verification [#10348413](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10348413): fail

syssimics:
Verification [#10348710](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10348710): pass

lwaern-intel:
Oh, I need to add DMLC checks that ensure that important parameters like type, documentation, flags, etc. aren't dependent on indices! That was previously ensured by the `param_const` calls.

syssimics:
Verification [#10388116](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10388116): pass

syssimics:
Verification [#10421049](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10421049): pass

syssimics:
Verification [#10434240](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10434240): pass

lib/1.2/dml-builtins.dml
```diff
@@ -26,6 +26,8 @@ import "simics/processor/event.dml";
 
 import "simics/simulator/python.dml";
 
+// import "attribute-softcoding.dml";
+
```
mandolaerik, line 30: forgotten leftovers

include/simics/dmllib.h
```diff
@@ -2560,12 +2593,144 @@ UNUSED static void _free_table(ht_int_table_t *table) {
...
+    set_error_t (*set_portobj_attr)(void *, conf_object_t *, attr_value_t *,
+                            attr_value_t *)) {
+    for (uint32 i = 0; i < sequence.num; ++i) {
+        _vtable_list_t list = attribute_vtables[i];
+        uint64 num = list.num / sequence.array_size;
+        uint64 start = num * sequence.array_idx;
+        _traitref_t traitref = { list.vtable,
+                                 { .id = list.id, .encoded_index = start }};
+        _dml_attr_conf_info_t attr_info = get_attribute_info(traitref);
+        if (!attr_info.generated) {
```
mandolaerik, line 2658: Confused me on my first read, past tense indicates it means "operation succeeded" rather than "please do it". Also I'd say we register stuff rather than generate. I suggest `skip`.

include/simics/dmllib.h
```diff
@@ -2560,12 +2593,144 @@ UNUSED static void _free_table(ht_int_table_t *table) {
...
+
+        _dml_attr_getset_info_t attr_getset_info = {
+            &id_info_array[list.id],
+            traitref.trait,
+            num,
+            attr_info.allow_cutoff
+        };
+        _dml_attr_getset_info_t *attr_get_info = NULL;
+        if (attr_info.readable) {
+            attr_get_info = MM_MALLOC(1, _dml_attr_getset_info_t);
```
mandolaerik, line 2670: note: WBN to have a function for malloc that never will be freed

include/simics/dmllib.h
```diff
@@ -2560,12 +2593,144 @@ UNUSED static void _free_table(ht_int_table_t *table) {
...
+        }
+        _dml_attr_getset_info_t *attr_set_info = NULL;
+        if (attr_info.writable) {
+            attr_set_info = MM_MALLOC(1, _dml_attr_getset_info_t);
+            *attr_set_info = attr_getset_info;
+        }
+
+        char type[1024];
+        char tmp_type[1024];
+        int written = snprintf(type, 1024, "%s", attr_info.type);
```
lwaern-intel, line 2681: This should just use `strbuf`. I was less acquainted with it at the time I wrote this code.

py/dml/c_backend.py
```diff
@@ -44,6 +45,8 @@ def get_attr_flags(obj):
         flags = 'Sim_Attr_Optional'
     elif conf == 'pseudo':
         flags = 'Sim_Attr_Pseudo'
+    elif conf == 'none':
+        flags = '0'
```
lwaern-intel, line 49: I don't remember why I added this. I should try ripping this out.

py/dml/c_backend.py
```diff
@@ -491,9 +488,35 @@ def generate_attr_getter(fname, node, port, dimsizes, cprefix, loopvars):
...
                               loopvars):
+    if (dml.globals.dml_version != (1, 2)):
+        check_attribute(node, port, prefix)
+        return
+        # parent = node.parent
+        # while parent.objtype not in ("device", "bank", "port", "subdevice"):
+        #     parent = parent.parent
+
+        # if parent.objtype == "device":
+        #     return
```
lwaern-intel, line 518: Trash, should be removed.

lib/1.4/dml-builtins.dml
```diff
@@ -568,21 +581,76 @@ session dml_qname_cache_t _qname_cache;
...
+extern typedef size_t ptrdiff_t;
+
+extern typedef struct {
+    bool valid;
+    bool is_bank;
+    bool is_array;
+    // uint32 array_size;
+    ptrdiff_t port_obj_offset;
+    const char *portname;
+} _attr_parent_obj_proxy_info_t;
```
mandolaerik, line 652: maybe use a longer consistent prefix for internal types, e.g. `_dml_`?

lib/1.4/dml-builtins.dml
```diff
@@ -568,21 +581,76 @@ session dml_qname_cache_t _qname_cache;
...
+    param _attr_type : const char *;
 }
 
+extern typedef size_t ptrdiff_t;
+
+extern typedef struct {
+    bool valid;
+    bool is_bank;
+    bool is_array;
+    // uint32 array_size;
```
mandolaerik, line 649: remove comment-out code

lib/1.4/dml-builtins.dml
```diff
@@ -568,21 +581,76 @@ session dml_qname_cache_t _qname_cache;
...
+template _conf_attribute is object {
+    shared method set_attribute(attr_value_t value) -> (set_error_t);
+    shared method get_attribute() -> (attr_value_t);
+
+    param _attr_name : const char *;
+    param _attr_name auto;
+    param _parent_obj_class : conf_class_t **;
+    param _parent_obj_class auto;
+    param _parent_obj_proxy_info : _attr_parent_obj_proxy_info_t;
+    param _parent_obj_proxy_info auto;
```
mandolaerik, line 593: I like your approach here. If I'd have written it I would probably just have given up on proxy attrs and kept them fully hard-coded; you manage to keep them soft-coded without sacrificing performance OR readability. Nice!

lib/1.4/dml-builtins.dml
```diff
@@ -32,7 +33,7 @@ param dml_1_4 = true;
 
 constant unmapped_offset = 0xffff_ffff_ffff_ffff;
 
-extern void *const NULL;
+param NULL auto;
```
mandolaerik, line 36: why isn't this needed in 1.2?

Interesting solution to use an autoparam for NULL. It means that you can't use NULL in early evaluation, but not much happens between extern and param. I guess this gives a regression in some weird corner, perhaps `typedef typeof(NULL) null_t;` goes from valid to invalid. I don't mind.

py/dml/structure.py
```diff
@@ -2394,6 +2437,173 @@ def mkexpr(self, indices):
...
+        self.node = node
+        self.cached = None
+
+    def mkexpr(self, indices):
+        if self.cached is not None:
+            return self.cached
+        parent = self.node.parent
+
+        while parent and parent.objtype not in ('bank', 'port', 'subdevice'):
+            parent = parent.parent
```
lwaern-intel, line 2492: Can be replaced with object_parent

py/dml/structure.py
```diff
@@ -2394,6 +2437,173 @@ def mkexpr(self, indices):
...
+        def no_valid_info():
+            return mkLit(
+                self.site,
+                '(_attr_parent_obj_proxy_info_t) { .valid = false }',
+                TNamed('_attr_parent_obj_proxy_info_t'))
+
+        parent = self.node.parent
+
+        while parent and parent.objtype not in ('port', 'bank'):
+            parent = parent.parent
```
lwaern-intel, line 2526: Can probably be replaced with object_parent

py/dml/structure.py
```diff
@@ -2394,6 +2437,173 @@ def mkexpr(self, indices):
...
+    def __init__(self, site, node):
+        self.site = site
+        self.node = node
+
+    def mkexpr(self, indices):
+        return mkNodeRef(self.site, self.node.object_parent,
+                         indices[:self.node.object_parent.dimensions])
+
+class NongroupParentParamExpr(objects.ParamExpr):
+    '''The _nongroup_parent parameter of any object'''
```
lwaern-intel, line 2596: Wrong! any object _except for the device_!

lib/1.4/dml-builtins.dml
lwaern-intel: :)

lib/1.4/dml-builtins.dml
```diff
@@ -568,21 +581,76 @@ session dml_qname_cache_t _qname_cache;
...
-    param readable default this.configuration != "none";
-    param writable default this.configuration != "none";
+template _conf_attribute is object {
+    shared method set_attribute(attr_value_t value) -> (set_error_t);
+    shared method get_attribute() -> (attr_value_t);
+
+    param _attr_name : const char *;
+    param _attr_name auto;
+    param _parent_obj_class : conf_class_t **;
+    param _parent_obj_class auto;
```
mandolaerik, line 591: Would in general be nice to comment on every `auto` why it's needed.

py/dml/c_backend.py
```diff
@@ -491,9 +488,35 @@ def generate_attr_getter(fname, node, port, dimsizes, cprefix, loopvars):
...
+    register_attribute(node.site, port, attrname)
+    if port and port.objtype in {'port', 'bank'}:
+        register_attribute(node.site, None, "%s_%s" % (port.name, attrname))
+
 # dimsizes, loopvars, prefix are relative to port.
 def generate_attribute_common(initcode, node, port, dimsizes, prefix,
                               loopvars):
+    if (dml.globals.dml_version != (1, 2)):
+        check_attribute(node, port, prefix)
+        return
```
mandolaerik, line 512: cleaner to move this to caller?

py/dml/c_backend.py
```diff
@@ -491,9 +488,35 @@ def generate_attr_getter(fname, node, port, dimsizes, cprefix, loopvars):
...
+    if not get_long_doc(node):
+        if (node.objtype in ['attribute', 'connect']
+            and config_param == 'required'):
+            report(WNDOCRA(node, node.logname()))
+        elif node.objtype != 'register':
+            report(WNDOC(node, node.logname()))
+    attrname = get_attr_name(prefix, node)
+    register_attribute(node.site, port, attrname)
+    if port and port.objtype in {'port', 'bank'}:
+        register_attribute(node.site, None, "%s_%s" % (port.name, attrname))
```
mandolaerik, line 505: I think this is technically inaccurate, the conditions for creating proxy attr is more complex; IIRC, attributes in a multidim port array or port in group does not get a proxy. It's a silly bug that we can live with (proxies will vanish eventually), so no need to write a test; however maybe code will look nicer if we extract the proxy precondition logic to a function?

py/dml/ctree.py
```diff
@@ -1270,12 +1270,19 @@ def make(cls, site, lh, rh):
                 rh_indices = [idx.value for idx in rh.indices]
                 return mkBoolConstant(site, (lh.node is rh.node
                                              and lh_indices == rh_indices))
+            if isinstance(lh, NullExpr) or isinstance(rh, NullExpr):
+                if isinstance(lh, NullExpr) and isinstance(rh, NullExpr):
+                    return mkBoolConstant(site, True)
+                elif (isinstance(lhtype, (TPtr, TArray))
+                      and isinstance(rhtype, (TPtr, TArray))):
+                    return mkBoolConstant(site, False)
```
mandolaerik, line 1278: this is incorrect? NULL can be equal to non-constant pointers.

Please add unit test to ctree_test!

py/dml/structure.py
```diff
@@ -2394,6 +2437,173 @@ def mkexpr(self, indices):
...
+    '''The _attr_name parameter of an _conf_attribute object'''
+    __slots__ = ('site', 'node', 'cached',)
+
+    def __init__(self, site, node):
+        self.site = site
+        self.node = node
+        self.cached = None
+
+    def mkexpr(self, indices):
+        if self.cached is not None:
```
mandolaerik, line 2463: Why not eager calculation? caching is a great root of evil, avoid it if you can.

py/dml/structure.py
```diff
@@ -2394,6 +2437,173 @@ def mkexpr(self, indices):
...
+            return self.cached
+        parent = self.node.parent
+        prefix = ""
+
+        while parent.objtype not in ('device', 'bank', 'port', 'subdevice'):
+            prefix = parent.name + "_" + prefix
+            parent = parent.parent
+
+        self.cached = mkStringConstant(self.site,
+                                       get_attr_name(prefix, self.node))
```
mandolaerik, line 2473: This is 1.4 only, where anonymization is softcoded, so no need for transformation.

py/dml/structure.py
mandolaerik: Note that the caching in ASTParamExpr is different: it saves a lot of time, and conditions for caching are complex, and they contain user-defined code, and the same param may give an error if evaluated too early, but be OK later on.

py/dml/ctree.py
lwaern-intel: String literals are constant. Method references converted to function pointers are constant. That's it.
Any PR which allows a pointer to be considered constant by DMLC _and_ can be NULL should be rejected wholesale.

py/dml/structure.py
lwaern-intel: I assume you're suggesting the eager calculation is to be done either in `make_autoparams` or in the `__init__`s of relevant ParamExprs?

I'm not entirely convinced regarding eager evaluation. Consider that my profiling does show that a few seconds are spent these on calculating these auto-optimized cached params. Though that number may be inflated by the profiling, I still wonder if we should risk spending unnecessary work if those params are left unused.

Besides, the complexity won't actually be reduced. All eager evaluation would do is to move all logic from `mkexpr` to `__init__`. Hell, you won't even eliminate the `cached` member as you need somewhere to store the result for later reference in `mkexpr` -- most you can do is rename the member and eliminate the `is not None` check.

py/dml/structure.py
lwaern-intel: I guess not... but if `get_attr_name` is supposed to be the canonical way to retrieve the C name for an attribute, circumventing it is Bad and Unidiomatic.

include/simics/dmllib.h
lwaern-intel: Eh, actually, `strbuf` isn't as clean as one would like since the loop doesn't simply append upon the string. Still, getting rid of the 1024 char cap would be nice. The code would look something like this:
```
        strbuf_t type = sb_newf("%s", attr_info.type)
        _id_info_t id_info = id_info_array[list.id];
        for (int64 i = (int64)id_info.dimensions - 1; i >= attr_info.start_dim;
             --i) {
            char *tmp_type = sb_detach(&type);
            sb_addfmt(&type, attr_info.allow_cutoff ? "[%s{0:%d}]" : "[%s{%d}]",
                    tmp_type, id_info.dimsizes[i]);
            MM_FREE(tmp_type);
        }
```

py/dml/c_backend.py
lwaern-intel: Oop, correct. I agree with moving the precondition out; the same condition is checkd within `ConfAttrParentProxyInfoParamExpr.mkexpr`.

include/simics/dmllib.h
lwaern-intel: In dmllib?

mandolaerik: I'd say simics core would be a more natural place. Not part of this PR.

py/dml/ctree.py
mandolaerik: Not obvious, consider `cast(0, foo_t*) == NULL`. Perhaps the `NullExpr` class should document that it's the only constant null-valued pointer expression?

py/dml/structure.py
mandolaerik: I agree that the damage is limited (the main cost is readability, the reader of those `ParamExpr`s doesn't understand why caching is needed). But the wins are also limited if not nonexistent, most of these ParamExprs exist because we know that they (almost) always are read at least once by dml-builtins, right?

So I cite Occam, please remove caching because why not remove it.

py/dml/ctree.py
lwaern-intel: If we add code to allow `cast(0, foo_t*)` to be constant we'd just add a case to `mkCast` that would create a `NullExpr`.
Now, a more interesting question would be `cast(1, foo_t *) - 1`!
For one, do we want non-zero pointer literals to be constant? My answer is "maybe, whatever."
For another, would we want pointer offsets of constant pointer be offset? My answer would be "certainly not offsets of pointer literals. Maybe for strings if we decide to do interning for string literals."

lib/1.4/dml-builtins.dml
lwaern-intel: I'm not sure that would contribute much. The answers would always be "we need this info for attribute softcoding, and its auto cause there's now way to retrieve it from within DML".

py/dml/ctree.py
mandolaerik: We should make a thing constant if it's technically feasible to do so and there is a real use case for it. Other things can also be considered constant if that leads to a simpler definition of constantness.

py/dml/expr.py
```diff
@@ -190,6 +191,20 @@ def writable(self):
 
 mkLit = Lit
 
+class NullExpr(Expression):
```
mandolaerik, line 194: Naming: other expression classes don't end with `Expr`. I suggest `Null` or `NullConstant` or `NullPointerConstant`

include/simics/dmllib.h
```diff
@@ -2560,12 +2593,144 @@ UNUSED static void _free_table(ht_int_table_t *table) {
         ht_clear_int_table(table, true);
 }
 
-UNUSED static void _memoized_recursion(const char *name) {
+UNUSED static void _signal_critical_error(const char *restrict format, ...) {
+
```
lwaern-intel, line 2597: what is this empty line eugh

py/dml/expr.py
lwaern-intel: `NullConstant` it is.

py/dml/structure.py
lwaern-intel: We've debated back and forth about this off-github. We've decided on keeping the caching, but adding a blurb to explain caching should be used with care.

py/dml/ctree.py
lwaern-intel: Now added documentation to `NullConstant` saying that expression class must be used for any null-valued pointer-typed expression considered constant by DMLC.

py/dml/c_backend.py
lwaern-intel: Done.

lib/1.4/dml-builtins.dml
lwaern-intel: Now added a general "every param below this line is for attribute softcoding", as well as a line explaining each param.

py/dml/structure.py
lwaern-intel: done

py/dml/structure.py
lwaern-intel: done

lib/1.4/dml-builtins.dml
lwaern-intel: Done, this one type was an outlier; every other one type uses `_dml_`.

py/dml/c_backend.py
lwaern-intel: Doesn't seem to cause any issues... Ripped out.

include/simics/dmllib.h
lwaern-intel: `strbuf` now used.

include/simics/dmllib.h
lwaern-intel: `generated` now replaced with `should_be_generated`. I prefer that to `skip`.

syssimics:
Verification [#10602539](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10602539): fail

py/__main__.py
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: MPL-2.0
 
 import sys
+import cProfile
```
mandolaerik, line 5: unused import?

lwaern-intel: Doh! Forgot to remove this.

py/dml/ctree.py
mandolaerik: Still no test in ctree_test.py?

py/dml/ctree_test.py
```diff
@@ -9,7 +9,7 @@
 import math
 from pathlib import Path
 
-from dml import ctree, types, logging, messages, output, symtab
+from dml import codegen, ctree, types, logging, messages, output, symtab
```
lwaern-intel, line 12: Oop. I imported this when I tried to construct `MethodFunc`s before quickly giving up. I'm removing this.

py/dml/ctree_test.py
```diff
@@ -9,7 +9,7 @@
 import math
 from pathlib import Path
 
-from dml import ctree, types, logging, messages, output, symtab
+from dml import  ctree, types, logging, messages, output, symtab
```
mandolaerik, line 12: excess space :P

lwaern-intel: aaaaAAAAAAAAA

lwaern-intel: I HAD THE EDIT ALREADY PREPARED IN THE EDITOR, BUT I DIDN'T SAVE THE FILE. WHY

lwaern-intel: mad

syssimics:
Verification [#10660926](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10660926): pass

mandolaerik (approved):

syssimics:
Verification [#10661572](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10661572): pass

syssimics:
Verification [#10673141](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10673141): pass

syssimics:
Verification [#10673805](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10673805): pass

test/1.2/structure/T_attribute_conf.py
```diff
@@ -3,7 +3,10 @@
 
 import stest
 
-obj = SIM_create_object('test', 'obj', [['b_a', 5]])
+obj = SIM_create_object('test', 'obj', [['b1_a', 5], ['b2_a', [0,1,2,3]]])
```
mandolaerik, line 6: note: you can also write `b1_a=5` if you like

test/1.2/structure/T_attribute_conf.py
```diff
@@ -3,7 +3,10 @@
...
+
+stest.expect_equal(obj.bank.b1.a, 5)
+stest.expect_equal([obj.bank.b2[i].a for i in range(4)], list(range(4)))
```
mandolaerik, line 9: note: `expect_true(all(b.a == i for (i, b) in enumerate(...)))` in case you like that better

mandolaerik (approved):

syssimics:
Verification [#10716730](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10716730): pass

PR #98, opened 2022-06-21
Don't expect <undefined> fields in register_view

mandolaerik:
Does not add value, and an upcoming bugfix in Simics Base solves the problem.


syssimics:
Verification [#10076860](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10076860): fail

mandolaerik:
Tightly coupled with https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/1940

syssimics:
Verification [#10079598](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10079598): fail

PR #99, opened 2022-06-22
dml-gdb touch-up, DML 1.4 support

lwaern-intel:


py/dml/c_backend.py
```diff
@@ -354,6 +354,7 @@ def generate_hfile(device, headers, filename):
         + '('+crep.structtype(device)+' *_obj);\n')
     out('void soft_reset_'+crep.cname(device)
         + '('+crep.structtype(device)+' *_obj);\n')
+    out('extern int _marker_' + crep.cname(device) + ';\n')
```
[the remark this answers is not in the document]
mandolaerik: Also, I think the `extern` decl is redundant, only needed for symbols referenced across modules.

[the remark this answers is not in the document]
lwaern-intel: It was, thanks for pointing that out.

py/dml/c_backend.py
```diff
@@ -354,6 +354,8 @@ def generate_hfile(device, headers, filename):
         + '('+crep.structtype(device)+' *_obj);\n')
     out('void soft_reset_'+crep.cname(device)
         + '('+crep.structtype(device)+' *_obj);\n')
+    out('int _dml_gdb_marker_'
```
mandolaerik, line 357: this declaration is still redundant, since you also have `generate_marker`? (Also, it would not surprise me if this causes `--split-c-file` to break)

lwaern-intel: Huh! Y'know, I was about to contradict you since I tested this earlier, but now I tested it again and sure enough, the declaration in the .h file is unnecessary. Weird.

About `--split-c-file`; yeah, I have no idea how this interacts with it, but I also wouldn't be surprised if it breaks. I'll test. If it doesn't work, how do we solve it? Perhaps we shouldn't? Add a documentation string and check stating that `--split-c-file` is incompatible with `-g`?

mandolaerik: It will or won't work, depending on how you use `--split-c-file`. The easy way is to have a single-class module and just link all .c files together; this should be unproblematic as long as you don't touch the .h file. This is what t126 tests. The harder way is to insert an extra link step that combines all .o files of one device into one big .o, while hiding all symbols except the `initialize_<classname>` function, to allow multiple DML classes in one module. The GDB symbol will vanish here unless you update the exportmap.

Speaking of which -- perhaps we should reuse the `initialize_*` function as dml-gdb marker instead of adding a new one?

lwaern-intel: > Speaking of which -- perhaps we should reuse the initialize_* function as dml-gdb marker instead of adding a new one?

That was the first thing I tried, didn't get it to work -- I figured either because the function's static and/or because it isn't declared in the .h file. Are you suggesting we change that?

py/dml/toplevel.py
```diff
@@ -301,9 +301,11 @@ def parse_main_file(inputfilename, explicit_import_path, strict):
...
     dml.globals.compat_dml12 = version == (1, 2) and not strict
 
-    implicit_imports = [
-        ast.import_(site, "dml-builtins.dml"),
-        ast.import_(site, "simics/device-api.dml")]
+    implicit_imports = (
+        [ast.import_(site, "dml-builtins.dml"),
+         ast.import_(site, "simics/device-api.dml")]
+        + ([ast.import_(site, "dml12-compatibility.dml")]
+           * (version == (1, 2) and dml.globals.debuggable)))
```
mandolaerik, line 308: I vote against this one

syssimics:
Verification [#10095499](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10095499): fail

syssimics:
Verification [#10097661](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10097661): fail

lwaern-intel:
Verification failure looks unrelated.

mandolaerik (approved):

PR #101, opened 2022-06-27
Log messages for unmapped accesses from unmapped_{read,write}

mandolaerik:
Apart from the effects mentioned in the release note, this also means
that the message can not be suppressed by bank_instrumentation.after_write


mandolaerik:
tightly coupled with https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/1966

syssimics:
Verification [#10099129](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10099129): fail

mandolaerik:
I just realized that I forgot to update the reference docs for unmapped_read etc.

lwaern-intel (approved):

PR #105, opened 2022-07-14
Repair IS_INT64 macro

mandolaerik:
It detected any 32-bit integer value as being int64


jhbaarnh (approved):
This also makes clang happier.

PR #106, opened 2022-08-26
Turn %b into a boolean format specifier -- SIMICS-19554

lwaern-intel:
There's still the question whether to make this DML 1.4 exclusive.

### DIFF
--- a/py/dml/codegen.py
+++ b/py/dml/codegen.py
@@ -855,7 +855,7 @@ def fix_printf(fmt, args, argsites, site):
             argi += 1
 
         arg = args[argi]
-        if conversion in "boudixX":
+        if conversion in "oudixX":
             # GCC emits warnings about ll vs l mismatches, even
             # though ll and l are both 64-bit
             # integers. Unfortunately, DMLC does not know the
@@ -890,6 +890,14 @@ def fix_printf(fmt, args, argsites, site):
                                               TPtr(TNamed('char',
                                                           const=True)))),
                               [mkAddressOf(site, arg)])
+        elif conversion == 'b':
+            argtype = safe_realtype(arg.ctype())
+            if not isinstance(argtype, TBool):
+                raise EFMTARGT(argsites[argi], arg,
+                                argi+1, "boolean")
+            conversion = 's'
+            arg = mkIfExpr(site, arg, mkStringConstant(site, "true"),
+                           mkStringConstant(site, "false"))
 
         filtered_fmt += "%" + flags + width + precision + length + conversion
         filtered_args.append(arg)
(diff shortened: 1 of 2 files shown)


py/dml/codegen.py
```diff
@@ -890,6 +890,14 @@ def fix_printf(fmt, args, argsites, site):
                                               TPtr(TNamed('char',
                                                           const=True)))),
                               [mkAddressOf(site, arg)])
+        elif conversion == 'b':
+            argtype = safe_realtype(arg.ctype())
+            if not isinstance(argtype, TBool):
+                raise EFMTARGT(argsites[argi], arg,
+                                argi+1, "boolean")
+            conversion = 's'
```
lwaern-intel, line 898: An alternative to doing this and relying on the line below (which can be ruined by bad flags/width/precision/length specification) is:
```
filtered_args.append(...)
filtered_fmt += "%s"
continue
```
Is that preferable?

JonatanWaern: I think this is an extremely non-confusing way to do it, I'd say its fine

JonatanWaern (approved):
Needs releasenote

JonatanWaern:
> Needs releasenote

wait, its 4:50pm and Im tired. do we do releasenotes now that we're our own repo?

py/dml/codegen.py
lwaern-intel: Which one? :)

lwaern-intel:
We still need releasenotes, I just kind of forgot about it.

mandolaerik (commented):
Unsure about the semantics here (strings `"true"` and `"false"` reflect the DML identifiers, but the reader is not a DML coder). I think a more obvious first step would be to extend `%d` to print 0/1 for bools; we have a clear and successful precedent that `"%d", any_integer_value` is magically accepted, and we freely reinterpret `bool` as a `uint1` elsewhere.

mandolaerik:
There were some discussions outside GH on this topic. The addition of `%b` is questionable, mainly because `"true"` and `"false"` is typically not what you want to display if the message is directed to the user; currently we force the user to choose which usually gives you something like `b?"1":"0"` or `b?"raised":"lowered"` or `b?"yes":"no"`; quick grep suggests that we pick `b?"true":"false"`  in <10% of the cases. So adding `%b` would encourage users to write worse messages. The strongest remaining use case for `%b` would be more convenient temporary debug printouts, but if you want easy-to-write debug print for bools then you are welcome to roll your own `method lb(bool b) { log info: "%s", b?"true":"false"; }`.

mandolaerik (changes requested):
_

PR #111, opened 2022-09-06
Generate deinit before processing method queue -- SIMICS-19598

lwaern-intel:


syssimics:
Verification [#10338771](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10338771): pass

mandolaerik (approved):

PR #112, opened 2022-09-07
Correct parts of old BZ xfail references

JonatanWaern:
- Update BZ references to JIRA references
- Port test and create issue for BZ 4582


test/1.4/structure/T_object_naming.dml
```diff
@@ -0,0 +1,23 @@
+/*
+  © 2021-2022 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+dml 1.4;
+device test;
+
+// bug 4582: Register names appear as identifiers in generated C code, causing
```
mandolaerik, line 8: SIMICS-19603, not 4582. Also remove the 1.2 variant of the test.

JonatanWaern: I suspect fixing it in 1.4 may not also fix it in 1.2, but if we don't care (which we probably dont) I am fine with discarding the 1.2 test

mandolaerik (approved):

PR #117, opened 2022-09-13
Don't generate #line directives for SimpleSite

mandolaerik:
- Revert previous SimpleSite change
- Polish site strings
- Don't generate #line directives for SimpleSite

### DIFF
--- a/py/dml/logging.py
+++ b/py/dml/logging.py
@@ -276,17 +276,14 @@ class Site(metaclass=abc.ABCMeta):
 class SimpleSite(Site):
     '''A variant of Site without line information. Useful for code
     that doesn't come from a real DML file'''
-    __slots__ = ('_name', '_line', '_dml_version')
-    def __init__(self, name, line=None, column=None, dml_version=(1, 4)):
+    __slots__ = ('_name', '_dml_version')
+    def __init__(self, name, dml_version=(1, 4)):
         self._name = name
         self._dml_version = dml_version
-        self._line = '' if line is None else f':{line}'
-        if column is not None:
-            self._line += f':{column}'
     def __repr__(self):
-        return f'<site {self.loc()}>'
+        return '<site %s>' % (self._name,)
     def loc(self):
-        return f'{self._name}{self._line}'
+        return self._name
     def filename(self):
         return self._name
     def dml_version(self):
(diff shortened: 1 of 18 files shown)


syssimics:
Verification [#10373509](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10373509): fail

py/dml/toplevel.py
```diff
@@ -70,7 +70,7 @@ def determine_version(filestr, filename):
         column = ver_start - filestr.rfind('\n', 0, ver_start)
         m = check_version.match(filestr, pos=ver_start)
         if not m:
-            raise ESYNTAX(SimpleSite(filename, lineno, column),
+            raise ESYNTAX(SimpleSite("%s:%d:%d" % (filename, lineno, column)),
```
lwaern-intel, line 73: You're being inconsistent about the use of f-strings vs `%` when creating `SimpleSite`s, I'm ok with either alternative, but alternating between both in the same PR for the same use-case is very strange to me.

py/dml/logging.py
```diff
@@ -276,17 +276,14 @@ def bitorder(self): pass
...
+    def __init__(self, name, dml_version=(1, 4)):
         self._name = name
         self._dml_version = dml_version
-        self._line = '' if line is None else f':{line}'
-        if column is not None:
-            self._line += f':{column}'
     def __repr__(self):
-        return f'<site {self.loc()}>'
+        return '<site %s>' % (self._name,)
     def loc(self):
```
lwaern-intel, line 285: Why move from f-string to `%`?

lwaern-intel (changes requested):
Address the `%` vs f-string dilemma and pick one.

py/dml/toplevel.py
mandolaerik: This comes from a commit that is a pure revert, I didn't bother to preserve the f-string migration parts of the reverted commit.

mandolaerik: (in general, f-string should be the default choice that should be used unless there's a good reason why % is better.)

syssimics:
Verification [#10378786](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10378786): pass

syssimics:
Verification [#10379061](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10379061): pass

lwaern-intel (approved):

PR #119, opened 2022-09-14
Improve documentation configuration param, SIMICS-18590

mandolaerik:


syssimics:
Verification [#10376268](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10376268): pass

lib/1.4/dml-builtins.dml
```diff
@@ -893,10 +893,10 @@ configuration [`"required"` | `"optional"` | `"pseudo"` | `"none"`]
...
 Specifies how Simics treats the automatically created
 [attribute](#attribute-objects)
-corresponding to the connect object. Note that for array objects,
-setting this parameter to `"optional"` has the effect that each
-instance of the connect can be individually set to a nil value.
-</dd><dt>
+corresponding to the connect object. The default value is `"optional"`.
+
+The attribute can be set to a nil value only if this parameter is `"optional"`
+or `"required"`. In an array of connects, this applies element-wise.</dd><dt>
```
lwaern-intel, line 899: ... _or_ required? Unless I'm being very dumb right now, that's not true.

JonatanWaern (changes requested):
We should similarly specify the configuration default for registers, as they also inherit directly from _conf_attribute.

lib/1.4/dml-builtins.dml
mandolaerik: Thanks! I meant `"pseudo"`.

PR #120, opened 2022-09-14
Add '...' syntax to allow partial designated initializers -- SIMICS-18705

lwaern-intel:


doc/1.4/language.md
```diff
@@ -1827,8 +1835,16 @@ C99-style designated initializers are supported for `struct`, `layout`, and
...
+Unlike C, partial initialization is not allowed implicitly; a designated
+initializer for each member must be specified.
+However, partial initialization can be done explicitly through the use of
+trailing `...` syntax:
+```
+session struct_t s = { .y = { .i = 2, ... }, ... }
+```
+
+Also unlike C, designator lists are not supported, and designated initializers
+for arrays are not supported.
```
mandolaerik, line 1847: Feels like we want to write "Initializers follow the same pattern as for `session` variables".
Or extract docs for initializer syntax to a separate section; in the current patch we don't document that partial designated initializers are allowed in e.g. `local` or `return` statements.

lwaern-intel: ugh, you're right. I think we've even said previously that with the increasing complexity to initializers and places we can use them, we're going to want a dedicated section. I feel that's a separate issue, though, deserving of its own JIRA and PR.

py/dml/codegen.py
```diff
@@ -1158,18 +1158,20 @@ def check_designated_initializers(site, etype, init_asts):
...
                                   + f"{etype}"
                                   for field in bad_fields)
-                        + ("\nmissing initializer(s) for field(s) "
+                        + ("\nmissing initializer(s) for member(s) "
                            + ", ".join(f"'{field}'"
                                        for field in remaining)
-                           if remaining else '')
+                           + "\n(partial initialization must be explicit "
+                           + "through trailing '...', e.g. '{.a = a, ...}')"
+                           if not allow_partial and remaining else '')
```
mandolaerik, line 1174: Consider `Blah blah.` instead of `(blah blah)`

lwaern-intel: I did! I picked `(blah blah)` so the user could not confuse this string with a sub-error, as missing members, duplicate members, and bad members sub-errors are presented through each category being on their own line.

doc/1.4/language.md
mandolaerik: ok

mandolaerik (approved):
don't forget to file a jira on initializer breakout if you don't fix it now

lwaern-intel:
Initializer documentation JIRA: https://jira.devtools.intel.com/browse/SIMICS-19643

PR #123, opened 2022-09-21
Only report None tokens as EOF

mandolaerik:
SIMICS-19465

### DIFF
--- a/py/dml/dmlparse.py
+++ b/py/dml/dmlparse.py
@@ -2512,15 +2512,15 @@ def reserved(t):
 # Error handling
 @prod
 def error(t):
-    if t and t.value:
-        value = str(t.value)
-        raise ESYNTAX(DumpableSite(t.lexer.file_info, t.lexpos), value, None)
-    else:
+    if t is None:
         # raise ESYNTAX(...) would have made sense here, but there
         # is no clean way to retrieve the file (because t is
         # None). Therefore, push the error reporting responsibility to
         # whoever invokes the parser.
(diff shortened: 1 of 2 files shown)


py/dml/dmlparse.py
```diff
@@ -2512,15 +2512,15 @@ def reserved(t):
...
-        raise ESYNTAX(DumpableSite(t.lexer.file_info, t.lexpos), value, None)
-    else:
+    if t is None:
         # raise ESYNTAX(...) would have made sense here, but there
         # is no clean way to retrieve the file (because t is
         # None). Therefore, push the error reporting responsibility to
         # whoever invokes the parser.
         raise UnexpectedEOF()
+    else:
+        value = str(t.value)
```
lwaern-intel, line 2522: What value did the previous `t.value` check have? You no longer perform it, so we must consider if it contributed anything.

mandolaerik: Likely no value. We should not preserve things out of superstition. Worst case is if t.value would be a false value that doesn't display nicely, and in this case we will move from an anonymous EOF error to an error that points at the correct token but presents that token as garbage. That's an improvement.

lwaern-intel: I'm inclined to agree.

syssimics:
Verification [#10414297](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10414297): pass

JonatanWaern (approved):

PR #124, opened 2022-09-21
Clarify documentation for the `destroy` method

mandolaerik:
SIMICS-19144

### DIFF
--- a/lib/1.4/dml-builtins.dml
+++ b/lib/1.4/dml-builtins.dml
@@ -3452,8 +3452,16 @@ nil.
 <small>destroy(void \*data)</small>
 </dt><dd>
 
-This method is called when an event is removed from the event
-queue
+This method is called on any posted events when the device object is deleted.
+<!-- (and also on `SIM_event_cancel_*`, but that function is not exposed for
+custom events so no need to mention) -->
+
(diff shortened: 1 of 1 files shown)


lib/1.4/dml-builtins.dml
```diff
@@ -3452,8 +3452,16 @@ nil.
 <small>destroy(void \*data)</small>
 </dt><dd>
 
-This method is called when an event is removed from the event
-queue
+This method is called on any posted events when the device object is deleted.
+<!-- (and also on `SIM_event_cancel_*`, but that function is not exposed for
+custom events so no need to mention) -->
```
lwaern-intel, line 3457: Heh. Double-check that nothing breaks and that this comment doesn't show up in the rendered documentation.

lwaern-intel (approved):
Looks good, I just want you to double-check that the resulting documentation looks good if you haven't already.

syssimics:
Verification [#10410703](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10410703): pass

lib/1.4/dml-builtins.dml
mandolaerik: The comment is still there in the generated HTML, but does not get rendered.

PR #128, opened 2022-09-30
Allow casts from any template type to object -- SIMICS-19950

lwaern-intel:


syssimics:
Verification [#10456647](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10456647): pass

lib/1.4/dml-builtins.dml
```diff
@@ -54,6 +54,14 @@ extern typedef _callback_entry_t * vect _callback_vect_t;
...
 
 extern typedef struct {} _id_info_t;
+extern typedef struct {
+    uint32 id;
+    uint32 encoded_index;
+} _identity_t;
+extern typedef struct {
+    void *trait;
+    _identity_t id;
+} _traitref_t;
```
mandolaerik, line 65: can we hide these refs and send &this to dmllib in a void* arg instead? or at least make the traitref_t type opaque?

py/dml/traits.py
```diff
@@ -348,8 +348,11 @@ def mktrait(site, tname, ancestors, methods, params, sessions,
     # referencing 'dev.xyz' from a shared method is always OK, even if
     # it's technically an untyped object parameter
     reserved_symbols.pop('dev', None)
-    return Trait(site, tname, ancestors, methods, params, sessions,
-                 ancestor_vtables, ancestor_method_impls, reserved_symbols)
+    trait = Trait(site, tname, ancestors, methods, params, sessions,
+                  ancestor_vtables, ancestor_method_impls, reserved_symbols)
+    if tname == 'object':
+        dml.globals.object_trait = trait
```
mandolaerik, line 354: side-effect, please move upward, probably to mkglobals

mandolaerik (changes requested):

lib/1.4/dml-builtins.dml
lwaern-intel: For this PR? Sure. For attribute softcoding? _No_. With that in mind I don't see the point.

py/dml/traits.py
lwaern-intel: That _might_ be too late, but I can try.

lwaern-intel: No issue!

lib/1.4/dml-builtins.dml
mandolaerik: Maybe make it an opaque type with accessor functions for the stuff attr softcoding needs? (I haven't reviewed that one yet so I don't really know what I'm talking about)

lwaern-intel: attr softcoding needs construction functions -- a way to create an `attribute` out of `id`, `encoded_index` and `trait`. If you have accessor functions for that, why bother making the type opaque.

Actually, in general, why bother making the type opaque. Don't quote Hyrum's at me, how would accessor functions be any better?

include/simics/dmllib.h
```diff
@@ -731,7 +731,8 @@ _deserialize_trait_reference(ht_str_table_t *id_info_ht,
     if (unlikely(error != Sim_Set_Ok)) {
         return error;
     }
-    void *trait = ht_lookup_int(vtable_ht, id.id);
+    void *trait = vtable_ht ? ht_lookup_int(vtable_ht, id.id)
+        : (void*)(uintptr_t)0XC0FFEE;
```
mandolaerik, line 735: there's nobody who will drink the coffee? isn't it better to ASSERT, or fail cleanly?

lwaern-intel: No, this is not a case of an error. This is the special case we discussed earlier; empty vtables have no hashtables, and since they're empty the vtable pointer of a trait reference for such a trait is never utilized. Thus, we need some arbitrary _non-`NULL`_ pointer for the vtable pointer during deserialization of trait references with empty trait. Non-`NULL` is important as trait reference serialization checks for a `NULL` vtable pointer to detect a zero-initialized trait reference. We could pick something like `0x1`, but I don't want someone debugging to mistake it for an offset of a null-pointer if they spot it. So I picked `0XC0FFEE`.

mandolaerik: Right. Can we `#define _DML_EMPTY_VTABLE 0xC0FFEE` with a comment?

lwaern-intel: Sure!

mandolaerik (approved):

PR #130, opened 2022-10-13
Specify that methods are static

pshishpo-intel:
The fix addresses also the following warning emitted by Pylint Python linter:

ctree.py:1082:40: E1120: No value for argument 'rh' in staticmethod call (no-value-for-parameter)


mandolaerik (approved):

syssimics:
Verification [#10516694](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10516694): pass

PR #133, opened 2022-10-24
Fix ICE from trait reference initializers -- SIMICS-20145

lwaern-intel:


py/dml/globals.py
```diff
@@ -24,6 +24,9 @@
 # Array of all declared composite objects
 objects = []
 
+# Array of all session/saved variables declared in methods
```
mandolaerik, line 27: `... as pairs (Symbol, init), where init is str or None`

syssimics:
Verification [#10568707](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10568707): pass

py/dml/objects.py
```diff
@@ -303,9 +302,6 @@ def get_unique_static_name(self, name):
         self.static_idx = self.static_idx + 1
         return 'static%d_%s' % (self.static_idx, name)
```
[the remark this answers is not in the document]
lwaern-intel: uhhhhhhhhhhhhhhh. I didn't consider this function and `static_idx`. I'll double-check if your suggested logic works, but also we should really break out this function, as it has no place in `Device` anymore.

py/dml/c_backend.py
```diff
@@ -1597,12 +1597,10 @@ def generate_init_static_vars(device):
...
-    with crep.DeviceInstanceContext():
-        for v in device.staticvars:
-            if v.init:
-                assert isinstance(v.init, Initializer)
-                var = mkStaticVariable(v.site, v)
-                v.init.assign_to(var, var.sym.type)
+    for (_, init) in dml.globals.static_vars:
+        if init:
+            for line in init.splitlines(keepends=True):
+                out(line)
```
mandolaerik, line 1603: I think `out` handles multiline strings correctly, no need to split

py/dml/objects.py
```diff
@@ -283,29 +283,20 @@ def set_traits(self, traits):
         self.traits = traits
 
 class Device(CompositeObject):
-    __slots__ = ('staticvars', 'initdata', 'static_idx')
+    __slots__ = ('initdata')
```
mandolaerik, line 286: I prefer tuple form, `('initdata',)`. IIRC python has a hack to support string form, if you insist on this then please remove brackets.

lwaern-intel: Oh, oops, yeah, you're quite right, I simply forgot.

py/dml/c_backend.py
lwaern-intel: Nope, it doesn't. It fails at indenting them properly.

syssimics:
Verification [#10572762](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10572762): pass

py/dml/c_backend.py
lwaern-intel: Check out `Output.out`, it's actually rather obvious when looking at the code.

... I suppose your next comment is "fix `Output.out` instead then". Ergh, I guess I could _try_, but I don't know how much of a pain that would be or what kind of ramifications it would have.

mandolaerik: OK, fair enough, it seems to support printing part of a line but not multiple lines.
Multiline `out` is uncommon so no need to fix.

mandolaerik (approved):

PR #134, opened 2022-10-25
Check that we don't write -- in markdown

mandolaerik:
- Check that we don't write -- in markdown

### DIFF
--- a/validate_md_links.py
+++ b/validate_md_links.py
@@ -95,6 +95,11 @@ def main():
                 line = md_files[path][:match.start()].count('\n') + 1
                 sys.stderr.write(f'{path}:{line}: error: {message}\n')
                 ok = False
+        for (i, line) in enumerate(md_files[path].split('\n')):
+            if ' -- ' in f' {line} ':
+                sys.stderr.write(
+                    f'{path}:{i+1}: error: replace -- with &mdash;\n')
+                ok = False
     sys.exit(0 if ok else 1)
 
 if __name__ == '__main__':
--- a/doc/1.4/language.md
+++ b/doc/1.4/language.md
@@ -1769,8 +1769,8 @@ session conf_object_t *obj;
 In the absence of explicit initializer expressions, a default
 "all zero" initializer will be applied to each declared object.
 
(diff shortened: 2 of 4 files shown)


syssimics:
Verification [#10575984](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10575984): fail

lwaern-intel:
heh. the `--` found in verification is also my fault.

validate_md_links.py
```diff
@@ -95,6 +95,11 @@ def main():
                 line = md_files[path][:match.start()].count('\n') + 1
                 sys.stderr.write(f'{path}:{line}: error: {message}\n')
                 ok = False
+        mdash = md_files[path].find(' -- ')
```
lwaern-intel, line 98: Consider also lines beginning or ending with `--`.

mandolaerik: done

syssimics:
Verification [#10586886](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10586886): pass

validate_md_links.py
```diff
@@ -95,6 +95,11 @@ def main():
                 line = md_files[path][:match.start()].count('\n') + 1
                 sys.stderr.write(f'{path}:{line}: error: {message}\n')
                 ok = False
+        for (i, line) in enumerate(md_files[path].split('\n')):
+            if ' -- ' in f' {line} ':
+                sys.stderr.write(
+                    f'{path}:{i+1}: error: replace -- with &mdash;\n')
+                ok = False
     sys.exit(0 if ok else 1)
```
lwaern-intel, line 103: `sys.exit(1 * ok)`! _Embrace_ the weakly-typed heresy!

<sub><sub>do not actually do this</sub></sub>

lwaern-intel (approved):

validate_md_links.py
mandolaerik: that would be `1-ok`, right?

One alternative representation that actually would make a bit of sense, is to go all-in on integers and rename `ok` to `exitcode`.

PR #135, opened 2022-10-27
Reverse order of conditions

mandolaerik:
This avoids unconditionally feeding the source file to the lexer,
which hindered a weird non-DML use case of mine.


syssimics:
Verification [#10586880](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10586880): pass

lwaern-intel:
alright.

lwaern-intel (approved):

PR #136, opened 2022-11-01
Fix trait reference params causing broken debug files -- SIMICS-20200

lwaern-intel:


py/dml/g_backend.py
```diff
@@ -39,8 +39,12 @@ def enc(expr):
             val = expr.value
             if isinstance(val, list):
                 return [enc(e) for e in val]
-            else:
+            elif isinstance(val, (int, str, bytes, float, bool)):
                 return val
+            else:
```
JonatanWaern, line 44: Can refactor this slightly to discard the duplicate else-cases. (drop the else from the return on 49, and drop the entire else clause on 44.

Tbh, Im not sure if this is an anti-pattern (leads to less readable code, but also eliminates some code duplication)

test/1.4/misc/test_dmlc_g.dml
```diff
@@ -0,0 +1,44 @@
...
+    }
+}
+
+connect pic {
+    interface sample;
+}
+
+event ev_cycles is simple_cycle_event {
+    method event() { }
+}
```
JonatanWaern, line 44: this test could be more covering of types, I think. In particular Im curious to see what happens if we output debug info for endian integers.

test/1.4/misc/test_dmlc_g.py
```diff
@@ -0,0 +1,69 @@
...
+            for j in i[3]:
+                if j[1] == 'event':
+                    t_method = True
+
+    for flag in (t_device, t_data, t_attribute, t_parameter,
+                 t_method, t_event, t_bank, t_group, t_register,
+                 t_field, t_port, t_implement, t_connect, t_interface):
+        stest.expect_true(flag)
+
+    stest.expect_equal(t_data_type, 'double')
```
JonatanWaern, line 64: t_data_type and similar might be uninitialized here, which causes ugly fails

JonatanWaern (changes requested):
Only changes _needed_ imo is the t_data_type (and similar) in the py test

test/1.4/misc/test_dmlc_g.dml
lwaern-intel: Interesting thought, I don't actually know how those are handled. t102 doesn't even test it.

py/dml/g_backend.py
lwaern-intel: Fisketur suggests even more extreme refactorings which I _don't_ like, but I don't mind reducing the duplicate else-cases.

syssimics:
Verification [#10607986](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10607986): pass

JonatanWaern (approved):

syssimics:
Verification [#10608738](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10608738): pass

PR #140, opened 2022-11-22
Make vtables live when the template type is serialized -- SIMICS-20301

lwaern-intel:
I'm not overly happy with the implementation.

### DIFF
--- a/RELEASENOTES-1.4.docu
+++ b/RELEASENOTES-1.4.docu
@@ -425,4 +425,9 @@
       such objects are always reported by the compiler, even if the
       <tt>configuration</tt> parameter is defined to be <tt>"none"</tt>.
   </add-note></build-id>
+  <build-id value="next"><add-note>Fixed a bug where leveraging serialization
+      of a template type could cause an internal compiler error if there is
+      some object in the model instantiating that template which never gets
+      referenced in a value of that template type <bug number="SIMICS-20301"/>.
+  </add-note></build-id>
 </rn>
--- a/py/dml/structure.py
+++ b/py/dml/structure.py
@@ -16,6 +16,7 @@ from . import toplevel
 from . import topsort
 from . import slotsmeta
 from . import ctree
+from . import serialize
 from .logging import *
 from .codegen import *
 from .symtab import *
@@ -806,9 +807,12 @@ def mksaved(spec, parent):
     if astinit:
         dml.globals.device.add_init_data(obj)
     try:
-        realtype(crep.node_storage_type(obj))
+        typ = realtype(crep.node_storage_type(obj))
     except DMLUnknownType as e:
         raise ETYPE(obj, e.type)
+    else:
+        serialize.mark_for_serialization(site, typ)
+
     return obj
 
 def register_fields(reg):
--- a/test/1.4/serialize/T_saved_dead_vtable.dml
+++ b/test/1.4/serialize/T_saved_dead_vtable.dml
@@ -1,0 +1,38 @@
+/*
+  © 2022 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+dml 1.4;
+
+device test;
+
+/// COMPILE-ONLY
+
+template t {
+    shared method m();
+}
+
+template u {
+    shared method n();
+}
+template uninstantiated_1 {
+    shared method o();
+}
+template uninstantiated_2 {
+    shared method p();
+}
+
+is t;
+
+group g is u {
+    method n() {}
+}
+
+saved t object_saved;
+saved uninstantiated_1 object_saved_uninstantiated;
+
+method m() {
+    saved u method_saved;
+    saved uninstantiated_2 method_saved_uninstantiated;
+}
+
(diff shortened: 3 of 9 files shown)


syssimics:
Verification [#10711178](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10711178): pass

py/dml/traits.py
```diff
@@ -618,6 +618,9 @@ def __init__(self, site, name, ancestors, methods, params, sessions,
         self.name = name
         self.site = site
 
+        # Set of CompositeObject
+        self.instances = set()
```
mandolaerik, line 622: can we move this to a global mapping `{Trait: {CompositeObject}}` instead? The `Trait` class is currently immutable and I'd prefer to keep it that way.

lwaern-intel: I'm deliberating between
```
class Trait(SubTrait):
   instances = {}
```
and `dml.globals.trait_instances`. Your opinion?

test/1.4/serialize/T_saved_dead_vtable.dml
```diff
@@ -0,0 +1,29 @@
...
+/// COMPILE-ONLY
+
+template t {
+    shared method m();
+}
+
+template u {
+    shared method n();
+}
+
```
mandolaerik, line 18: add a saved uninstantiated template? I suppose this use case is why you write `trait_instances.get(x, ())` instead of `trait_instances[x]`?

test/1.4/serialize/T_saved_dead_vtable.dml
```diff
@@ -0,0 +1,29 @@
...
+group g is u {
+    method n() {}
```
mandolaerik, line 20: Maybe add a negative test case, like
```
trait x { shared method m(); }
trait y { shared method n(); }
/// WERRSTMT
group x is x { method m() { error; }}
// no WERRSTMT because unused (weird)
group y is y { method n() { error; }}
saved x z;
```

lwaern-intel: That's already checked in `T_trait_dce`.

test/1.4/serialize/T_saved_dead_vtable.dml
lwaern-intel: Sure, that's reasonable.

lwaern-intel: And _am I glad you pointed that out_, it revealed a bug.

test/1.4/serialize/T_saved_dead_vtable.dml
lwaern-intel: Actually, no, it isn't. I'd rather update `T_trait_dce`.

test/1.4/serialize/T_saved_dead_vtable.dml
lwaern-intel: Bug was that `_tinit_...` functions were generated despite never being used, which then gcc complains about. Fixed by not marking traits as referenced ourselves, but instead relying on `ObjTraits.mark_referenced` to do it for us.

py/dml/traits.py
lwaern-intel: Discussed on R.C. `dml.globals` is the way to go.

syssimics:
Verification [#10717295](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10717295): pass

py/dml/structure.py
```diff
@@ -806,9 +807,12 @@ def mksaved(spec, parent):
     if astinit:
         dml.globals.device.add_init_data(obj)
     try:
-        realtype(crep.node_storage_type(obj))
+        typ = realtype(crep.node_storage_type(obj))
     except DMLUnknownType as e:
         raise ETYPE(obj, e.type)
+    else:
+        serialize.map_dmltype_to_attrtype(site, typ)
```
mandolaerik, line 814: We want to eventually make all traits always alive, and then this will be removed. I suggest open a jira for this and sprinkle greppable comments `# should be removed when SIMICS-NNN is resolved` where applicable, so we have a chance to clean it up afterward.

syssimics:
Verification [#10741023](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10741023): fail

py/dml/structure.py
lwaern-intel: invalidated by 54ee113

RELEASENOTES-1.4.docu
```diff
@@ -412,4 +412,9 @@
       that leverages <tt>init_val</tt> in order to initialize the <tt>val</tt>
       variable provided by the template <bug number="SIMICS-20108"/>.
   </add-note></build-id>
+  <build-id value="next"><add-note>Fixed a bug where leveraging serialization
+      of a template type could cause an internal compiler error if there is
+      some object in the model instantiating that template which never gets
+      referenced in a value of that template type.<bug number="SIMICS-20301"/>.
```
mandolaerik, line 418: aww, `s/type.<bug/type <bug/`

lwaern-intel: :(

PR #141, opened 2022-11-28
Fix bug in DMLC_DUMP_INPUT_FILES

mandolaerik:
If a device has no relative imports with `..`, then we created
self-referencing symlinks because we couldn't distinguish between
directories `''` and `'.'`. `normpath` solves the problem.

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -829,10 +829,15 @@ class DMLCProfileTestCase(CTestCase):
                 raise TestFail(f'cannot load profile data')
         else:
             raise TestFail(f'stats file not generated')
+
+
 class DumpInputFilesTestCase(CTestCase):
     '''Test that the DMLC_DUMP_INPUT_FILES variable works.  It creates a
     tarball of input DML files that can be compiled standalone.'''
-    __slots__ = ()
+    __slots__ = ('prefix',)
+    def __init__(self, *args, prefix='', **kwargs):
+        self.prefix = prefix
+        super().__init__(*args, **kwargs)
     def test(self):
         super().test()
         dir = Path(self.scratchdir) / 'dumped'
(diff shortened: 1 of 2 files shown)


test/tests.py
```diff
@@ -841,12 +844,13 @@ def test(self):
                           'r:xz') as tf:
             tf.extractall(dir)
         if is_windows():
-            assert (dir / '_' / os.path.basename(self.filename)).is_file()
+            assert (dir.joinpath(*self.prefix)
```
lwaern-intel, line 847: This spreading seems wrong. `self.prefix` is a string, meaning each character of the prefix will become a separate argument to `joinpath`.

syssimics:
Verification [#10742161](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10742161): pass

mandolaerik:
[test W](https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=10746338)

test/tests.py
mandolaerik: good catch, fixed

syssimics:
Verification [#10746331](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10746331): pass

syssimics:
Linux &amp; Windows (fast) [#10746338](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10746338): pass

test/tests.py
```diff
@@ -829,25 +829,31 @@ def test(self):
...
+        assert (dir.joinpath(*self.prefix)
+                / os.path.basename(self.filename)).is_file()
+        if not is_windows():
             # This does not work on Windows, for unknown reasons.
             # Seems related to symlink semantics somehow, but no
             # need to explore deeper until we have a use case for it.
-            cmd = dmlc + ['_/' + os.path.basename(self.filename),
-                          self.shortname]
+            cmd = dmlc + [
+                join(*self.prefix + [os.path.basename(self.filename)]),
```
lwaern-intel, line 855: Sooo the precedence rules will parse this as `*(self.prefix + [os.path.basename(self.filename)])`?
That's... not very intuitive, even though I guess it makes sense. Python's weird.

mandolaerik: True, I now added a parenthesis and that made it easier to read.

syssimics:
Verification [#10758774](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10758774): fail

syssimics:
Verification [#10758831](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10758831): pass

lwaern-intel (approved):

PR #142, opened 2022-12-02
register_view: unique names for fields in arrays -- SIMICS-18244

lwaern-intel:


mandolaerik (approved):
Fixing the bug might clash with Sven's workarounds for the bug.

lwaern-intel:
let's see what verification says.

PR #143, opened 2022-12-05
Support `.len` for any expression of sequence type -- SIMICS-20278

lwaern-intel:

PR #144, opened 2022-12-05
Hooks and after-on-hooks -- SIMICS-20358

lwaern-intel:
Bam-boom.

Depends on #143.

After-on-hooks still needs an illustrative example; the documentation in the reference manual is complete, but it's not enough to paint where hooks and after-on-hooks can be useful. The documentation is enough for @gwikland to read and understand what hooks _are_, though if we want him to actually experiment with this and come back with feedback I'd wager he want us to show him how this can applied. My thinking is for the moment we'll do that on R.C., and perhaps from those discussions we can conjure a good example to put on Stack Overflow or something once this is merged.


syssimics:
Verification [#10779260](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10779260): pass

syssimics:
Verification [#10779643](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10779643): pass

test/1.4/hooks/T_cancellation.dml
```diff
@@ -0,0 +1,66 @@
...
+    // Three `after h: m(domain)` are done per domain...
+    assert h.suspended == domains.len * 3;
+    // ... but even in spite of this, and that each m(domain) performs four
+    // additional `after h: m(domain)`, the cancellation performed in `m`
+    // causes the domain's pending second and third after of the send_now in
+    // progress to be cancelled, and also cancels two of the afters that it
+    // enqueues.
+    local uint64 resumed = h.send_now();
+    assert resumed == domains.len && h.suspended == domains.len * 2;
+    // There are two elements in the after queue per domain
```
lwaern-intel, line 59: *hook queue

py/dml/c_backend.py
```diff
@@ -1327,14 +1499,43 @@ def generate_init_identity_hashtable():
...
 
+def generate_hook_auxiliary_info_array():
+    items = []
+    for hook in dml.globals.hooks:
+        offset = ('offsetof(%s, %s)'
+                  % (crep.structtype(dml.globals.device),
+                     crep.cref_hook(
+                         hook, (mkIntegerConstant(hook.site, 0, False),)
+                         * hook.dimensions)))
+
```
lwaern-intel, line 1524: redundant extra newline

doc/1.4/language.md
```diff
@@ -1879,6 +1882,97 @@ be supported.
...
+  and each message component must be compatible with the corresponding message
+  component type of the hook.
+
+  `send_now` is *synchronous*: every computation suspended on the hook will
+  execute before `send_now` completes.
+  `send_now` returns the number of suspended computations that were successfully
+  resumed from the message being sent. Currently, with the only possible
+  suspended computations being method calls suspended via `after`, every
+  suspended computation is guaranteed to successfully be resumed unless
+  cancelled by a previous computation reusmed by a `send_now`. This will not
```
lwaern-intel, line 1944: > resumed by the `send_now`
A double whammy of mistakes here.

syssimics:
Verification [#10783394](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10783394): pass

syssimics:
Verification [#10786552](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10786552): pass

include/simics/dmllib.h
```diff
@@ -1221,6 +1219,454 @@ _deserialize_simple_event_data(
...
+                     const void *args, const void *msg);
+    _serializer_t args_serializer;
+    _deserializer_t args_deserializer;
+    // Size of the data args_de/serializer works with
+    uint32 args_size;
+    // id of the object parent of the target callback method
+    uint32 method_parent_id;
+} _dml_after_on_hook_info_t;
+
+// An element of an hook queue. Currently, the only elements of a hook queue
```
mandolaerik, line 1248: nitpick: *a* hook queue

include/simics/dmllib.h
```diff
@@ -1221,6 +1219,454 @@ _deserialize_simple_event_data(
     return error;
 }
 
+// All auxiliary info needed to de/serialize suspended calls created
+// from a specific after-on-hook.
+//
+// An after-on-hook is unique in respect to three things:
```
mandolaerik, line 1225: Do we define the term `after-on-hook` somewhere? We use it informally in speech, but perhaps we should find a clearer term to distinguish caller from callee, like "after hook statement"?

doc/1.4/language.md
```diff
@@ -1879,6 +1882,97 @@ be supported.
...
+// Two-dimensional hook array with two message component types
+hook (int *, bool) h3[5][3];
+```
+
+Beyond suspending computations on it, a hook <tt><em>h</em></tt> has two
+associated operations:
+
+* <pre><em>h</em>.send_now(<em>msg1</em>, ... <em>msgN</em>)</pre>
+  Sends a message through the hook, with message components
+  <tt><em>msg1</em></tt> through <tt><em>msgN</em></tt>. The number of message
```
mandolaerik, line 1933: note: markdown wizardry possible, *`msg1`*

doc/1.4/language.md
```diff
@@ -1879,6 +1882,97 @@ be supported.
...
+  `send_now` returns the number of suspended computations that were successfully
+  resumed from the message being sent. Currently, with the only possible
+  suspended computations being method calls suspended via `after`, every
+  suspended computation is guaranteed to successfully be resumed unless
+  cancelled by a preceding computation resumed by the `send_now`. This will not
+  remain true in the future: coroutines are planned to be able to reject a
+  message and reattach themselves to the hook.
+
+  <div class="note">
+  <b>Note:</b> Hooks are planned to be extended with an asynchronous variant of
```
mandolaerik, line 1949: note: can also be written
 ```
**Note:**
```

File: doc/1.4/language.md [hunk 8]
```diff
@@ -1879,6 +1882,97 @@ be supported.
...
+  <tt>send_now</tt>, to be simply called <tt>send</tt>. A call to <tt>send</tt>
+  will delay the message from being sent until the current line of execution
+  within the device is completed, and control would otherwise be returned to
+  the simulation engine. <tt>send</tt> is expected to often be preferable to
+  <tt>send_now</tt>, as it prevents subtle ordering bugs from occurring &mdash;
+  i.e. a message being sent through a hook before the computation meant to react
```
mandolaerik, line 1955: i.e.,

doc/1.4/language.md
mandolaerik: (but this is not possible for partially emphasized tt blocks)

include/simics/dmllib.h
```diff
@@ -1221,6 +1219,454 @@ _deserialize_simple_event_data(
...
+
+
+UNUSED static void
+_DML_register_hook_attribute(conf_class_t *cls, const char *attrname,
+                             get_attr_t getter, set_attr_t setter,
+                             const _dml_hook_aux_info_t *aux_info,
+                             _id_info_t hook_id_info, uint32 port_dims) {
+
+    // Conceptually, "[[sa]*]". Since the only elem representation is that of
+    // "after", instead of "a" that representation is used directly.
```
mandolaerik, line 1655: note: when expanding this, it will probably make sense to expand the type string to `[[sX]|[sY]|[sZ]*]`, where X,Y,Z are complex expressions, rather than `[[sa]]`.

include/simics/dmllib.h
```diff
@@ -2479,9 +2943,49 @@ _get_port_saved_variable(lang_void *saved_access, conf_object_t *_portobj,
                                   acc->dimension_sizes,
                                   acc->dimension_strides + portobj->ndims,
                                   acc->ndims,
+                                  _get_device_member_via_serializer,
                                   acc->getter);
 }
 
+UNUSED static uint32 *
+_DML_dimension_strides_from_dimsizes(const uint32 *dimsizes, uint32 dims,
```
mandolaerik, line 2951: why not let the caller stack-allocate the stride array? It's small unless an adversary creates a thousand dimensions of size 1, in which case implicit index args are a bigger problem.

doc/1.4/language.md
```diff
@@ -1879,6 +1882,97 @@ be supported.
...
+
+  `send_now` is *synchronous*: every computation suspended on the hook will
+  execute before `send_now` completes.
+  `send_now` returns the number of suspended computations that were successfully
+  resumed from the message being sent. Currently, with the only possible
+  suspended computations being method calls suspended via `after`, every
+  suspended computation is guaranteed to successfully be resumed unless
+  cancelled by a preceding computation resumed by the `send_now`. This will not
+  remain true in the future: coroutines are planned to be able to reject a
+  message and reattach themselves to the hook.
```
mandolaerik, line 1946: heavy text, perhaps remove `, with the only .. via after,`

py/dml/c_backend.py
```diff
@@ -1093,14 +1097,145 @@ def generate_simple_events(device):
...
+    reset_line_directive()
+
+    if dml.globals.after_on_hook_infos:
+        items = (
+            '{%s, %s, %s, %s, %s, %d}'
+            % ((string_literal(info.string_key), info.cident_callback)
+               + ((info.cident_args_serializer, info.cident_args_deserializer)
+                  if info.has_serialized_args else ('NULL', 'NULL'))
+               + (f'sizeof({info.args_type.declaration("")})'
+                  if info.args_type else '0', info.parent.uniq))
```
mandolaerik, line 1183: this looks pretty odd, in particular tuple arithmetic, but I can see why you do it (moving out conditionals to separate statements could improve formatting, but at the cost of abandoning generators). Did you consider expressing it as a sum of three f-strings would look better? (would probably be an improvement code-wise, but at the cost of disguising that the result is a 6-element initializer)

py/dml/c_backend.py
```diff
@@ -1093,14 +1097,145 @@ def generate_simple_events(device):
...
+            for info in dml.globals.after_on_hook_infos)
+        init = '{\n%s\n}' % (',\n'.join('    %s' % (item,) for item in items),)
```
mandolaerik, line 1185: `f'    {item}'`

Or consider merging with the previous statement.

py/dml/c_backend.py
```diff
@@ -1327,14 +1466,42 @@ def generate_init_identity_hashtable():
...
+                         * hook.dimensions)))
+
+        try:
+            typeseq_uniq = dml.globals.type_sequence_infos.get(
+                hook.msg_types, createnew=True).uniq
+            items.append('{%s, %d, %d}' % (offset, typeseq_uniq, hook.uniq))
+        except DMLUnkeyableType:
+            pass # already reported
+    if dml.globals.hooks:
+        init = '{\n%s\n}' % (',\n'.join('    %s' % (item,) for item in items),)
```
mandolaerik, line 1499: `f'    {item}'`

py/dml/c_backend.py
```diff
@@ -1431,48 +1598,60 @@ def generate_deinit(device):
...
         for event in events:
-            method = event.get_component('_cancel_all', 'method')
-            for index_ints in event.all_indices():
-                # Functions called from pre_delete_instance shouldn't throw any
-                # exceptions. But we don't want to force them to insert
+            by_dims.setdefault(event.dimsizes, []).append(event)
+
+        for (dims, events) in by_dims.items():
+            for i in range(len(dims)):
+                idxvar = '_i%d' % (i,)
```
mandolaerik, line 1610: `f'_i{i}'`, more cases below

py/dml/c_backend.py
```diff
@@ -2623,6 +2814,173 @@ def register_saved_attributes(initcode, node):
...
+    if node is dml.globals.device:
+        for port in node.get_components('bank', 'port', 'subdevice'):
+            register_hook_attributes(initcode, port)
+        cls = 'class'
+        getter = '_DML_get_hook_attr'
+        setter = '_DML_set_hook_attr'
+    else:
+        if dml.globals.dml_version == (1, 2) and not node.name:
+            # anonymous bank
+            return
```
mandolaerik, line 2839: ICE?

mandolaerik: oh, never mind, it's done to stop recursion.

There's a anonymous bank corner that deserves ICE,

```
dml 1.2;
import "14.dml";
bank { is foo; }
```
14.dml:
```
dml 1.4;
template foo {
  hook int x;
}
 ```
but I guess it's so obscure that we can accept it's silently broken.

However, anonymous fields are a bigger problem, maybe you can write something like
```
dml 1.2;
import "14.dml";
bank b { register r size 4 @ 0; }
```
14.dml:
```
dml 1.4;
in each field { hook int x; }
```
where I suppose the hooks will be silently broken using your patch.

So I suppose we should write `# anonymous bank or field` here, and perhaps add a 1.2-specific explicit ICE or error to reject hooks in anonymous objects.

File: doc/1.4/language.md (hunk 6 above)
Comment (lwaern-intel) [reply doc/1.4/language.md:1933]: Alright, I'll do the markdown wizadry whenever possible

File: doc/1.4/language.md (hunk 7 above)
Comment (lwaern-intel) [reply doc/1.4/language.md:1949]: I _tried_ that, but it went wrong when viewed in the Simics documentation framework for whatever reason.

File: doc/1.4/language.md (hunk 8 above)
Comment (lwaern-intel) [reply doc/1.4/language.md:1955]: umm... maybe. Gut feeling makes me prefer `&mdash;`, but I don't have any valid explanation other than "it feels like it flows better when you read it."

File: include/simics/dmllib.h (hunk 10 above)
Comment (lwaern-intel) [reply include/simics/dmllib.h:2951]: The only reason is so the user wouldn't have to allocate the array. That's not much of a reason, admittedly, since this instead forces the user to do `MM_FREE`. I'll do stack-allocation.

File: py/dml/c_backend.py [hunk 17]
```diff
@@ -2623,6 +2814,173 @@ def register_saved_attributes(initcode, node):
...
+    initcode.out('for (uint32 idx = 0; idx < %d; ++idx) {\n'
+                 % (len(hook_attr_names),) , postindent = 1)
+    initcode.out(f'_DML_register_hook_attribute({cls}, hook_attr_names[idx], '
+                 + f'{getter}, {setter}, &_hook_aux_infos[hook_ids[idx] - 1], '
+                 + f'_hook_id_infos[hook_ids[idx] - 1], {node.dimensions});\n')
+    initcode.out('}\n', preindent=-1)
+    initcode.out('}\n', preindent=-1)
+
+
+def generate_hook_attribute_funs():
```
mandolaerik, line 2870: Significant bodies of verbatim C code, maybe it's time to pay the cost (in terms of boilerplate in ctree-test, I suppose?) of moving these into dmllib.h. Can be pushed to separate PR.

I suppose you put the functions here because they reference global static arrays. One option could be optional forward declarations, like:

dmllib.h:
```
#ifndef SKIP_STATIC_ARRAYS
static ht_str_table_t _typeseq_after_on_hook_hts[];
attr_value_t
_DML_get_hook_attr(...)
...
#endif
```
and `-DSKIP_STATIC_ARRAYS` when compiling ctree-test

doc/1.4/language.md
lwaern-intel: Agreed.

py/dml/c_backend.py
lwaern-intel: The `%` syntax was actually for readability and my sanity. When you have a bunch of concatenated `f-strings`, I feel their readability easily becomes _horrible_, especially when the embedded expressions get long. It makes it too easy to lose track of what part of the string you're writing, too easy to mess up details about the string outside of the embedded expressions (for example, makes it too easy to forget writing a comma), and can make it extremely difficult to gauge what kind of string you're actually _formatting._ My stance on `%` vs f-strings has changed because of this: f-strings are superior for strings with a few number of short embedded expressions, but `%` is superior for strings with a large number of large embedded expressions.

py/dml/c_backend.py
lwaern-intel: Agreed, either suggestion is superior to what I have currently. Not sure which one I'll pick yet though.

lwaern-intel: Oh, no, I'll definitely pick `f'    {item}'`.

include/simics/dmllib.h
lwaern-intel: The eternal debate! Do we call them "after-on-hook statements" or "hook-based after statements"! And I'm still not sure which I prefer. Even if we have a long, formal name I want to have the option for a shorter informal name: "after-on-hook" allows me to e.g. say "illegal after statement on hook {hookexpr}" in error messages, which would instead become "illegal after statement based on hook {hookexpr}", which I honestly find rather terrible.

include/simics/dmllib.h
lwaern-intel: yup

py/dml/c_backend.py
lwaern-intel: 👍

py/dml/c_backend.py
lwaern-intel: 👍

include/simics/dmllib.h
mandolaerik: I think your "after statement on hook" example is a clear expression that stands on its own, it's clear what it means. What I'm afraid of about `after-on-hook` is that it feels like a label of the feature added by this PR, rather than a way to disambiguate it  from other `after` statements. We didn't introduce terms `after-on-time` and `after-on-cycles` for  `after 7 s` and `after 7 cycles`.

doc/1.4/language.md
mandolaerik: Ah right, because you are inside a HTML element, and in there everything is html with no markdown extensions.

include/simics/dmllib.h
lwaern-intel: True, but that's because we morally considered them of the same form (despite the fact they use different API call), and we only had that form until now: _after-delay_. Now that we're introducing new forms with _completely_ different semantics, we _need_ to qualify the different forms of after to separate them. Indeed, in the immediate after PR I refer to `after ... s` and `after ... cycles` as "After Delay Statements".

include/simics/dmllib.h
mandolaerik: no rather than `[[sa]*]`. If `[[s(X|Y|Z)]*]` is possible then that's even better.

py/dml/c_backend.py
mandolaerik: I think you can do both! An intermediate variable with a generator is kind of odd, compare:
```
[g(item) for item in (f(x) for x in y)]
```
```
items = (f(x) for x in y)
[g(item) for item in items]
```
```
[g(item) for x in y for f(x) in x]
```
I prefer the last one: even if the nested for syntax in generator expressions is admittedly a bit weird, it's a standard python idiom and we already depend on the reader to decode them fluently in many other places.

py/dml/c_backend.py
lwaern-intel: You're quite right, it does indeed get broken. Time to figure out the best way to catch that and ICE about it then...

lwaern-intel: We should do the same re. saved variables.

lwaern-intel: The issue re. fields is not connected to this line, though -- in this context the node we're studying has to correspond to a Simics object, i.e. it's a `device`, `subdevice`, `bank`, or `port` and of those only banks may be anonymous.

mandolaerik: ah right. This can be expressed as a trusted comment `assert isinstance(obj, Bank)`

py/dml/codegen.py
```diff
@@ -394,6 +395,462 @@ def memoization_common_prelude(name, site, outp, throws, make_ref):
...
+            self.after_on_hooks[(aoh_prim_key, param_to_msg_comp_key)] = info
+            return info
+
+class TypeSequenceInfoDict:
+    """Mappings from a unique type sequence to its TypeSequenceInfo"""
+    def __init__(self):
+        self._dict = {}
+        self._list = []
+
+    def get(self, index, realtyped=False, createnew=False):
```
mandolaerik, line 431: `create_new`

py/dml/c_backend.py
lwaern-intel: I'm gonna do this in a separate PR, or perhaps a future review round when all else is settled -- this will require some thought and testing to ensure it Doesn't Go Wrong.

py/dml/codegen.py
```diff
@@ -394,6 +395,462 @@ def memoization_common_prelude(name, site, outp, throws, make_ref):
...
+        index = tuple(index if realtyped
+                      else (safe_realtype(typ) for typ in index))
```
mandolaerik, line 433: I'd say a redundant realtype is cheap enough to be done unconditionally. Or if number of callers is small, just leave the class+realtype mapping to the caller.

If you insist not to then I'm fine with that, but in that case consider:
```
if not realtyped:
  index = ...
```

py/dml/codegen.py
```diff
@@ -394,6 +395,462 @@ def memoization_common_prelude(name, site, outp, throws, make_ref):
...
+                info = TypeSequenceInfo(index, len(self._list))
+                self._dict[classes] = [info]
+                self._list.append(info)
+                return info
+            else:
+                return None
+        for info in infos:
+            if (len(info.types) == len(index)
+                and all(typK.cmp(typI) == 0
+                        for (typK, typI) in zip(info.types, index))):
```
mandolaerik, line 448: avoid camelCase variables

py/dml/codegen.py
lwaern-intel: We need the forced tuplification anyway. I can agree with redundant realtyping if you're fine with that.

py/dml/codegen.py
```diff
@@ -394,6 +395,462 @@ def memoization_common_prelude(name, site, outp, throws, make_ref):
...
+                return info
+        if createnew:
+            info = TypeSequenceInfo(index, len(self._list))
+            self._list.append(info)
+            return info
+        else:
+            return None
+
+    def items(self):
+        return self._list
```
mandolaerik, line 458: `s/items/values` if you want to mimic the dict API. Also, why not `return (i for l in self._dict for i in l.values())`? That would allow `s/_list=[]/_num=0/`

py/dml/codegen.py
mandolaerik: Also I don't like this signature, the classname announces "dict" but it differs a lot from dict.get. Perhaps rename to `lookup` to mark the difference? Or create separate `__contains__` method for the  `create_new=False` case? Or shoehorn into what a `collections.defaultdict` use would look like?

py/dml/codegen.py
```diff
@@ -394,6 +395,462 @@ def memoization_common_prelude(name, site, outp, throws, make_ref):
...
+        except KeyError:
+            if createnew:
+                info = TypeSequenceInfo(index, len(self._list))
+                self._dict[classes] = [info]
+                self._list.append(info)
+                return info
+            else:
+                return None
+        for info in infos:
+            if (len(info.types) == len(index)
```
mandolaerik, line 446: comparison always true, so can be `assert`ed? or what am I missing?

py/dml/codegen.py
lwaern-intel: I _refuse_ to complicate this by toying around with Pythons magic methods or the classes of `collections`. They are so horribly badly documented to the point it's impossible to know whether you're doing things _right_. I'd _much much much_ rather just rename this class if you really do hate the name that much.

py/dml/codegen.py
lwaern-intel: _I do not want to mimic the dict API_. I don't even want to make the attempt, because it's impossible get it completely right.

py/dml/codegen.py
```diff
@@ -394,6 +395,462 @@ def memoization_common_prelude(name, site, outp, throws, make_ref):
...
+            infos = self._dict[classes]
+        except KeyError:
+            if createnew:
+                info = TypeSequenceInfo(index, len(self._list))
+                self._dict[classes] = [info]
+                self._list.append(info)
+                return info
+            else:
+                return None
+        for info in infos:
```
mandolaerik, line 445: Oh, a quite different solution: just make it a plain `dict`, and enrich `TypeSequenceInfo` with `__hash__` and `__eq__` instead? hash collisions are resolved by linearly eq:ing

py/dml/codegen.py
```diff
@@ -394,6 +395,462 @@ def memoization_common_prelude(name, site, outp, throws, make_ref):
...
+            info = TypeSequenceInfo(index, len(self._list))
+            self._list.append(info)
+            return info
+        else:
+            return None
+
+    def items(self):
+        return self._list
+
+class AfterInfo(metaclass=ABCMeta):
```
mandolaerik, line 460: the latest syntax is `class AfterInfo(abc.ABC):`

py/dml/codegen.py
```diff
@@ -394,6 +395,462 @@ def memoization_common_prelude(name, site, outp, throws, make_ref):
...
+            f'deserialization of arguments to {self.method.name}',
+            pre, post)
+
+    @property
+    def args_type(self):
+        return self._args_type
+
+    @property
+    def types_to_declare(self):
+        return (self._args_type,) * (self._args_type is not None)
```
mandolaerik, line 699: I geekily wonder if this can be expressed
```
@property
def types_to_declare(self):
  if self._args_type  is not None:
    yield self._args_type
```
If so, I don't have an opinion on which one is better.

py/dml/codegen.py
```diff
@@ -394,6 +395,462 @@ def memoization_common_prelude(name, site, outp, throws, make_ref):
...
+
+    @abstractproperty
+    def args_type(self): pass
+
+    @abstractproperty
+    def types_to_declare(self): pass
+
+    @property
+    def cident_callback(self):
+        return self.cident_prefix + 'callback'
```
mandolaerik, line 479: WBN with docstrings instead of `pass`

py/dml/codegen.py
```diff
@@ -394,6 +395,462 @@ def memoization_common_prelude(name, site, outp, throws, make_ref):
...
+                 param_to_msg_comp, inp, has_serialized_args):
+        self.typeseq_info = typeseq_info
+        self.parent = parent
+        # TODO drop?
+        self.inp = inp
+        self.has_serialized_args = has_serialized_args
+        self.prim_key = prim_key
+        self.param_to_msg_comp = param_to_msg_comp
+        self.uniq = len(dml.globals.after_on_hook_infos)
+        dml.globals.after_on_hook_infos.append(self)
```
mandolaerik, line 523: I tend to prefer side-effect free constructors; create a separate global function for instantiation+registration if needed, to allow separation of "real" instances from unit-test instances. Not a strong opinion, will not insist on it.

py/dml/codegen.py
mandolaerik: Right, then `items` is still not good because it mimics (an unrelated part of) the dict API.

py/dml/codegen.py
lwaern-intel: Getting a uniq is so critical to the workings of `AfterOnHookInfo` and everything interacting with it that it doesn't make sense to me to separate it.

py/dml/codegen.py
lwaern-intel: It's certainly not the same semantically. One returns a generator, the other doesn't, so e.g. with your definition `bool(obj.types_to_declare)` will always return `True`.

mandolaerik: so depends on whether you write a docstring or type annotation that specifically prescribes a tuple, or if it says an iterable is enough.

py/dml/c_backend.py
lwaern-intel: Have now sprinkled some ICEing where relevant.

py/dml/codegen.py
lwaern-intel: Now irrelevant.

py/dml/codegen.py
lwaern-intel: Now doing redundant realtyping.

py/dml/codegen.py
lwaern-intel: Now irrelevant.

py/dml/codegen.py
lwaern-intel: Now irrelevant.

py/dml/codegen.py
lwaern-intel: Done -- though with a novel `TypeSequence` class instead. `get_type_sequence_info` has now replaced `dml.globals.type_sequence_infos.get`.

py/dml/codegen.py
lwaern-intel: Refactored other usages in codegen to also do that.

py/dml/codegen.py
lwaern-intel: Funny you should say that! I did write a docstring for `types_to_declare` without reading this latest comment, and in it I indeed state "iterable"! So yeah, either works.

py/dml/codegen.py
lwaern-intel: Done. Read over it and give feedback, it's hard to explain `AfterInfo`.

py/dml/c_backend.py
```diff
@@ -1177,25 +1184,24 @@ def error_out(exc, msg):
...
+        init = '{\n%s\n}' % (',\n'.join(
+            '    {%s, %s, %s, %s, %s, %d}'
             % ((string_literal(info.string_key), info.cident_callback)
                + ((info.cident_args_serializer, info.cident_args_deserializer)
                   if info.has_serialized_args else ('NULL', 'NULL'))
                + (f'sizeof({info.args_type.declaration("")})'
                   if info.args_type else '0', info.parent.uniq))
-            for info in dml.globals.after_on_hook_infos)
-        init = '{\n%s\n}' % (',\n'.join('    %s' % (item,) for item in items),)
+            for info in dml.globals.after_on_hook_infos))
```
lwaern-intel, line 1194: forgot a comma before the last parenthesis.

py/dml/types.py
```diff
@@ -199,7 +217,7 @@ def const_str(self):
             return ""
 
     def sizeof(self):
-        '''Return size, or None if not known'''
+        '''Return size, o None if not known'''
```
lwaern-intel, line 220: i'm not sure how this happened

syssimics:
Verification [#11039173](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11039173): fail

syssimics:
Verification [#11039717](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11039717): pass

py/dml/types.py
```diff
@@ -61,6 +66,16 @@ def __init__(self, t):
...
         return "Unknown type: %s" % (self.type,)
 
+class DMLUnkeyableType(DMLTypeError):
+    def __init__(self, t, clarification=None):
+        DMLTypeError.__init__(self)
+        self.type = t
+        self.clarification = clarification
+    def __str__(self):
+        return "Unkeyable type: %s%s" % (self.type,
+                                         ('; ' + self.clarification)
```
mandolaerik, line 76: 1. punctuation: I think `;` is inaccurate here, perhaps `Unkeyable type '{type}': {clarification}`? (And no need to make clarification optional)

syssimics:
Verification [#11342775](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11342775): pass

include/simics/dmllib.h
```diff
@@ -1221,6 +1219,454 @@ _deserialize_simple_event_data(
...
+                            attr_value_t val,
+                            _dml_hook_queue_t *q) {
+    uint32 queue_len = SIM_attr_list_size(val);
+    _dml_hook_queue_t temp_q = VNULL;
+    set_error_t error = Sim_Set_Ok;
+    VRESIZE(temp_q, queue_len);
+    uint32 i = 0;
+    for (; i < queue_len; ++i) {
+        attr_value_t item = SIM_attr_list_item(val, i);
+        const char *elem_kind     = SIM_attr_string(SIM_attr_list_item(item,
```
mandolaerik, line 1606: weird space?

include/simics/dmllib.h
```diff
@@ -1221,6 +1219,454 @@ _deserialize_simple_event_data(
...
+// Serializes an after elem of a hook queue, and INCLUDES the "after" tag
+UNUSED static attr_value_t
+_DML_serialize_hook_queue_elem_after(ht_str_table_t *callback_ht,
+                                     const _id_info_t *id_infos,
+                                     _dml_hook_queue_elem_t elem) {
+    _dml_after_on_hook_info_t *callback_info =
+        ht_lookup_str(callback_ht, elem.callback_key);
+    ASSERT(callback_info);
+    _id_info_t id_info = id_infos[callback_info->method_parent_id - 1];
+    attr_value_t callback_key_attr =
```
mandolaerik, line 1567: C style says `=` after newline. (though I think it talks about binary operators, which `=` arguably is not in this case. But I think the intent of the style is clear)

py/dml/types.py
mandolaerik: ping

syssimics:
Verification [#11797949](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11797949): pass

py/dml/dmlc.py
```diff
@@ -513,6 +519,15 @@ def show_illegal_attributes(option, opt, value, parser):
...
     dml.globals.state_change_dml12 = options.state_change_dml12
 
+    dml.globals.enable_testing_features = options.enable_testing_features
+
+    if dml.globals.enable_testing_features:
+        prerr("dmlc: ***Features for internal testing are enabled!!! "
+              + "These are near-guaranteed to NOT SURVIVE in their current "
+              + "form. They are COMPLETELY UNSUPPORTED. "
+              + "The DMLC developers WILL NOT respect their use. "
+              + "NEVER enable this flag for any kind of production code!!!***")
```
mandolaerik, line 529: nitpick: we usually emit `***` in a different style, `*** tool: message`. Another options could be to make this a `WEXPERIMENTAL`.

Does not block merge.

mandolaerik (approved):

py/dml/dmlc.py
lwaern-intel: Fairly certain `*** dmlc: ` would cause our TC builds to fail.

PR #146, opened 2022-12-12
Immediate After Statement -- SIMICS-20379

lwaern-intel:
Depends on #144 . Reviewing that should take priority.

This was very quick and easy to do thanks to the architecture revamp that after-on-hooks does... but it just further highlights the copy-pasting necessary from what the architecture does not cover. 1400 insertions, ugh (though admittedly reducing the copy-pasting will just remove like ~350 loc if one is being _extremely_ optimistic, so perhaps it's not as bad as I think).

### DIFF
--- a/doc/1.4/language.md
+++ b/doc/1.4/language.md
@@ -1914,7 +1914,7 @@ computations* may be attached for execution at a later point. By sending a
 *message* through the hook, every computation suspended on the hook will become
 detached from the hook, and then executed &mdash; receiving the message as data.
 Computations suspended on a hook are executed in order of least recently
-attached.
+attached; in other words, FIFO semantics.
 
 Currently, the only computations that can be suspended and attached to hooks are
 single method calls, which is done through the use of the [`after`
@@ -1942,6 +1942,29 @@ hook(int *, bool) h3;
 Beyond suspending computations on it, a hook <tt><em>h</em></tt> has two
 associated operations:
 
+* <pre><em>h</em>.send(<em>msg1</em>, ... <em>msgN</em>)</pre>
+  Sends a message through the hook, with message components
+  <tt><em>msg1</em></tt> through <tt><em>msgN</em></tt>. The number of message
+  components must match the number of message component types of the hook,
+  and each message component must be compatible with the corresponding message
+  component type of the hook.
+
+  `send` is *asynchronous*: the message will only be sent &mdash; and suspended
+  computations executed &mdash; once all current device entries on the call
+  stack have been completed. It is exactly equivalent to
+  <tt>after: <em>h</em>.send_now(<em>msg1</em>, ... <em>msgN</em>)</tt>, except
+  it's not possible to prevent the message from being sent via `cancel_after()`.
+  For more information, see [Immediate After
+  Statements](#immediate-after-statements).
+
+  Like immediate after statements, pointers to stack-allocated data **must not**
+  be passed as message components to a `send`. If you must use pointers to
+  stack-allocated data, then `send_now` should be used instead of `send`. If you
+  want the message to be delayed to avoid ordering bugs, create a method which
+  wraps the `send_now` call together with the declarations of the local
+  variable(s) which are pointed to, and then use an immediate after statement
+  (`after: m(...)`) to delay the call to that method.
+
 * <pre><em>h</em>.send_now(<em>msg1</em>, ... <em>msgN</em>)</pre>
   Sends a message through the hook, with message components
   <tt><em>msg1</em></tt> through <tt><em>msgN</em></tt>. The number of message
@@ -1951,6 +1974,7 @@ associated operations:
 
   `send_now` is *synchronous*: every computation suspended on the hook will
   execute before `send_now` completes.
+
   `send_now` returns the number of suspended computations that were successfully
   resumed from the message being sent. Currently, every suspended computation is
   guaranteed to successfully be resumed unless cancelled by a preceding
@@ -1958,18 +1982,6 @@ associated operations:
   future: coroutines are planned to be able to reject a message and reattach
   themselves to the hook.
 
-  <div class="note">
-  <b>Note:</b> Hooks are planned to be extended with an asynchronous variant of
-  <tt>send_now</tt>, to be simply called <tt>send</tt>. A call to <tt>send</tt>
-  will delay the message from being sent until the current line of execution
-  within the device is completed, and control would otherwise be returned to
-  the simulation engine. <tt>send</tt> is expected to often be preferable to
-  <tt>send_now</tt>, as it prevents subtle ordering bugs from occurring &mdash;
-  i.e. a message being sent through a hook before the computation meant to react
-  to that message is suspended on the hook. Ordering bugs are especially
-  difficult to resolve with the planned design of coroutines, making
-  <tt>send</tt> important for their use.
-  </div>
 * <pre><em>h</em>.suspended</pre>
   Evaluates to the number of computations currently suspended on the hook.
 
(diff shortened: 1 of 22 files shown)


doc/1.4/language.md
```diff
@@ -3085,48 +3191,270 @@ return m(e1)
...
+have been completed, and the device is guaranteed to be in a consistent state
+where it is ready to handle the method call.
+
+For example, consider the following device, representing an i2c master connected
+to a slave via `i2c_link`, and receiving responses from the slave via the
+`i2c_in` port. Its function is simple: once prompted, the device will start the
+slave, and then continually read packets from it until done, handling the
+asynchronous logic through an FSM.
+```
+connect i2c_link {
```
lwaern-intel, line 3347: This example is _too long_! It also assumes that the user has knowledge about the `i2c_master_v2` and `i2c_slave_v2` interfaces. Issue is that we need to present an example of an ordering bug realistic enough that _people might recognize it_. So...

1. We need an interaction with something external. That means connect+implement
2. The device needs to have some associated state, where in one state it can handle a particular interface call, but in another it can't.
3. The state logic that prompts the external device should be separate from the implement handling its own interface calls, and it's within the former logic that the ordering bug happens...
4. ... And the ordering bug can be resolved by using immediate after _in the implement_, without having to touch the state logic.

These are _pretty_ hard to satisfy simultaneously while keeping it short. We'll have to think about this.

syssimics:
Verification [#10810260](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10810260): fail

syssimics:
Verification [#10810480](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10810480): pass

lib/1.4/dml-builtins.dml
```diff
@@ -908,20 +909,19 @@ method _get_attribute_attr(const _dml_attr_getset_info_t *_conf_info,
...
-            local attr_value_t *new_frag = SIM_attr_list(new_items[j]);
+            items_buf[j] = SIM_alloc_attr_list(new_frag_size);
+            local attr_value_t *new_frag = SIM_attr_list(items_buf[j]);
             for (local int k = 0; k < new_frag_size; ++k) {
                 new_frag[k] = items[j * new_frag_size + k];
             }
         }
-        delete items;
-        items = new_items;
+        (items, items_buf) = (items_buf, items);
```
lwaern-intel, line 919: Oh, this was just a stray optimization I realized one can do, not related to immediate after.

py/dml/c_backend.py
```diff
@@ -426,14 +442,12 @@ def generate_attr_setter(fname, node, port, dimsizes, cprefix, loopvars,
...
-        out('if (_status != Sim_Set_Ok) return _status;\n')
+        out('if (_status != Sim_Set_Ok) goto exit;\n')
         for _ in dimsizes:
             out('}\n', preindent = -1)
-        output_dml_state_change("_dev")
-        out('return Sim_Set_Ok;\n')
-    else:
-        output_dml_state_change("_dev")
-        out('return _status;\n')
+        out('exit:\n')
```
lwaern-intel, line 448: err. wrong. this should be the last `out` of the `if dimsizes` branch.

lwaern-intel: oh wait i removed the `else`, pfft, nevermind

py/dml/c_backend.py
```diff
@@ -1648,8 +1693,18 @@ def generate_deinit(device):
...
-            codegen_inline_byname(device, (), 'destroy', [], [],
-                                  device.site).toc()
+            code = codegen_inline_byname(device, (), 'destroy', [], [],
+                                         device.site)
+        if not code.is_empty:
+            # TODO to consider: move the device entry to before _cancel_all
+            # of events are called? This would mean any immediate afters of
+            # events' destroy calls will be delayed until after destroy on the
+            # device is called, instead of being executed once the event's
+            # destroy would return.
```
lwaern-intel, line 1703: oh wait the `_cancel_all` inline call doesn't generate device entries and exits. I figured it would use the wrapped `__destroy` event methods, but it doesn't. That's a bug.

py/dml/codegen.py
```diff
@@ -2805,6 +2889,79 @@ def stmt_afteronhook(stmt, location, scope):
     return [mkAfterOnHook(site, domains, hookref_expr, aoh_info, indices,
                           args_init)]
 
+class ImmediateAfterContext:
```
lwaern-intel, line 2892: Artifact to be removed.

lwaern-intel:
Making this a draft pending discussion regarding what we want immediate after to be and how to resolve issues with `SIM_transaction_wait`.

doc/1.4/language.md
lwaern-intel: I still am not fond of this example, and I don't know how to best address it. For the moment I've mitigated this issue by presenting it as the final part of this subsubsubsection, and describing it as a scenario rather than a simple example.

py/dml/c_backend.py
lwaern-intel: N/A anymore with the `VT_stacked_post` based implementation

doc/1.4/language.md
```diff
@@ -3300,29 +3355,156 @@ after h -> (x, y): my_callback(i, y + 1.5, false)
...
+```
+The FSM transitions to the state capable of handling the slave's acknowledgement
+only after the `start` call returns, so if the slave responds to the start
+synchronously &mdash; as part of the `start` call &mdash; `on_acknowledge`
+will be called while the device still considers itself to be in its unstarted
+state.
+
+This bug can be solved in numerous ways &mdash; the most obvious is to
+transition the state before making the `start` call &mdash; but immediate after
+provides a solution which doesn't require carefully managing the FSM:s logic, by
```
lwaern-intel, line 3489: Oops, small grammar issue: `FSM's`, not `FSM:s`.

doc/1.4/language.md
```diff
@@ -3300,29 +3355,156 @@ after h -> (x, y): my_callback(i, y + 1.5, false)
...
+after statements are executed across multiple device instances is not defined.
+
+Within an immediate after statement, every argument provided to the called
+method is evaluated at the time the `after` statement is executed, and stored so
+that they may be used when the method call is to be performed. Unlike the other
+forms of `after` statements, the input parameters of the method are never
+required to be of serializable type, meaning pointers can be passed as arguments
+to the callback. But **beware**: pointers to stack-allocated data (pointers to
+or within `local` variables) must **never** be passed as arguments. The
+stack with which the `after` statement is executed is _not_ preserved, so any
```
lwaern-intel, line 3399: `s/_not_/*not*`

doc/1.4/language.md
```diff
@@ -3300,29 +3355,156 @@ after h -> (x, y): my_callback(i, y + 1.5, false)
...
-a message component.
+#### Immediate After Statements
+<pre>
+after: <em>method</em>(<em>e1</em>, ... <em>eN</em>);
+</pre>
 
-This means that `after` statements cannot be used with methods that e.g. have
-pointer input parameters, unless the arguments for those input parameters are
-message component parameters of the `after`.
+In this form, the specified point in the future when control is given back to
```
lwaern-intel, line 3363: Grammatical artifact from too many revisions, should be "the specified point in the future is when"

doc/1.4/language.md
```diff
@@ -3300,29 +3355,156 @@ after h -> (x, y): my_callback(i, y + 1.5, false)
...
+the simulation engine such that the ongoing simulation of the current processor
+may progress, and the simulation would otherwise be ready to move onto the next
+cycle.
```
lwaern-intel, line 3366: Can be shortened to "may progress, and would otherwise be ready to move onto the next cycle."

doc/1.4/language.md
```diff
@@ -3300,29 +3355,156 @@ after h -> (x, y): my_callback(i, y + 1.5, false)
...
+
+Semantically, the immediate after statement is very close to
+`after 0 cycles: ...`, but has a number of advantages. In general, the immediate
+after statement is designed to execute the callback as promptly as possible
+while satisfying the above conditions, while `after 0 cycles: ...` is not.
+In particular, in Simics, callbacks delayed via `after 0 cycles` are always
+bound to the clock associated with the device instance, which is not always
+that of the processor currently under simulation &mdash; in such cases the
+simulated processor may progress indefinitely without the posted callback being
+executed. In addition, if an immediate after statement is executed while the
```
lwaern-intel, line 3382: Clarification: "... without the posted callback being executed. The immediate after statement does not have this issue."

doc/1.4/language.md
mandolaerik: Perhaps some wire fiddling over the signal interface? Raise an outgoing signal A toward a power management device, and wait for incoming signal B to be raised as an ack that it's OK to proceed with some operation (and lower A again). Similar patterns are pretty common in real HW.

doc/1.4/language.md
```diff
@@ -1942,6 +1942,29 @@ hook(int *, bool) h3;
...
+  it's not possible to prevent the message from being sent via `cancel_after()`.
+  For more information, see [Immediate After
+  Statements](#immediate-after-statements).
+
+  Like immediate after statements, pointers to stack-allocated data **must not**
+  be passed as message components to a `send`. If you must use pointers to
+  stack-allocated data, then `send_now` should be used instead of `send`. If you
+  want the message to be delayed to avoid ordering bugs, create a method which
+  wraps the `send_now` call together with the declarations of the local
+  variable(s) which are pointed to, and then use immediate after (`after:
```
mandolaerik, line 1965: "an immediate after statement", grammatically problematic to use "after" as a noun. This is the only end-user visible occurrence I could spot, I won't oppose the identifiers.

doc/1.4/language.md
```diff
@@ -3396,114 +3396,79 @@ forms of `after` statements, the input parameters of the method are never
...
+param STATE_EXPECTING_ACK = 1;
+saved int curr_state = STATE_IDLE;
+
+connect manager {
+    interface signal;
+}
+
+implement signal {
+    method signal_raise() {
+        on_acknowledgement();
```
mandolaerik, line 3426: consider grouping these two in a `port`. Implementing interfaces on top level is generally discouraged.

doc/1.4/language.md
lwaern-intel: Example redone, and you gave it a thumbs up, so I'm keeping it (modulo the use of a `port`).

doc/1.4/language.md
lwaern-intel: Done.

mandolaerik (approved):

PR #148, opened 2022-12-24
Enhance hard reset signal implementation

AleksandrDanilovIntel:


lib/1.4/utility.dml
```diff
@@ -330,6 +330,7 @@ template hard_reset {
         foreach obj in (this._each_hard_reset)
             obj.hard_reset();
     }
+    shared method hard_reset_deassert() default {}
```
mandolaerik, line 333: This is redundant in a sense; you can override signal_lower and call `default` instead.

(One could argue that it would be better with a simulator-agnostic method to override, but that sort of functionality would rather belong to the `signal_port` template.)

lib/1.4/utility.dml
```diff
@@ -342,11 +343,15 @@ template _init_val_hard_reset is (init_val, hard_reset) {
...
+        saved bool reset_asserted;
         implement signal {
             method signal_raise() default {
+                reset_asserted = true;
                 dev.hard_reset();
             }
             method signal_lower() default {
+                dev.hard_reset_deassert();
+                reset_asserted = false;
             }
```
mandolaerik, line 355: It would make more sense to start with `template hreset is (hard_reset, signal_port)` and continue from there. IIRC, @svenwest tried this when adding `signal_port`, but it was dropped since it caused too much short-term SSM breakage. I think it can be done, but requires a careful migration plan.

mandolaerik (changes requested):

lib/1.4/utility.dml
AleksandrDanilovIntel: Both templates are define default methods. Using spec violation is also doubtful.

lib/1.4/utility.dml
AleksandrDanilovIntel: Would be better to hide the implementation and expose only the high level methods as hard_reset()
is hreset;
port HRESET {
    saved bool reset_asserted;
    implement signal {
        method signal_lower() {
            hard_reset_deassert();
            default();
        }

svenwest: I'm with Erik on this. What's the added value of overriding `hard_reset_deassert` instead of overriding `signal_lower`?

If we add a `hard_reset_deassert` method like this we'd need to consider some more questions:
For example
* Should `hard_reset_deassert` be called after a `dev.hard_reset()`?
* Should `reset_asserted` be true during a call to `dev.hard_reset()`?
* Why is there no `hard_reset_assert()` method?

If we instead let the user override `signal_lower` in the HRESET port it's very clear what behavior is being overridden, and what can be expected.

AleksandrDanilovIntel: agreed

lib/1.4/utility.dml
AleksandrDanilovIntel: Is it possible to call default() from the default method?
            method signal_lower() default {
                default();
            }

AleksandrDanilovIntel: I would prefer to copy a logic from signal port template
`template hreset is hard_reset {
    port HRESET {
        saved bool high;
        implement signal {
            method signal_raise() default {
                high = true;
                dev.hard_reset();
            }
            method signal_lower() default {
                high = false;
            }`

PR #149, opened 2023-01-02
Update copyright year

mandolaerik:


mandolaerik:
Copyright recommendations changed recently so perhaps it's the test that should be changed. This is a quick fix to unblock verification builds, we may want to revert it later

syssimics:
Verification [#10906107](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10906107): fail

PR #151, opened 2023-01-05
Don't declare a temporary struct instance as a local variable

mandolaerik:
- remove obsolete XFAIL
- Add test that ambiguous inheritance can be resolved by override
- Avoid having to pass -L when re-running test manually
- Don't declare a temporary struct instance as a local variable

### DIFF
--- a/py/dml/serialize.py
+++ b/py/dml/serialize.py
@@ -450,6 +450,7 @@ def generate_deserialize(real_type):
 
     func_code = output.StrOutput()
     with func_code:
+        cleanup = []
         output.out(function_decl + " {\n", postindent = 1)
         out_arg_decl.toc()
         output.out("set_error_t _success UNUSED = Sim_Set_Ok;\n")
@@ -462,14 +463,17 @@ def generate_deserialize(real_type):
             stmts.append(ctree.mkInline(site, 'goto _exit;'))
             return stmts
         if isinstance(real_type, TStruct):
+            (tmp_out_decl, tmp_out_ref) = declare_variable(
+                site, "_tmp_out", TPtr(real_type),
+                ctree.mkNew(site, real_type))
+            tmp_out_decl.toc()
+            cleanup.append(ctree.mkDelete(site, tmp_out_ref))
             statements = []
             imm_attr_decl, imm_attr_ref = declare_variable(
                 site, "_imm_attr", attr_value_t)
-            tmp_out_decl, tmp_out_ref = declare_variable(
-                site, "_tmp_out", real_type)
             for (i, (name, val_type)) in enumerate(real_type.members.items()):
                 index = ctree.mkIntegerConstant(site, i, False)
-                val_ref = ctree.mkSubRef(site, tmp_out_ref, name, ".")
+                val_ref = ctree.mkSubRef(site, tmp_out_ref, name, "->")
                 sim_attr_list_item = apply_c_fun(
                     site, "SIM_attr_list_item",
                     [ctree.mkDereference(site, in_arg), index],
@@ -483,10 +487,11 @@ def generate_deserialize(real_type):
                 statements += [imm_set, sub_deserialize]
             tmp_set = ctree.mkAssignStatement(
                 site, ctree.mkDereference(site, out_arg),
-                ctree.ExpressionInitializer(tmp_out_ref))
+                ctree.ExpressionInitializer(
+                    ctree.mkDereference(site, tmp_out_ref)))
             deserialization = ctree.mkCompound(
                 site,
-                [imm_attr_decl, tmp_out_decl] + statements + [tmp_set])
+                [imm_attr_decl] + statements + [tmp_set])
             is_struct_expr = expr.mkLit(
                 site,
                 'SIM_attr_is_list(*in) && '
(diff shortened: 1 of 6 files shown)


syssimics:
Verification [#10921055](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10921055): fail

syssimics:
Verification [#10923336](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10923336): pass

test/1.4/serialize/T_large_struct.dml
```diff
@@ -0,0 +1,26 @@
...
+// This used to happen in serialization code, HSD-ES 18026246959
+
+constant size = 20_000_000;
+
+typedef struct {
+    uint8 data[size];
+} huge_t;
+saved huge_t x;
+
+method test() throws {
```
JonatanWaern, line 22: I don't like testing this from within the device file, even if its simpler, since this test will now incidentally break if we change the internal naming or attribute typing of saved variables. It's better to do an actual checkpoint save/restore from python

py/dml/serialize.py
```diff
@@ -462,14 +463,17 @@ def error_out(exc, msg):
...
+            cleanup.append(ctree.mkDelete(site, tmp_out_ref))
             statements = []
             imm_attr_decl, imm_attr_ref = declare_variable(
                 site, "_imm_attr", attr_value_t)
-            tmp_out_decl, tmp_out_ref = declare_variable(
-                site, "_tmp_out", real_type)
             for (i, (name, val_type)) in enumerate(real_type.members.items()):
                 index = ctree.mkIntegerConstant(site, i, False)
-                val_ref = ctree.mkSubRef(site, tmp_out_ref, name, ".")
+                val_ref = ctree.mkSubRef(site, tmp_out_ref, name, "->")
```
JonatanWaern, line 476: As a complete sidenote;
Doesnt mkSubRef's argument know about it's typing? Why do we have to specify which access operator to use?
And sidenote sidenote;
If we _do_ have to specify the operator, shouldn't this be two separate functions (since we only have two valid operators)?

JonatanWaern (changes requested):
The commit stack seems to be a little wider than the pull request name (which is a whatever)

test/1.4/serialize/T_large_struct.dml
mandolaerik: Creating an actual checkpoint comes with a 10x performance penalty, though. Probably pays off more to multiply the size by 10; currently we break the stack by 2x margin so if stack size is doubled someday then the current test will silently stop to cover this bug. 20x margin feels more safe.

I'm not worried about the representation issue. If we change the representation of saved variables, then updating our own tests will be the least of our problems.

test/1.4/serialize/T_large_struct.dml
```diff
@@ -0,0 +1,26 @@
...
+    local uint8 *mem = new uint8[size];
+    local attr_value_t a = SIM_make_attr_list(1, SIM_make_attr_data(size, mem));
+    SIM_set_attribute(obj, "x", &a);
```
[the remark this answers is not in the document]
mandolaerik: Good point

test/1.4/serialize/T_large_struct.dml
JonatanWaern: What is the 10x penalty (and is it actually significant for this test, even in this extreme case)?

I am kinda holding onto hope that people are not relying on the internal rep of saved variables atm, but probably that is not true.

mandolaerik: Actually the 10x factor was from saving only; load is apparently slower so the factor is somewhere 50-100x. 

The number 200 million is tuned to match a time budget of 0.1s; a test case that saves/restores a checkpoint that just barely would evade a stack overflow before the fix needs 0.5 s.

I don't understand why we shouldn't allow our tests of saved variables to know the attribute representation.

In any case, perhaps an even better approach is to do some linux-only hacks to make sure the stack is almost full before accessing a medium-sized attribute. E.g., `setrlimit(RLIMIT_STACK, 100000)` or `alloca(getrlimit(RLIMIT_STACK) - 100000)` before playing with a 200k attribute.

mandolaerik:
yes, I sneaked in some stray commits that I had lying around

test/1.4/serialize/T_large_struct.dml
mandolaerik: I can agree that it's a mild form of whitebox testing, and pure blackbox is better in some ways, but not so relevant given that writing resource usage tests is tough.

After the rlimit trickery the cost is no longer a problem, but it needs to be called from within the %header function so would need to save/restore using `SIM_run_command("read-configuration foo -prefix bar");` and it's perhaps not worth the ugliness.

syssimics:
Verification [#10935721](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10935721): pass

test/1.4/serialize/T_large_struct.dml
JonatanWaern: agreed

PR #152, opened 2023-01-11
Distribute dmlc as .py instead of .pyc

mandolaerik:
This greatly simplifies pypy usage


syssimics:
Verification [#10946431](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10946431): fail

syssimics:
Verification [#10949339](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10949339): pass

RELEASENOTES.docu
```diff
@@ -1709,4 +1709,7 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
   <build-id value="6176"><add-note> <tt>default</tt> can now be called within
       <tt>destroy()</tt> without causing an internal compiler error
       <bug number="SIMICS-19598"/>.</add-note></build-id>
+  <build-id value="next"><add-note> The DMLC implementation is now
+      shipped as source code, as <tt>.py</tt> files instead
+      of <tt>.pyc</tt>.</add-note></build-id>
```
lwaern-intel, line 1714:
```
The DMLC implementation provided as part of Simics Base is now shipped as source code; i.e. as <tt>.py</tt> files instead of <tt>.pyc</tt>.</add-note></build-id>
```
Or something like that? I think we should clarify that we mean DMLC as part of Simics Base.

mandolaerik: I thought so too at first, but then I realized that the release note is only exposed as part of the "Release notes for Simics Base" document, so the user should understand this context, right? the note should only be confusing to readers of release-note source code.

lwaern-intel (approved):
The question re. the release-note isn't a dealbreaker for me.

PR #154, opened 2023-01-17
Initializer syntax for arguments -- SIMICS-20473

lwaern-intel:
I wanted to implement this before starting implementation of RAII proper.

### DIFF
--- a/py/dml/codegen.py
+++ b/py/dml/codegen.py
@@ -1445,55 +1445,50 @@ def stmt_local(stmt, location, scope):
 
     decls = list(map(convert_decl, decls))
 
-    def mk_sym(name, typ):
-        sym = LocalSymbol(name, scope.unique_cname(name), type=typ,
-                          site=stmt.site, stmt=True)
-        stmts.append(sym_declaration(sym))
-        return sym
-
-    def add_var(name, typ, init = None):
-        sym = scope.add_variable(
-            name, type = typ, site = stmt.site, init = init, stmt = True,
-            make_unique=not dml.globals.debuggable)
-        stmts.append(sym_declaration(sym))
-        return sym
+    def mk_sym(name, typ, mkunique=not dml.globals.debuggable):
+        cname = scope.unique_cname(name) if mkunique else name
+        return LocalSymbol(name, cname, type=typ, site=stmt.site, stmt=True)
 
     if method_call_init:
-        late_sym_adds = []
-        late_declares = []
-        tgts = []
+        syms_to_add = []
+        tgt_syms = []
+        late_declared_syms = []
 
         for (name, typ) in decls:
+            sym = mk_sym(name, typ)
             tgt_typ = safe_realtype_shallow(typ)
             if tgt_typ.const:
                 nonconst_typ = tgt_typ.clone()
                 nonconst_typ.const = False
-                sym = mk_sym('_tmp_' + name, nonconst_typ)
-                late_declares.append((name, typ, sym))
+                tgt_sym = mk_sym('_tmp_' + name, nonconst_typ, True)
+                sym.init = ExpressionInitializer(mkLocalVariable(stmt.site,
+                                                                 tgt_sym))
+                late_declared_syms.append(sym)
             else:
-                sym = mk_sym(name, typ)
-                late_sym_adds.append(sym)
-            tgts.append(mkLocalVariable(stmt.site, sym))
+                tgt_sym = sym
+            syms_to_add.append(sym)
+            tgt_syms.append(tgt_sym)
 
+        tgts = [mkLocalVariable(stmt.site, sym) for sym in tgt_syms]
         method_invocation = try_codegen_invocation(stmt.site,
                                                    inits,
                                                    tgts, location, scope)
         if method_invocation is not None and stmt.site.dml_version != (1, 2):
+            for sym in syms_to_add:
+                scope.add(sym)
+            stmts.extend(sym_declaration(sym, True) for sym in tgt_syms)
             stmts.append(method_invocation)
+            stmts.extend(sym_declaration(sym, True)
+                         for sym in late_declared_syms)
         else:
             if len(tgts) != 1:
-                report(ERETLVALS(site, 1, len(tgts)))
+                report(ERETLVALS(stmt.site, 1, len(tgts)))
             else:
-                tgt = tgts[0]
-                init = ExpressionInitializer(
+                sym = syms_to_add[0]
+                sym.init = ExpressionInitializer(
                     codegen_expression(inits[0].args[0], location, scope))
-                stmts.append(mkAssignStatement(tgt.site, tgt, init))
-
-        for (name, typ, temp_sym) in late_declares:
-            init = ExpressionInitializer(mkLocalVariable(stmt.site, temp_sym))
-            add_var(name, typ, init)
-        for sym in late_sym_adds:
-            scope.add(sym)
+                scope.add(sym)
+                stmts.append(sym_declaration(sym, True))
     else:
         # Initializer evaluation and variable declarations are done in separate
         # passes in order to prevent the newly declared variables from being in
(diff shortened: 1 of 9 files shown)


syssimics:
Verification [#10974104](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10974104): fail

py/dml/expr.py
```diff
@@ -228,6 +229,74 @@ def typecheck_inargs(site, args, inp, kind="function", known_arglen=None):
...
+                                             scope, False).as_expr(ptype))
+            except EASTYPE as e:
+                if e.site is init.site:
+                    raise EPTYPE(site, e.source, e.target_type, pname,
+                                 kind) from e
+                raise
+            # better error message
+            except EDISCONST as e:
+                if e.site is init.site:
+                    raise ECONSTP(site, pname, kind + "call") from e
```
lwaern-intel, line 269: missing a space, should be `" call"`.

py/dml/codegen.py
```diff
@@ -1974,9 +1968,8 @@ def stmt_return(stmt, location, scope):
                     report(ERETARGS(stmt.site, len(outp_typs), len(outargs)))
                     # avoid control flow errors by falling back to statement
                     # with no fall-through
-                    return mkAssert(stmt.site,
-                                    mkBoolConstant(stmt.site, False))
-
+                    return [mkAssert(stmt.site,
+                                    mkBoolConstant(stmt.site, False))]
```
lwaern-intel, line 1972: bad indentation

syssimics:
Verification [#10976964](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10976964): pass

py/dml/codegen.py
```diff
@@ -1445,63 +1444,61 @@ def convert_decl(decl_ast):
...
-                late_declares.append((name, typ, sym))
+                tgt_sym = mk_sym('_tmp_' + name, nonconst_typ, True)
+                sym.init = ExpressionInitializer(mkLocalVariable(stmt.site,
+                                                                 tgt_sym))
+                late_declared_syms.append(sym)
             else:
-                sym = mk_sym(name, typ)
-                late_sym_adds.append(sym)
-            tgts.append(mkLocalVariable(stmt.site, sym))
+                tgt_sym = sym
```
mandolaerik, line 1467: I suppose the `const` adventures come from `local const int x = m()` where `m` throws?

Would be nice with a short reminder (just a comment with an example is enough) to remind the reader of this use case. (not really new for this PR so not blocking)

py/dml/expr.py
```diff
@@ -243,6 +312,31 @@ def read(self):
...
+        raise EAPPLY(fun)
+
+    args = typecheck_inarg_inits(
+        site, inits,
+        [(str(i + 1), t) for (i, t) in enumerate(funtype.input_types)],
+        location, scope, 'function', funtype.varargs)
+
+    return Apply(site, fun, args, funtype)
+
 def mkApply(site, fun, args):
```
mandolaerik, line 340: IIRC, mkApply belongs to expr.py only because of `Expression.apply`; perhaps it can be demoted to ctree now?

py/dml/expr.py
```diff
@@ -15,10 +15,11 @@
     'Expression',
     'NonValue',
     'mkLit', 'Lit',
-    'mkApply', 'Apply',
+    'mkApply', 'mkApplyInits', 'Apply',
```
mandolaerik, line 18: Naming: did you consider flipping the naming like `mkApply` for initializers vs `mkApplyExprArgs` or `mkApplyOld` for expressions, making initializers the new normal?

py/dml/expr.py
```diff
@@ -228,6 +229,74 @@ def typecheck_inargs(site, args, inp, kind="function", known_arglen=None):
...
 
+def typecheck_inarg_inits(site, inits, inp, location, scope,
+                          kind="function", variadic=False,
+                          allow_undefined_args=False):
+    if (not variadic and len(inits) != len(inp)) or len(inits) < len(inp):
+        raise EARG(site, kind)
+
+    from .expr_util import coerce_if_eint
+    from .codegen import eval_initializer, codegen_expression, \
+                         codegen_expression_maybe_nonvalue
```
mandolaerik, line 240: should some of these be promoted to expr.py? the original purpose of the expr.py / ctree.py split was to avoid some cyclic imports, so good to keep down the number of cycles. But only if it doesn't come with a big cost.

py/dml/expr.py
```diff
@@ -228,6 +229,74 @@ def typecheck_inargs(site, args, inp, kind="function", known_arglen=None):
...
+                              + 'a simple expression')
+            arg = codegen_expression_maybe_nonvalue(init.args[0], location,
+                                                    scope)
+            if (isinstance(arg, NonValue)
+                and not (allow_undefined_args and arg.undefined)):
+                raise arg.exc()
+            args.append(arg)
+        elif (not dml.globals.compat_dml12_int(site)
+            or not dml.globals.compat_dml12_int(init.site)
+            or init.kind != 'initializer_scalar'):
```
mandolaerik, line 257: perhaps `X and Y and Z==W` is easier to read than `not X or not Y or Z!=W` (if you flip true/false branch)

mandolaerik (approved):
some minor comments, none of which blocks a merge

py/dml/expr.py
lwaern-intel: this is cyclical in codegen.py and expr.py, which is much harder to resolve.

py/dml/codegen.py
lwaern-intel: Yeah, and also to deal with multiple return values
```C
local (const int x, const int y) = m();
```

But yeah, that's old logic. What this change is (mostly) about is so that:
```C
method m() -> (int) { ... }
...
local int x = m();
```
Compiles to:
```C
int x = m();
```
rather than
```C
int x;
x = m();
```

py/dml/expr.py
lwaern-intel: True, it's only in expr.py due to `Expression.apply` -- but how do you figure it could be demoted? By adding
```C
from .ctree import mkApply`
```
in `Expression.apply`?

py/dml/expr.py
lwaern-intel: I did a little, but not seriously. My thought is not only is `mkApply` still used for 1.2, it is also used _a lot_ internally, while `mkApplyInits` isn't. Plus, `mkApply` is simpler, and in a sense more primitive. It makes more sense to me for it to be the unqualified variant.

py/dml/expr.py
lwaern-intel: Agreed.

py/dml/expr.py
mandolaerik: That was my gut feeling too, until I actually grepped for it. 
The 1.2 use (int_register) is obscure and dying, so those don't count. Other than that, there's ~15 uses, of which you turn a handful into `mkApplyInits`, and more or less everything else is for calling stuff in dmllib.h. I always found the `mkApply(site, mkLit(site, 'blah', TFunction([...], TInt), [...])` syntax a bit clumsy, for just calling a dmllib function. (Also, the inarg typechecking added by the TFunction is usually redundant; it's usually guaranteed to be correct so entrusing gcc with that typecheck is more than acceptable.)

Maybe the longer-term path is to add a separate primitive for "call dmllib function"?

py/dml/codegen.py
mandolaerik: what's the point of that change? `int x = m()` looks better to a human, but I don't think gcc cares. Unless the variable is `const`, of course.

lwaern-intel: Before the patch where I allowed method calls as initializers to local variable declarations, then it was genned as:
```
int x = m();
```
so the current behavior can be seen as a regression, which has always bugged me, in particular because it makes dealing with constness more awkward. So this tweak to the logic is to address that regression. It's just a straight-up improvement.

py/dml/expr.py
lwaern-intel: Hmm... you are  right. I'd be convinced, but I'm torn on one last point: `mkApply`'s original role is the constructor for `Apply` which corresponds to a C application which works exclusively with expressions and not initializers. It could _maybe possibly_ be argued that having `mkApplyInits` be renamed to `mkApply` would bastardize the role of constructors, and, in fact, having the `mk-` prefix of `mkApplyInits` is already a mistake and it should be named something else like `codegen_apply`.

PR #155, opened 2023-01-21
Make DMLC output deterministic

mandolaerik:
- Adapt to rename of state-change notifier
- Enable line directives by default
- Allow disabling gcc optimizations
- Use dicts and lists instead of sets to avoid indeterministic output
- Sort sets before using, to avoid indeterministic output
- Validate that two runs with different pythons give the same C output
- fix minor typos
- Use ordinary dictionaries

### DIFF
--- a/py/dml/set_test.py
+++ b/py/dml/set_test.py
@@ -1,0 +1,99 @@
+# © 2023 Intel Corporation
+# SPDX-License-Identifier: MPL-2.0
+
+from dml.set import Set
+
+import unittest
+
+class TestSet(unittest.TestCase):
+    def test_contains(self):
+        (o1, o2) = (object(), object())
+        x = Set([0, 2, 'foo', o1])
+        self.assertTrue(0 in x)
+        self.assertFalse(1 in x)
+        self.assertTrue('foo' in x)
+        self.assertFalse('bar' in x)
+        self.assertTrue(o1 in x)
+        self.assertFalse(o2 in x)
+
+    def test_eq(self):
+        x = Set(range(10))
+        self.assertEqual(x, Set(range(10)))
+        self.assertNotEqual(x, Set(range(11)))
+        self.assertEqual(x, set(range(10)))
+        assert set(range(10)) == {e: e for e in range(10)}.keys()
+        self.assertEqual(x, {e: e for e in range(10)}.keys())
+
+    def test_str_repr(self):
+        for f in [str, repr]:
+            self.assertEqual(f(Set()), 'Set()')
+            self.assertEqual(f(Set([1, 'one'])), "Set([1, 'one'])")
+
+    def test_iter(self):
+        els = [object() for _ in range(100)]
+        x = Set(x for x in els)
+        self.assertEqual(list(x), els)
+
+    def test_len(self):
+        self.assertEqual(len(Set()), 0)
+        self.assertEqual(len(Set(range(10))), 10)
+        # the bool function relies on __len__
+        self.assertFalse(bool(Set()))
+        self.assertTrue(bool(Set([1])))
+
+    def test_add(self):
+        x = Set()
+        x.add(4)
+        x.add(10)
+        x.add(8)
+        self.assertTrue(4 in x)
+        self.assertTrue(8 in x)
+
+    def test_update(self):
+        x = Set([10, 4, 2])
+        x.update(x for x in [4, 8, 10])
+        self.assertEqual(list(x), [10, 4, 2, 8])
+
+    def test_remove(self):
+        x = Set(range(10))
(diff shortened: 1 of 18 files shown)


syssimics:
Verification [#10994099](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10994099): fail

mandolaerik:
All cases of inteterminisms were caused by set iteration order being undefined. This did not only cause differences between pypy and mini-python; it also caused differences between mini-python based runs.

Maintaining determinism adds a maintenance burden: if you use sets carelessly, this will not be captured until 226, which normally happens after merge, in Integration builds. We also need to define some guidelines for when sets can be used. The PR's current convention is roughly to avoid using collections with undefined iteration order *across functions*. This means:
* a set can be used within a function, but if the function constructs some list or dictionary based on it, then the set's elements must be sorted first
* If sharing data across functions which naturally would have a set datatype (no duplicate elements, order insignificant), then two options:
  * use list if possible (i.e. if fast `in` is not needed, and the collection is final)
  * otherwise, use a dict with None values, e.g. 
  * If nobody iterates the collection, then it's also possible to represent as a set, but this is brittle and should be a avoided
* Currently I made an exception for global reference counts, which are typically iterated once only. I found it more readable to represent this as a set, and sort it on iteration.
* Be generous about applying redundant sorting (not just deterministic order based on order of appearance in input) when this is cheap and potentially improves debugging experience, e.g. a centrally used collection like Trait.ancestors or when outputting a list of declarations/calls/etc. in C code, like vtable initialization calls.

These conventions are up for discussion, and the conventions changed over time so I haven't followed them strictly in this PR.

syssimics:
Verification [#10994184](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10994184): fail

syssimics:
Verification [#10995016](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=10995016): pass

py/dml/traits.py
```diff
@@ -444,7 +445,7 @@ def sort_method_implementations(implementations):
                             impl.name)
         rank_to_method[impl.rank] = impl
 
-    ranks = set(rank_to_method)
+    ranks = {r: None for r in rank_to_method}
```
JonatanWaern, line 448: For other reading this:
As of python 3.7 dictionaries retain insertion order for keys _as part of the official spec_, so this is deterministic if rank_to_method order is deterministic

JonatanWaern (approved):

mandolaerik:
Some commits in this series are related to https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/3102

We could in principle dissociate test running parts of this PR into a separate  commit, but deterministic output is closely related to what t226 does so felt natural to combine them

mandolaerik:
[V 3](https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=11015149)

awmoore-intel:
Will this result in output .c/.h files to be exactly the same, allowing users to use ccache in direct mode? https://ccache.dev/manual/latest.html#_how_ccache_works

syssimics:
Verification [#11015149](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11015149): fail

mandolaerik:
> Will this result in output .c/.h files to be exactly the same, allowing users to use ccache in direct mode? https://ccache.dev/manual/latest.html#_how_ccache_works

Yes

py/dml/set_test.py
```diff
@@ -0,0 +1,99 @@
...
+        self.assertFalse('bar' in x)
+        self.assertTrue(o1 in x)
+        self.assertFalse(o2 in x)
+
+    def test_eq(self):
+        x = Set(range(10))
+        self.assertEqual(x, Set(range(10)))
+        self.assertNotEqual(x, Set(range(11)))
+        self.assertEqual(x, set(range(10)))
+        assert set(range(10)) == {e: e for e in range(10)}.keys()
```
JonatanWaern, line 24: this plain assert sticks out, why is this like this?

JonatanWaern (approved):
Other than a few lingering questions regarding when to use lists and when to use Sets, this seems fine to me

py/dml/set_test.py
mandolaerik: The class under test is `Set`, and this line doesn't test `Set` but is rather an executable comment that confirms that ordinary sets work the way that the previous tests expect Set to work (to me it's far from obvious that set and dict.keys() are expected to compare equal)

syssimics:
Verification [#11020120](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11020120): fail

PR #157, opened 2023-01-27
Remove the expect function from the 1.4 test framework

mandolaerik:
'assert' better serves its purpose


JonatanWaern (commented):
I am not sure how assert is superior here, is there off-github discussion I've missed?
The ability of assert to both encourage documentation of _what_ exactly is being tested, as well as the ability to give different errors for, for example, different failing indexes in an array seem beneficial to me.

syssimics:
Verification [#11024254](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11024254): pass

test/1.4/lib/T_unmapped_access.dml
```diff
@@ -75,64 +75,47 @@ method test() throws {
...
+    assert b.r2.read_register(0xfffffffff, NULL) == 0x0f0ff000;
 
     // Check overridden semantics
     b.r3.write_register(0xfffff0ff, 0xfffffff0, cast(11, void*));
-    expect(most_recent_val == 0xf0ff, "simple custom write_unmapped val");
-    expect(most_recent_mask == 0xfff0, "simple custom write_unmapped mask");
-    expect(most_recent_aux == cast(11, void*), "simple custom write_unmapped aux");
+    assert most_recent_val == 0xf0ff;
+    assert most_recent_mask == 0xfff0;
+    assert most_recent_aux == cast(11, void*);
```
lwaern-intel, line 91: oh right this is UB. Hm. Perhaps we should modify the `aux` parameter to be `uintptr_t`.

test/1.4/structure/T_trait.dml
```diff
@@ -152,24 +152,24 @@ method test() throws {
     pp.m(0);
     c.m(0);
     c.o(0);
-    expect(ppp->o(3) == 16, "ppp->o(3) == 16");
-    expect(p[0].o(3) == 12, "p[0].o(3) == 12");;
+    assert ppp->o(3) == 16;
+    assert p[0].o(3) == 12;;
```
lwaern-intel, line 156: Double `;`

lwaern-intel (approved):
Conversion looks good, and the questionable stuff I can spot existed in the tests to begin with.

syssimics:
Verification [#11024273](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11024273): pass

mandolaerik:
> is there off-github discussion I've missed?

Tests added lately have used `assert` exclusively

> The ability of assert to both encourage documentation of _what_ exactly is being tested, as well as the ability to give different errors for, for example, different failing indexes in an array seem beneficial to me.

Two problems: First, `expect` doesn't provide a line number, so you often end up writing redundant noise for uniqueness, like `expect(x == 14, "x == 14");`, whereas an `assert` failure points you right to the failing line. Second, in order to create a parameterized message you must do string formatting on the side, which is so cumbersome that we never have done that.

We had a handful of tests where the message was used to declare intent; I changed those into comments.

I think the main remaining advantage of the expect/assert split was that you could use both within one test, where `expect` is used for the compiler functionality under test, whereas `assert` is used for testing the test itself. But we didn't make much use of that distinction.

test/1.4/lib/T_unmapped_access.dml
mandolaerik: you mean `cast(most_recent_aux, uintptr_t) == 11`?

PR #158, opened 2023-01-27
Remove the testing.dml lib for DML 1.4

mandolaerik:
Generally replace test() with init() as entry point for tests


syssimics:
Verification [#11025265](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11025265): pass

test/1.4/expressions/T_endian_int.dml
```diff
@@ -183,11 +181,15 @@ method test() throws {
 
     // Trait method call with eint arg
     // indirect, throws
-    tp.s_m_t(var);
+    try {
+        tp.s_m_t(var);
+    } catch {}
```
JonatanWaern, line 186: Rather than wrapping in try/catch, just remove the throws from s_m_t

test/1.4/lib/T_registers.dml
```diff
@@ -77,34 +78,36 @@ bank array[b < 2] {
...
+            assert val == 0x50;
+            val = read(0x44, 0xffffffff, NULL);
+            assert val == 0x44;
+            val = read(0x48, 0xffffffff, NULL);
+            assert val == 0x48;
+            val = read(0x54, 0xffffffff, NULL);
+            assert val == 0x54;
+            val = read(0x58, 0xffffffff, NULL);
+            assert val == 0x58;
+        } catch assert false;
```
JonatanWaern, line 110: This would be more optimal with a try/catch around each read (so you get the exact read that fails), but its not a big deal

test/1.4/methods/T_trunc.dml
```diff
@@ -24,11 +22,15 @@ method d(int64 i) -> (int5) {
     return i;
 }
 
-method test() throws {
+method init() {
     local int64 i;
-    i = a(0x5f);
+    try {
```
JonatanWaern, line 27: similarly, remove the throws from methods that cant throw (instead of this some-hiding catch in case they somehow do throw)

test/1.4/structure/T_subobjs.dml
```diff
@@ -37,7 +36,11 @@ bank b[i < 2] {
     saved int bs;
 }
 
-method test() throws {
```
JonatanWaern, line 40: If this pattern is common (say, more than twice) we should keep the testing lib and just not use it unless needed

test/1.4/syntax/T_shadowing.dml
```diff
@@ -20,31 +18,33 @@ method m2() -> (int, int) {
...
+            local (int m2, int v) = m2();
+            assert m2 == 1 && v == 2;
+        }
+        // Original identifiers aren't lost
+        assert x == 1 && !y;
+        assert m() == 1;
+        local int a = mt();
+        local (int b, int c) = m2();
+        assert a == 1 && b == 1 && c == 2;
+    } catch {}
```
JonatanWaern, line 49: similarly, remove the throw from throwing methods if they are not supposed to be able to throw, or assert in the catch if they are not expected to throw

JonatanWaern (dismissed):

test/1.4/expressions/T_endian_int.dml
lwaern-intel: This was my first thought too, but `s_m_t` throwing is intentional and part of the test -- that's why it has the `_t` suffix.

JonatanWaern: you are right. In cases like this I suppose the empty try/catch is fine. If the other cases I've complained about are similar then disregard all of it

test/1.4/structure/T_subobjs.dml
mandolaerik: It's not that common. Also I don't fully agree -- for simple logic, the value of being explicit and standalone often outweighs the cost of duplication. The attribute is the communication channel between X.py and X.dml, and that channel is easier to see if the attribute is explicitly declared rather than inherited from some lib.

test/1.4/expressions/T_endian_int.dml
mandolaerik: Still makes sense to extend the catch block with an `assert false`, for clarity.

test/1.4/lib/T_registers.dml
mandolaerik: I think the main purpose of this code is to validate the read values, the assert false thing might be viewed more as a test consistency check.

test/1.4/methods/T_trunc.dml
mandolaerik: agreed

syssimics:
Verification [#11278880](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11278880): pass

test/1.4/events/T_after_cancel.dml
```diff
@@ -150,6 +149,6 @@ template operator is (int64_attr) {
 attribute multi_operator[i < 2][j < 2] is (operator);
 attribute single_operator is (operator);
 
-method test() throws {
+method init() {
```
lwaern-intel, line 152: Remove this method entirely

test/1.4/expressions/T_stringify.dml
```diff
@@ -29,4 +27,4 @@ method test() throws {
      log info: "float_val evaluated to %s", stringify(float_val);
      /// GREP .* register name became b_r_18
      log info: "register name became %s", b.r.name;
-}
\ No newline at end of file
+}
```
lwaern-intel, line 30: do we like endlines at end of files, do we not, or don't we care either way?

test/1.4/lib/T_qname.dml
```diff
@@ -8,15 +8,12 @@ device test;
 
 group test;
 
-attribute runtest is (write_only_attr) {
```
lwaern-intel, line 11: ain't that the silliest thing. Not sure why I didn't use `testing.dml` here.

test/1.4/methods/T_startup_memoized.py
```diff
@@ -21,5 +21,3 @@ def on_startup(node, meth):
 stest.expect_equal(startup_calls, expected_after_startup)
 obj = SIM_create_object('test', 'obj', [])
 stest.expect_equal(startup_calls, expected_after_startup)
-stest.expect_true(obj.runtest)
-stest.expect_equal(startup_calls, expected_after_startup)
```
lwaern-intel, line 25: I'd complain that this changes the semantics of the test, but it doesn't really.

test/1.4/structure/T_name_parameter.dml
```diff
@@ -38,4 +36,4 @@ method test() throws {
     log info: "b_r_f qname is %s", b.r.f.qname;
     /// GREP .* c_r_f qname is c.NOT_THE_BANK.f
     log info: "c_r_f qname is %s", c.r.f.qname;
-}
\ No newline at end of file
+}
```
lwaern-intel, line 39: here too a newline is added at the end of file

test/1.2/misc/dml14.dml
```diff
@@ -430,26 +430,28 @@ method len_test() {
...
+        assert SIM_attr_integer(ua_get) == 32;
+        cast(ba, bool_attr).set(SIM_make_attr_boolean(true));
+        local attr_value_t ba_get;
+        (ba_get) = cast(ba, bool_attr).get();
+        assert SIM_attr_boolean(ba_get);
+        cast(fa, double_attr).set(SIM_make_attr_floating(4.13));
+        local attr_value_t fa_get;
+        (fa_get) = cast(fa, double_attr).get();
+        assert SIM_attr_floating(fa_get) == 4.13;
+    } catch assert false;
```
lwaern-intel, line 451: Throughout your changes, in some places you do `} catch assert false;`, and in others you do `} catch { assert false; }`. Pick the one you prefer, and use that consistently.

lwaern-intel (changes requested):

test/1.4/structure/T_name_parameter.dml
mandolaerik: I tend to add missing newlines at EOF. I don't have a particular strong aversion for them, only a weak one. Tools tend to agree with me, e.g. diff viewers marks missing nelwines as anomalies (it says "no newline at EOF" on them, rather than "newline at EOF" on my line)

I don't protest when people forget to add EOF newlines, but I don't mind adding them myself.

test/1.2/misc/dml14.dml
mandolaerik: no strong opinion (probably wrote these different days, across moods). Now changed according to majority vote.

syssimics:
Verification [#11295552](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11295552): pass

lwaern-intel (approved):

PR #162, opened 2023-02-16
Fix nondeterminism in `Template.traits()`

lwaern-intel:


py/dml/template.py
```diff
@@ -136,9 +136,9 @@ def site(self): return self.spec.site
     def traits(self):
         '''Return a set of all traits implemented by this template'''
         if self.trait:
-            return {self.trait} | self.trait.ancestors
+            return Set((self.trait, *self.trait.ancestors))
         else:
-            return set().union(
+            return Set().union(
                 *[sup.traits() for (_, sup) in self.spec.templates])
```
JonatanWaern, line 142: assuming you've checked that self.spec.templates is deterministic here

JonatanWaern (approved):

py/dml/template.py
lwaern-intel: Yeah, `ObjectSpec.templates` is always a list.

syssimics:
Verification [#11127895](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11127895): pass

PR #165, opened 2023-03-14
Miscellaneous changes from #144

lwaern-intel:
This PR breaks out a couple of miscellaneous (non after-on-hook related) changes in #144 that were done in order to support the after-on-hook changes, in order to help them get approved faster, as they are also relevant for the work on RAII types.

Some of the changes have been tweaked compared to how they appear in #144. In particular, `deserialize_list_to_targets` no longer has the `pre` and `post` parameters, and its `error_out` parameter has become `error_out_at_index`, which also accepts the index of the target at which it is called. This will be relevant for the work on RAII types.

### DIFF
--- a/py/dml/ctree.py
+++ b/py/dml/ctree.py
@@ -541,6 +541,12 @@ def mkExpressionStatement(site, expr):
         return mkNull(site)
     return ExpressionStatement(site, expr)
 
+def toc_constsafe_pointer_assignment(site, source, target, typ):
+    target_val = mkDereference(site,
+        Cast(site, mkLit(site, target, TPtr(void)), TPtr(typ)))
+    mkAssignStatement(site, target_val,
+                      ExpressionInitializer(mkLit(site, source, typ))).toc()
+
 class After(Statement):
     @auto_init
     def __init__(self, site, unit, delay, domains, method, eventinfo, indices,
@@ -572,7 +578,10 @@ class After(Statement):
                        ', '.join(arg.read() for arg in self.inargs)))
                 out('_data->args = MM_MALLOC(1, '
                     + f'{args_struct});\n')
-                out(f'*({args_struct} *)_data->args = _event_args;\n')
+                toc_constsafe_pointer_assignment(self.site,
+                                                 '_event_args',
+                                                 '_data->args',
+                                                 self.eventinfo['args_type'])
 
             if self.domains:
                 out('_identity_t *_event_domains = '
@@ -960,7 +969,7 @@ class LValue(Expression):
         rt = realtype(self.ctype())
         if isinstance(rt, TEndianInt):
             return (f'{rt.dmllib_fun("copy")}(&{self.read()},'
-                    + f' {source.read()})\n')
+                    + f' {source.read()})')
         return '%s = %s' % (self.read(), source.read())
 
 class IfExpr(Expression):
--- a/py/dml/serialize.py
+++ b/py/dml/serialize.py
@@ -25,8 +25,9 @@ uint64_t = TInt(64, False)
 const_void = void.clone()
 const_void.const = True
 
-serializer_t = TFunction([TPtr(const_void), TPtr(attr_value_t)], void)
-deserializer_t = TFunction([TPtr(attr_value_t), TPtr(void)], set_error_t)
+serializer_t = TFunction([TPtr(const_void), TPtr(void)], attr_value_t)
+deserializer_t = TFunction([attr_value_t, TPtr(void), TPtr(void)],
+                           set_error_t)
 
 # code to insert before body
 serialize_prototypes = []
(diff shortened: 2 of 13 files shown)


syssimics:
Verification [#11258295](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11258295): pass

syssimics:
Verification [#11262762](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11262762): pass

py/dml/ctree.py
```diff
@@ -541,6 +541,12 @@ def mkExpressionStatement(site, expr):
         return mkNull(site)
     return ExpressionStatement(site, expr)
 
+def toc_constsafe_pointer_assignment(site, source, target, typ):
```
mandolaerik, line 544: better naming and purity if you let the caller call `.toc()`

py/dml/serialize.py
```diff
@@ -372,80 +393,158 @@ def type_signature(dmltype, is_for_serialization):
                       if not is_for_serialization else '')
     assert False
 
+def serialize_sources_to_list(site, sources, out_attr):
```
mandolaerik, line 396: naming: the word `sources` is kind of redundant, can be removed (or replaced if you prefer).

py/dml/serialize.py
```diff
@@ -372,80 +393,158 @@ def type_signature(dmltype, is_for_serialization):
...
+            else:
+                sub_serialize = ctree.mkAssignStatement(
+                    site, imm_attr_ref, ctree.ExpressionInitializer(source))
+        sim_attr_list_set_statement = call_c_fun(
+            site, "SIM_attr_list_set_item", [ctree.mkAddressOf(site, out_attr),
+                                             index,
+                                             imm_attr_ref])
+        statements += [sub_serialize, sim_attr_list_set_statement]
+    ctree.mkCompound(
+        site, [attr_assign_statement, imm_attr_decl] + statements).toc()
```
mandolaerik, line 426: moving the `.toc()` to caller would make this function pure

py/dml/serialize.py
```diff
@@ -372,80 +393,158 @@ def type_signature(dmltype, is_for_serialization):
...
-                      ctree.mkDereference(site, out_arg)).toc()
+                      out_arg).toc()
         else:
             assert False
+        output.out("return out;\n")
         output.out("}\n", preindent = -1)
     serialize_function_code.append(func_code.buf)
     return function_name
 
+def deserialize_list_to_targets(site, val_attr, targets, error_out_at_index,
```
mandolaerik, line 468: `targets` word redundant, `deserilize_list` would suffice

py/dml/serialize.py
```diff
@@ -372,80 +393,158 @@ def type_signature(dmltype, is_for_serialization):
...
+        site, is_list_expr, deserialization,
+        ctree.mkCompound(
+            site,
+            error_out_at_index(
+                0, # TODO: None instead?
+                'Sim_Set_Illegal_Type',
+                f'expected list of size {len(targets)}'
+                + f' during {error_context}' * bool(error_context))
+            )
+    ).toc()
```
mandolaerik, line 532: Delegating `.toc` to caller would make this function pure

mandolaerik (approved):
Looks good overall, good if you can resolve some naming issues before merge but it's not blocking really.

Some improvements seem to be observable, like improved codegen for list serialization. You could consider a releasenoted, though not strictly needed.

py/dml/serialize.py
lwaern-intel: I don't really like the terseness of `serialize_list`... I feel like it's too easy to confuse it with talking about a specific serialized datatype, while this function is for a much more generic purpose than that. If you insist upon it though...

py/dml/serialize.py
lwaern-intel: The fact `.toc()` of statements are used to implement this is an implementation detail, just like everything else in serialize.py.

py/dml/ctree.py
lwaern-intel: err... Not sure, when I wrote this I did not mean for this to be used outside of C emit phase. True, it _could_ with its current implementation, but do I want it to _need_ to be able to be used in the codegen phase? I.e. do I want to restrict it from being able to call `out()`?

In other words, I dislike the fact that it would be expressed in terms of codegen expression/statements. If we care about purity, I'd much rather we express it in terms of a datatype representing C emitting.

py/dml/serialize.py
lwaern-intel: See other comment.

py/dml/serialize.py
lwaern-intel: See other comment.

PR #166, opened 2023-03-15
Add porting rule for converting int1 to uint1

mandolaerik:
Some users reported that after `int1 x = 1`, the expression `x == 1`
evaluates to `true` in 1.2 and `false` in 1.4; grepping around, this
pattern is fairly common; thus, the uint1 type seems to match better
what int1 means in practice in 1.2.


syssimics:
Verification [#11263833](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11263833): fail

syssimics:
Verification [#11268292](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11268292): fail

syssimics:
Verification [#11268835](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11268835): pass

PR #168, opened 2023-03-21
Handle third party trademarks

lwaern-intel:
SIMICS-20769, SIMICS-20537, and SIMICS-20761

### DIFF
--- a/md_to_github.py
+++ b/md_to_github.py
@@ -53,6 +53,11 @@ copyright = '''<!--
 -->
 '''
 
+acknowledgement = r'''
+***
+<sub>\* Other names and brands may be claimed as the property of others.</sub>
+'''
+
 frontpage_links = []
 
 filename_map = {}
@@ -93,6 +98,9 @@ with tarfile.open(outfile, "w:gz") as tgz:
             end = match.end(1)
             if end > start:
                 body = body[:start] + filename_map[match[1]] + body[end:]
+
+        if validate_md_links.third_party_trademark_re.search(body) is not None:
+            body += acknowledgement
         add_to_tar(tgz, path, copyright + body)
 
     head = [copyright,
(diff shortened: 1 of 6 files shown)


validate_md_links.py
```diff
@@ -24,6 +24,12 @@ def toc_md_files(toc, dirs):
 
 assert anchor_re.findall('<a id="comparison-to-c"/>') == ['comparison-to-c']
 
+third_party_trademarks = {'synopsys', 'coverity'}
+
+third_party_trademark_re = re.compile(
+    r'(?i:(' + '|'.join(third_party_trademarks)
+    + r'))([*®™]|&reg;|&trade;|&copy;|\\\*)?')
```
mandolaerik, line 31: urgh, I can't read this. Can you document what you mean with a unit test? A simple `assert third_party_trademark_re.match(...) == ...` on top level is sufficient.

Also, if you want to capture this in the future, then I suggest allowlist rather than blocklist, i.e. complain if the word before `®` or `™` does *not* match smth like Intel, Simics, DML.

md_to_github.py
```diff
@@ -53,6 +53,11 @@ def toc_md_files(toc, indices):
 -->
 '''
 
+acknowledgement = r'''
+***
+<sub>\* Other names and brands may be claimed as the property of others.</sub>
+'''
+
```
mandolaerik, line 60: This should not be part of the md_to_github script: The md sources are compiled into both simics-native docs and into github wiki, so better fix this in the md sources instead. Or is this stuff also auto-added by dodoc?

lwaern-intel: The Simics Online Documentation will have a separate solution that won't require authors to manually add acknowledgment lines. So this DML wiki exclusive logic should indeed be DML wiki exclusive.

lwaern-intel: Alternatively... the DML wiki could have a `* Legal Information` page... hmm...

validate_md_links.py
mandolaerik: If you want to protect from using external trademarks without `*`, then  you need a much bigger list, because the world has more than two trademarked things.

And if we are to make this kind of check, then it's better to do it globally for all simics docs.

PR #171, opened 2023-04-11
Support Simics API 7 for DML 1.2

jhbaarnh:
DML 1.2 is marked as legacy in 7, not deprecated.

This is blocking builds in 7 when the major version is bumped.

### DIFF
--- a/py/dml/toplevel.py
+++ b/py/dml/toplevel.py
@@ -353,9 +353,9 @@ def parse_main_file(inputfilename, explicit_import_path, strict):
                      orig_path]]
 
     if version == (1, 2) and dml.globals.api_version not in (
-            # we may want to add "7" if we want to postpone the
+            # we may want to add "8" if we want to postpone the
             # deprecation of DML 1.2
-            "4.8", "5", "6", "internal"):
+            "4.8", "5", "6", "7", "internal"):
         raise ESIMAPI(site, fmt_version(version),
                        dml.globals.api_version)
 


syssimics:
Verification [#11394215](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11394215): pass

lib/1.4/dml-builtins.dml
```diff
@@ -551,8 +551,10 @@ template device {
     param be_bitorder             default _be_bitorder;
     param log_group                 = undefined;
 
-    param use_io_memory             default simics_api_version == "6"
-        || simics_api_version == "5" || simics_api_version == "4.8";
+    param use_io_memory             default simics_api_version == "7"
+        || simics_api_version == "6"
+        || simics_api_version == "5"
+        || simics_api_version == "4.8";
```
mandolaerik, line 557: io_memory is a legacy feature that we want to deprecate, so please don't extend "default true" into the 7 API

jhbaarnh: Ok you're saying I should instead set all the devices using it to SIMICS_API = 6.

lwaern-intel: ... Or that devices with banks using it should set `param use_io_memory = true` explicitly. Note that top-level `use_io_memory` is just a default for banks within the device, and banks can override it locally on a bank-by-bank basis.

jhbaarnh: Or alternatively, explicitly set `use_io_memory = true` in those devices.

syssimics:
Verification [#11398387](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11398387): pass

lib/1.4/dml-builtins.dml
mandolaerik: And not just in devices *using* it, but in devices actually *needing* it. Most "normal" devices should work fine wtih a silent switch to the transaction iface.

py/dml/toplevel.py
```diff
@@ -353,9 +353,7 @@ def parse_main_file(inputfilename, explicit_import_path, strict):
                      orig_path]]
 
     if version == (1, 2) and dml.globals.api_version not in (
-            # we may want to add "7" if we want to postpone the
-            # deprecation of DML 1.2
```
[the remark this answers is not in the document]
jhbaarnh: Fixed

mandolaerik (approved):

lib/1.4/dml-builtins.dml
jhbaarnh: I created SIMICS-20984 for this.

PR #172, opened 2023-04-13
Improve porting of methods with one outarg

mandolaerik:
- Repair indentation
- Random polish
- Polish porting of return statements
- Avoid inserting locals for outargs if unused
- Add release note


syssimics:
Verification [#11403412](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11403412): fail

syssimics:
Verification [#11403597](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11403597): fail

syssimics:
Verification [#11403705](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11403705): pass

py/dml/ctree.py
```diff
@@ -4610,7 +4610,7 @@ def log_statement(site, node, indices, logtype, level, groups, fmt, *args):
         groups = mkIntegerLiteral(site, 0)
 
     inargtypes = (([TInt(32, True)] if lvl else [])
-                  + [TPtr(TNamed("conf_object_t")), TInt(32, True),
+                  + [TPtr(TNamed("conf_object_t")), TInt(64, False),
```
lwaern-intel, line 4613: What's with this change?

py/dml/dmlparse.py
```diff
@@ -479,6 +479,32 @@ def maybe_default_no(t):
...
+            report(POUTARGRETURN(lh.site, start_site(rh.site), None))
+            # suppress PRETVAL's insertion of return statement
+            ends_with_return = True
+    report(PRETVAL(
+        site,
+        DumpableSite(file_info, start),
+        DumpableSite(file_info, end),
+        DumpableSite(file_info, rparen),
+        [(psite.loc(), pname) for (_, psite, pname, _) in outp]))
+    if not ends_with_return:
```
lwaern-intel, line 501: This is overly aggressive, as method bodies may not be fallthrough (e.g. last statement may be `throw;`).

py/dml/dmlparse.py
```diff
@@ -479,6 +479,32 @@ def maybe_default_no(t):
...
+def report_pretval(site, file_info, start, end, rparen, outp, stmts):
+    ends_with_return = stmts and stmts[-1].kind == 'return'
+    if (len(outp) == 1
+        and not ends_with_return
+        and stmts
+        and stmts[-1].kind == 'expression'
+        and stmts[-1].args[0].kind == 'set'):
+        (lh, rh) = stmts[-1].args[0].args
+        if lh.kind == 'variable_dml12' and lh.args[0] == outp[0][2]:
+            report(POUTARGRETURN(lh.site, start_site(rh.site), None))
```
lwaern-intel, line 492: Isn't this redundant in the face of the check in `stmt_compound`?

lwaern-intel: Oh, no, it isn't, because in this case it doesn't end with return... hah.

py/dml/codegen.py
```diff
@@ -1394,6 +1394,21 @@ def stmt_compound(stmt, location, scope):
     [stmt_asts] = stmt.args
     lscope = Symtab(scope)
     statements = codegen_statements(stmt_asts, location, lscope)
+    if (logging.show_porting and stmt.site.dml_version() == (1, 2)
+        and not stmt.site.filename().endswith('dml-builtins.dml')):
+        method = location.method()
+        if method is not None and len(method.outp) == 1:
```
lwaern-intel, line 1400: Why restrict this check to the single-output case? It could be generalized -- go through the list, once you encounter an assignment to an output parameter, walk through the subsequent statements as long as they're assignments to output parameters. Upon the first non-output parameter assignment you encounter, if it's a return, report `POUTARGRETURN` for all the output parameters you've encountered.

py/dml/codegen.py
```diff
@@ -1394,6 +1394,21 @@ def stmt_compound(stmt, location, scope):
     [stmt_asts] = stmt.args
     lscope = Symtab(scope)
     statements = codegen_statements(stmt_asts, location, lscope)
+    if (logging.show_porting and stmt.site.dml_version() == (1, 2)
```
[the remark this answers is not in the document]
mandolaerik: Perhaps cleaner to rewrite it to operate on AST objects instead, i.e. before calling `codegen_statements`? That way we don't need to know how `return;` is transformed into `Goto`.

py/dml/ctree.py
mandolaerik: This is from the 'random cleanups' commit. uint64 is now a better approximation than uint32 of the loggroup arg of `SIM_LOG_*`. I don't think this change makes a difference to the behaviour, it's just for conciseness.

py/dml/dmlparse.py
mandolaerik: True, but this is not a regression in this patch; previously ends_with_return was passed to PRETVAL, which triggered what is now PRETVAL_END; this is only a refactoring. I think bodies that end with throw give parse errors that are easy to fix.

syssimics:
Verification [#11429209](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11429209): pass

py/dml/codegen.py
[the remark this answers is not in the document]
mandolaerik: It can't be done in the parser, since it relies on `location.method().outp`. We could add some parser state that bookkeeps "current method", but I think that would be unnecessarily intrusive. There is a value in keeping porting hacks local and contained.

There is maybe a risk that we miss porting some corners; we can fix that if it appears to be a problem in practice.

py/dml/codegen.py
mandolaerik: Also, note that the typical two-outarg case is something like:
```
method m() -> (bool ok, int value) {
  if ($valid()) {
   ok = false;
   return;
 } else {
   ok = true;
   value = $result();
 }
}
```
where one of the args is dontcare, which is somewhat hard to translate automatically.

mandolaerik: I realized that I missed one case: I don't remove the intermediate variable in the case `call $foo() -> (value); return;`  where `value` is an outarg. Crude grepping suggests that this happens in about 10 cases out of ~300 in our repo, so not necessarily worthwhile fixing.

lwaern-intel: Hmm! That's an argument in favor of checking the `ctree` representation! DML 1.2 calls with one return value are codegen:d using `mkCopyData`, i.e. `ExpressionStatement(AssignOp(...))`, in order to assign to the outarg, which means the old check _would_ catch that case.

mandolaerik: It would, but the porting rule converts `x = Y;` by replacing the `x =` string with `return `, which may be problematic if the sites point into the expression `call Y() -> (x)`. Though, with a bit of luck the porting script might have converted that call to `x = Y()` and managed to do a corresponding transformation of all sites.

lwaern-intel: Up to you. Like you said, it's about a "good enough" heuristic. I'm ok with approving this PR as-is.

PR #174, opened 2023-04-19
Support function_mapped_bank on banks with use_io_memory=false

mandolaerik:
The old PCI framework (pci-bus) depends on function_mapped_bank when
mapping pci_config, and it is natural for devices to transition to
transaction interface as an intermediate step before moving to the new
PCIe framework. Furthermore, the default of use_io_memory changes to
false in Simics 7.

### DIFF
--- a/lib/1.4/utility.dml
+++ b/lib/1.4/utility.dml
@@ -1226,12 +1226,34 @@ template function_io_memory {
 
     method operation(generic_transaction_t *mem_op,
                      map_info_t map_info) -> (exception_type_t) {
+        local uint64 offset = map_info.start - map_info.base;
         foreach b in (each function_mapped_bank in (dev)) {
             if (b.function == map_info.function) {
-                if (b.io_memory_access(mem_op,
-                                       (SIM_get_mem_op_physical_address(mem_op)
-                                        - map_info.base + map_info.start),
-                                       NULL)) {
+                if (!b.use_io_memory) {
+                    local map_target_t *mt = SIM_new_map_target(b._bank_obj(),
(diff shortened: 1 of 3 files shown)


mandolaerik:
Needed for https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/3737

syssimics:
Verification [#11436709](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11436709): pass

svenwest (commented):
This looks very strange to me. Why would you have transaction-banks in a device, but still use function-mapped io-memory on the top level?

mandolaerik:
> This looks very strange to me. Why would you have transaction-banks in a device, but still use function-mapped io-memory on the top level?

Because of the wonders of legacy PCI!

As I understand it, a PCI dev asks the pci-bus to map its pci_config space to function number 255 of a memory-space. The knowledge of how to map 255 to a bank is internal to the device (maybe it does the same for other banks as well, but pci_config was the one that first triggered a test failure). So if the config bank is based on `transaction`, which is the default in Simics 7, then we end up in this corner. I think it is important to allow people to move from `io_memory` to `transaction` within the old pci lib as a separate step , before moving to the new pcie.

Note that the top level of a legacy PCI device must always specifically implement io_memory for this reason, because the bus requires an interface that support function mapping.

We could change pci-bus to map `.bank.pci_config` instead, but that would maybe fail on existing non-DML PCI implementations, so I thought it's easier and less risky to repair this legacy feature.

svenwest (approved):

PR #175, opened 2023-04-24
Avoid relying on which exact exception an attribute getter throws

mandolaerik:
This is not relevant for the correctness of DMLC. Needed for smoother
migration of:
https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/3767

### DIFF
--- a/test/1.2/structure/T_attribute_throw.py
+++ b/test/1.2/structure/T_attribute_throw.py
@@ -3,30 +3,14 @@
 
 import stest
 
-try:
+with stest.expect_exception_mgr(SimExc_IllegalValue):
     obj.a = None
-except SimExc_IllegalValue as e:
-    pass
-else:
-    stest.fail('expected exception')
 
-try:
-    _ = obj.a
-except SimExc_Attribute as e:
-    pass
-else:
-    stest.fail('expected exception')
+with stest.expect_exception_mgr(SimExc_General):
+    SIM_get_attribute(obj, "a")
 
-try:
+with stest.expect_exception_mgr(SimExc_IllegalValue):
     obj.b = None
-except SimExc_IllegalValue as e:
-    pass
-else:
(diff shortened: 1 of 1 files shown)


syssimics:
Verification [#11463197](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11463197): pass

test/1.2/structure/T_attribute_throw.py
```diff
@@ -3,30 +3,14 @@
...
-else:
-    stest.fail('expected exception')
 
-try:
-    _ = obj.b
-except SimExc_Attribute as e:
-    pass
-else:
-    stest.fail('expected exception')
+with stest.expect_exception_mgr(SimExc_General):
```
lwaern-intel, line 15: Does this really catch `AttributeError`?

mandolaerik: Documentation of `SIM_get_attribute` says:
```
   <li><b>SimExc_General</b> If <fun>SIM_attribute_error</fun> or
   <fun>SIM_c_attribute_error</fun> was called inside the getter function.</li>
```
It doesn't mention what to expect when a setter returns a non-`Sim_Set_Ok` value, though.

And yes, `SimExc_General` is a superclass of all `SimExc_*` exceptions, so the new check will still succeed if we decide to change the raised exception to a plain `SimExc_Attribute` in some cases.

lwaern-intel: I don't understand -- https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/3767 is about modifying the exception _from_ `SimExc_Attribute` _to_ `AttributeError`. `SimExc_General` is not a superclass of `AttributeError` -- how will this aid in migration of the Simics Base PR?

lwaern-intel (approved):

PR #177, opened 2023-04-27
Do not use 5 API tests in Simics 7

jhbaarnh:


test/tests.py
```diff
@@ -117,7 +117,7 @@ def dmlc_reaper_args(exitcode_file, timeout_multiplier = 1):
 
 os.environ['DMLC_DEBUG'] = 't'
 
-latest_api_version = "6"
+latest_api_version = {6: '6', 7: '7'}[int(get_simics_major())]
```
mandolaerik, line 120: why `int()`? wouldn't `{'6': '6', '7':'7'}`  be more direct? Which in turn can be further simplified, which I leave as an exercise to the reader :)

Maybe there is some point in keeping some form of `assert latest_api_version in {'6', '7'}` somewhere, to early smoke out gross dmlc vs simics mismatches. But perhaps that check is better done in `generate_env.py`?

mandolaerik: btw, note that there is a function `simicsutils.internal.default_api_version` that you may or may not prefer over `get_simics_major` (I don't have a strong opinion myself).

jhbaarnh: Changing `latest_api_version` leads to other test failures, so I will do that change separately:
https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11477578

E.g. these errors show up again
```
Exception: Legacy format for Register constructor arguments unsupported for banks implementing the transaction interface. Please change Register((obj, "bankname", offset)) into Register(obj.bank.bankname, offset)

syssimics:
Verification [#11477990](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11477990): fail

test/tests.py
mandolaerik: ok, that's a convincing reason to keep `get_simics_major`

syssimics:
Verification [#11478285](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11478285): fail

syssimics:
Verification [#11478436](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11478436): fail

syssimics:
Verification [#11478589](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11478589): pass

mandolaerik (approved):

PR #179, opened 2023-05-02
Remove integer_t and uinteger_t in Simics 7

jhbaarnh:


syssimics:
Verification [#11500530](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11500530): fail

syssimics:
Verification [#11500853](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11500853): pass

mandolaerik (commented):
There is no test. See 1.4/errors/T_ETYPE.dml, but I suppose you need to do the same trick as   for WREF.

syssimics:
Verification [#11503601](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11503601): fail

test/1.4/errors/INTEGER.dml
mandolaerik, line 1: I would merge these two tests, and name them ETYPE_integer_t. And T_ETYPE takes care of all declaration corners, so sufficient with one declaration per type. Also, WBN with a test that the types *are* available in simics 6 (`/// DMLC-FLAG --simics-api=6` should make this test work in 7 too).

Some other alternatives:
* Softcode  api_versions somehow, i.e. make `--simics-api=7` permitted also in 6, so we can use the same tests.
* Write a low-level test instead, i.e., `types_test.py` that does `realtype(TNamed('integer_t'))` and expects exception or result depending on version.
  * with this alternative, it's easy to switch to checking `simics_major` instead of `api_version`; this variable is not currently exposed but that's easy to fix, see `generate_env.py`.

We should also find out why `dml-api` didn't include `integer_t` automatically in `simics/obsolete/4_8.dml`. This does however not resolve the problem for us; the 6 API already allows the no-longer-simics type integer_t and we need a compiler hack to preserve that.

syssimics:
Verification [#11504386](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11504386): fail

syssimics:
Verification [#11509303](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11509303): fail

syssimics:
Verification [#11509563](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11509563): pass

test/1.4/errors/INTEGER.dml
jhbaarnh: Tests merged and your other suggested test in the 1st paragraph added. The last point I leave to you for later.

mandolaerik (approved):
Don't forget to mention this in a migration guide somewhere.

jhbaarnh:
Yes these types were already in the deprecation table.

PR #180, opened 2023-05-05
Use unique identifier for the device in `index_ident` -- SIMICS-21044

lwaern-intel:

PR #181, opened 2023-05-05
Avoid broken C code generation, SIMICS-21044

mandolaerik:
If a top-level object's name equals dev.name, then they could yield
conflicting declarations of the `_each_X__in__NAME` variable.

### DIFF
--- a/RELEASENOTES-1.4.docu
+++ b/RELEASENOTES-1.4.docu
@@ -469,4 +469,9 @@
       This syntax can't be used for variadic arguments or any argument
       corresponding to a method parameter declared `inline`
       <bug number="SIMICS-20473"/>.</add-note></build-id>
+  <build-id value="next"><add-note>Fixed a bug where a top-level
+      object named the same as the device object could result in
+      invalid generated C when both objects are targets of <tt>each
+      .. in ..</tt> expressions
+      <bug number="SIMICS-21044"/>.</add-note></build-id>
 </rn>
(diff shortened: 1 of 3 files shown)


syssimics:
Verification [#11517255](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11517255): pass

mandolaerik:
duplicates #180

RELEASENOTES-1.4.docu
```diff
@@ -469,4 +469,9 @@
       This syntax can't be used for variadic arguments or any argument
       corresponding to a method parameter declared `inline`
       <bug number="SIMICS-20473"/>.</add-note></build-id>
+  <build-id value="next"><add-note>Fixed a bug where a top-level
+      object named the same as the device object could result in
+      invalid generated C when both objects are targets of <tt>each
+      .. in ..</tt> expressions
+      <bug number="SIMICS-21044"/>.</add-note></build-id>
```
lwaern-intel, line 476: Curious line break. How come you didn't do
```
      .. in ..</tt> expressions <bug number="SIMICS-21044"/>.
  </add-note></build-id>
```
?

mandolaerik: That's a convention I follow for merge conflict resolution reasons: if two notes are added in simultaneous PRs to the same file, and both add identical `</add-note></build-id>` lines, then merge tools will tend to merge those lines. So if you always attach the </add-note> to the previous word, then your note will never be part of an unpleasant merge.

lwaern-intel: I had guessed as much. Wasn't sure it would trump having line breaks be at regular points, but it doesn't matter to me.

syssimics:
Verification [#11534916](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11534916): pass

lwaern-intel (approved):

PR #182, opened 2023-05-05
Improve DML_DUMP_INPUT_FILES

mandolaerik:
In some SSM use cases, some files are imported both as
full/path/to/file.dml and file.dml, which somehow prohibits unification.
Here we fix this by unifying files identical basenames and content.


syssimics:
Verification [#11518992](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11518992): pass

PR #184, opened 2023-05-22
Fix various issues relating to `after` on methods with const parameters

lwaern-intel:
Our unconstifying logic has always been suspect, but it turns out some pieces are not even tested -- deserialization in particular.

Four different issues, fixes for which have been packaged in three different commits. Check each commit for a brief comment on each.

The burden of maintaining deconstifying logic is a rather compelling argument to try to avoid its need in the first place -- by restricting the places where `const` may occur. We already have some restrictions in place, but only a few are codified in the language, while other are explicit limitations of the compiler, upheld by ICEs, and to be fixed at a later date. If we want to commit to "we don't want to bother with this idiocy" then we need to formalize it. For example, we could say "input parameters may not be (partially) const." This is likely to be breaking, but since the fix is rather damn easy (just remove the const qualifier) for every case but for partially const types and external const types -- both of which are rather niche -- we may be ok with codifying such restrictions. We're gonna need to discuss this.

### DIFF
--- a/RELEASENOTES-1.4.docu
+++ b/RELEASENOTES-1.4.docu
@@ -482,4 +482,7 @@
   <build-id value="next"><add-note>Fixed an error where const method parameters
       would cause an internal error if that method is used as the callback of
       an `after` statement.</add-note></build-id>
+  <build-id value="next"><add-note>Fixed an issue where having a const-qualified
+      layout type could lead to internal compiler errors in certain
+      scenarios.</add-note></build-id>
 </rn>
--- a/RELEASENOTES-1.4.docu
+++ b/RELEASENOTES-1.4.docu
@@ -477,4 +477,6 @@
   <build-id value="next"><add-note><tt>--coverity</tt> now suppresses a false
       positive that the <tt>COPY_PASTE_ERROR</tt> checker reports in
       <tt>dml-builtins.dml</tt>.</add-note></build-id>
+  <build-id value="next"><add-note>It's now allowed to cast an expression
+      of a struct type to that same struct type.</add-note></build-id>
 </rn>
--- a/RELEASENOTES-1.4.docu
+++ b/RELEASENOTES-1.4.docu
@@ -479,4 +479,7 @@
       <tt>dml-builtins.dml</tt>.</add-note></build-id>
   <build-id value="next"><add-note>It's now allowed to cast an expression
       of a struct type to that same struct type.</add-note></build-id>
+  <build-id value="next"><add-note>Fixed an error where const method parameters
+      would cause an internal error if that method is used as the callback of
+      an `after` statement.</add-note></build-id>
 </rn>
--- a/py/dml/ctree.py
+++ b/py/dml/ctree.py
@@ -4058,16 +4058,22 @@ def mkCast(site, expr, new_type):
             return mkTraitUpcast(site, expr, real.trait)
     old_type = safe_realtype(expr.ctype())
     if (dml.globals.compat_dml12_int(site)
-        and (isinstance(old_type, (TStruct, TVector, TLayout))
-             or isinstance(real, (TStruct, TVector, TLayout)))):
+        and (isinstance(old_type, (TStruct, TVector))
+             or isinstance(real, (TStruct, TVector)))):
         # these casts are permitted by C only if old and new are
         # the same type, which is useless
         return Cast(site, expr, new_type)
-    if isinstance(real, (TVoid, TStruct, TArray, TVector, TTraitList,
-                         TLayout, TFunction)):
+    if isinstance(real, TStruct):
+        if isinstance(old_type, TStruct) and old_type.label == real.label:
+            return expr
         raise ECAST(site, expr, new_type)
-    if isinstance(old_type, (TVoid, TStruct, TVector, TTraitList, TLayout,
-                             TTrait)):
+    if isinstance(real, TExternStruct):
+        if isinstance(old_type, TExternStruct) and old_type.id == real.id:
+            return expr
+        raise ECAST(site, expr, new_type)
+    if isinstance(real, (TVoid, TArray, TVector, TTraitList, TFunction)):
+        raise ECAST(site, expr, new_type)
+    if isinstance(old_type, (TVoid, TStruct, TVector, TTraitList, TTrait)):
         raise ECAST(site, expr, new_type)
     if old_type.is_int and old_type.is_endian:
         expr = as_int(expr)
--- a/test/1.4/expressions/T_cast.dml
+++ b/test/1.4/expressions/T_cast.dml
@@ -1,0 +1,32 @@
+/*
+  © 2023 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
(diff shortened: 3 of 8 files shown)


RELEASENOTES-1.4.docu
```diff
@@ -474,4 +474,12 @@
...
       .. in ..</tt> expressions
       <bug number="SIMICS-21044"/>.</add-note></build-id>
+  <build-id value="next"><add-note>It's now allowed to cast an expression
+      of a struct type to that same struct type.</add-note></build-id>
+  <build-id value="next"><add-note>Fixed an error where const method parameters
+      would cause an internal error if that method is used as the callback of
+      an `after` statement.</add-note></build-id>
+  <build-id value="next"><add-note>Fixed an issue where having a const-qualified
+      layout type could lead to internal compiler errors in certain
+      scenarios.</add-note></build-id>
```
lwaern-intel, line 484: Do these largely-independent microissues deserve JIRA issues? Or maybe a single JIRA issue for them should be created collectively, despite the fact they're independent?

py/dml/ctree.py
```diff
@@ -4063,8 +4063,16 @@ def mkCast(site, expr, new_type):
...
-    if isinstance(real, (TVoid, TStruct, TArray, TVector, TTraitList,
-                         TLayout, TFunction)):
+    if isinstance(real, TStruct):
+        if isinstance(old_type, TStruct) and old_type.label == real.label:
+            return expr
+        raise ECAST(site, expr, new_type)
+    if isinstance(real, TExternStruct):
+        if isinstance(old_type, TExternStruct) and old_type.id == real.id:
+            return expr
+        raise ECAST(site, expr, new_type)
```
lwaern-intel, line 4073: The relevant use I discovered of `mkCast` where this change is important is `ExpressionInitializer.assign_to` under the `deep_const` case. Since our internal use of `mkCast` is so prolific, I think this is preferable to an ad-hoc solution inside `ExpressionInitializer.assign_to`.

RELEASENOTES-1.4.docu
```diff
@@ -474,4 +474,12 @@
       invalid generated C when both objects are targets of <tt>each
       .. in ..</tt> expressions
       <bug number="SIMICS-21044"/>.</add-note></build-id>
+  <build-id value="next"><add-note>It's now allowed to cast an expression
+      of a struct type to that same struct type.</add-note></build-id>
```
lwaern-intel, line 478: This is DML 1.4 exclusive (or rather, non-`compat_dml12_int` exclusive) as DMLC under 1.2 just allows any cast from or to a struct type to slip by it, thus making GCC pick up the slack of reporting bad casts.

syssimics:
Verification [#11599515](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11599515): fail

syssimics:
Verification [#11599533](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11599533): pass

RELEASENOTES-1.4.docu
mandolaerik: I'd say zero or one jira makes sense. It's kind of nice to have a jira for historic reference, but the use is probably limited in practice (unclear how much effort we'll put into publicizing history if we move to a public issue tracker).

py/dml/ctree.py
mandolaerik: OTOH, one could argue that internal mkCasts are supposed to know what they are doing, so it should generally be hard to inject a TExternStruct into them.

But that's currently an academic question, I'm all for keeping the check in mkCast, I would do the same given what our code base looks like.

test/1.4/events/T_after_const_param.py
```diff
@@ -0,0 +1,13 @@
...
+import stest
+
+cpu = SIM_create_object("clock", "clock", [["freq_mhz", 1]])
+obj.queue = cpu
+
+obj.trigger = None
+SIM_continue(99999)
+stest.expect_equal(obj.res, [0, 0])
+SIM_continue(2)
+stest.expect_equal(obj.res, [4, 7 << 32 & 4])
```
mandolaerik, line 13: `7 << 32 & 4` evaluates to zero? did you mean to use `|`?

Also, this is a rather specific test, its running time is grossly dominated by startup. Would it make sense to merge it into some other test, maybe `T_after`?

test/1.4/expressions/T_cast.dml
mandolaerik, line 1: Perhaps promote `1.2/operators/cast` to 1.4 and merge with that?

BTW, did this code fail before your fix?

lwaern-intel: It did, yes.

test/1.4/events/T_after_const_param.py
lwaern-intel: Oops, yeah I did mean to use `|`.

> Would it make sense to merge it into some other test, maybe `T_after`?

Not _really_, as `T_after` is meant to test basic functionality, but... one could _force_ it to fit, if you insist.

mandolaerik: I'd say the scope of T_after is all functional aspects of after statements. It's generally good to keep down the number of tests, for performance reasons.

test/1.4/expressions/T_cast.dml
lwaern-intel: `1.2/operators/cast` features one piece of 1.2 exclusive logic:
```
    local void *f = cast($dev, void *);
```

I remember this. I had to add the following to `mkCast`:
```
    # Allow casts from dev to pointer types in DML 1.2 for compatibility
    # reasons
    if (dml.globals.dml_version == (1, 2) and isinstance(expr, NodeRef)
        and expr.get_ref()[0].objtype == 'device' and isinstance(real, TPtr)):
        return Cast(site, expr, new_type)
```` 
... when I reworked `mkCast` to be restrictive by default.

lwaern-intel: Since this means `mkCast` has both 1.2 exclusive logic and 1.4 exclusive logic to it -- do you still favor merging the two tests? That would mean throwing away tests for one of the two.

mandolaerik: Aww. Yes, better to preserve that lone test case in 1.2 and move the rest. But if you insist that this is a procrastinatable separate step, then I can't argue with that.

lwaern-intel: > But if you insist that this is a procrastinatable separate step, then I can't argue with that.

Insist, no, but I'll take the excuse ;)

syssimics:
Verification [#11638121](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11638121): fail

syssimics:
Verification [#11638473](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11638473): pass

mandolaerik (approved):

syssimics:
Verification [#11642057](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11642057): pass

PR #186, opened 2023-06-07
Don't rely on UB enum semantics for more than 32 log groups

lwaern-intel:
(Edited: see https://github.com/intel/device-modeling-language/pull/186#issuecomment-1602099028)

### DIFF
--- a/RELEASENOTES.docu
+++ b/RELEASENOTES.docu
@@ -1721,4 +1721,7 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
   <build-id value="6205"><add-note> Fixed a bug with how template types were
       handled by DMLC that could cause the compiler to still exhibit
       nondeterministic behaviour.</add-note></build-id>
+  <build-id value="next"><add-note> Added an error if a device declares
+      more than 63 log groups (including the two built-in log groups.)
+  </add-note></build-id>
 </rn>
--- a/doc/1.4/language.md
+++ b/doc/1.4/language.md
@@ -2360,6 +2360,9 @@ is similar to a `constant` declaration, but the value is
 allocated automatically so that all log groups are represented by
 distinct powers of 2 and can be combined with bitwise *or*.
 
+A maximum of 63 log groups may be declared per device (61 excluding the built-in
+`Register_Read` and `Register_Write` log groups.)
+
 ### Typedef Declarations
 
 <pre>
--- a/py/dml/structure.py
+++ b/py/dml/structure.py
@@ -185,9 +185,11 @@ def mkglobals(stmts):
                 new_typedefs.add(name)
             elif stmt[0] == 'loggroup':
                 _, site, name = stmt
+                if len(dml.globals.log_groups) >= 63:
+                    report(ELOGGROUPS(site))
                 dml.globals.log_groups.append(name)
                 new_symbols.append(LiteralSymbol(
-                    name, TNamed('int', const=True), site,
+                    name, TInt(64, False, const=True), site,
                     crep.cloggroup(name)))
             elif stmt.kind != 'constant': # handled above
                 raise ICE(stmt.site, 'bad AST kind: %s' % (stmt.kind,))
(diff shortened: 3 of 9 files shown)


MagnusIntel:
This is only fixed for DML-1.4?

lwaern-intel:
Oops, no, it fixes it for both 1.2 and 1.4.

MagnusIntel:
I was searching the docs for the 32 limit, but could not find anything in
https://simics-download.pdx.intel.com/simics-6/docs/html/model-builder-user-guide/programming-with-dml.html#log-groups
https://simics-download.pdx.intel.com/simics-6/docs/html/dml-1.4-reference-manual/language.html#loggroup-declarations
https://simics-download.pdx.intel.com/simics-6/docs/html/dml-1.4-reference-manual/language.html#log-statements
even though I referenced 2 of the above when closing the HSD. Hence, I don't know if we have documented the limitation somewhere, but if we have, I guess we should change that as well.

doc/1.4/language.md
```diff
@@ -2360,6 +2360,9 @@ is similar to a `constant` declaration, but the value is
 allocated automatically so that all log groups are represented by
 distinct powers of 2 and can be combined with bitwise *or*.
 
+A maximum of 63 log groups may be declared per device (61 exluding the built-in
```
lwaern-intel, line 2363: exluding

lwaern-intel:
I'll delay merging of this even if approved due to the strange fact that despite the seemingly obvious culprit, I can't actually reproduce the original issue

syssimics:
Verification [#11722159](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11722159): fail

lwaern-intel:
Turns out that the original issue had nothing to do with a supposed 32 loggroup limit,, and we all just misdiagnosed what the problem was. So yes, this change is seemingly only cosmetic. It's still a good thing to implement, both for the new error, and also seeing how the C standard says enum constants must be representable by `int`s, and GNUC documentation seemingly not explicitly stating that they're extending upon that -- so we should not assume the enum approach works, even though it really does.

RELEASENOTES.docu
```diff
@@ -1721,4 +1721,9 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
   <build-id value="6205"><add-note> Fixed a bug with how template types were
       handled by DMLC that could cause the compiler to still exhibit
       nondeterministic behaviour.</add-note></build-id>
+  <build-id value="next"><add-note>Fixed an issue where <tt>loggroup</tt>
+      declarations in a device would erroneously and silently cause the 32:nd
+      log group and beyond to not function properly. A DML device may now have
+      up to 63 log groups (61 excluding built-in log groups) without issue
+      <bug number="22018374443"/>.</add-note></build-id>
```
lwaern-intel, line 1728: Fun question: now that the original issue turned out to be completely unrelated, and (mostly) user error, should the releasenote still reference it?

syssimics:
Verification [#11757482](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11757482): pass

RELEASENOTES.docu
mandolaerik: If the change is not user-visible in practice, then no release note is needed. The stakeholders of that ticket don't need a notification of this change, because they are unaffected by this package upgrade.

py/dml/structure.py
```diff
@@ -185,9 +185,11 @@ def mkglobals(stmts):
                 new_typedefs.add(name)
             elif stmt[0] == 'loggroup':
                 _, site, name = stmt
+                if len(dml.globals.log_groups) >= 63:
+                    report(ELOGGROUPS(site))
                 dml.globals.log_groups.append(name)
                 new_symbols.append(LiteralSymbol(
-                    name, TNamed('int', const=True), site,
+                    name, TInt(64, False, const=True), site,
```
mandolaerik, line 192: I think this is a real bugfix, probably possible to capture with some nonsensical expression like `lg32 < 0`, but probably not worth adding a test.

mandolaerik (approved):

PR #187, opened 2023-06-14
Remove module signing

jhbaarnh:
Module signing is not needed since Simics-Base 6.0.72.

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -698,15 +698,7 @@ class CTestCase(DMLFileTestCase):
                                  env = env)
         if status != 0:
             self.pr("LD: %r" % args)
-            return status
-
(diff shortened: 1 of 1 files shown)


syssimics:
Verification [#11715277](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11715277): fail

jhbaarnh:
This is particularly needed for 7 (and blocks merge of a PR), where module signing is being phased out even more. As mentioned above we have not required signed modules for more than 2 years.

jhbaarnh:
The failure in verification is due to current infra problems in IGK.

lwaern-intel (approved):
Yup, doesn't seem to cause any obvious issues.

PR #188, opened 2023-06-15
Make equalities with zero-initialized traitrefs raise critical error

lwaern-intel:
Relies on [Have zero-initialized _identity_t be invalid](https://github.com/intel/device-modeling-language/commit/3b036b5a62d50ec269b8336203a8eef10e586dbc), cherry-picked from #144.


include/simics/dmllib.h
```diff
@@ -766,11 +766,40 @@ _deserialize_object_trait_reference(ht_str_table_t *id_info_ht,
...
 
-UNUSED __attribute__((const)) static inline bool
+UNUSED static void
+_signal_critical_error(const char *restrict format, ...) {
+    va_list va;
+    va_start(va, format);
+    char msg[512];
+    vsnprintf(msg, 512, format, va);
+    va_end(va);
+    VT_critical_error(msg, msg);
```
mandolaerik, line 776: This adds an ABI dependency requirement (user-built modules will require a recent Simics version), the reason why `_DML_fault` does its SIM_get_api_function dance is to avoid this.

This is OK (only breaks with base packages >2 years old), but please run a round of package tests to see if some compat test needs to be bumped. And if you conciously do this break, you should also clean up the DML_fault hack.

py/dml/ctree.py
```diff
@@ -1418,6 +1413,31 @@ def mkEquals(site, lh, rh):
...
+            lh = '('+lh+')'
+        if self.rh.priority <= Equals.priority:
+            rh = '('+rh+')'
+        return lh + ' == ' + rh
+
+    def read(self):
+        if self.site is None or isinstance(self.site, SimpleSite):
+            return f'_identity_eq({self.lh.read()}, {self.rh.read()})'
+        else:
+            return (f'_identity_eq_at_site({self.lh.read()}, {self.rh.read()}'
```
mandolaerik, line 1436: I'd invert the logic, if `isinstance(self.site, DumpableSite)` then we can use the operations provided by DumpableSite.

lwaern-intel: I _was_ about to say "but it's `SimpleSite` that's the outlier in not supporting `filename()` and `lineno`", but now actually I can see that it does have simplistic definitions for those. Because of that I'd rather get rid of the `isinstance` check entirely and just keep the `is None`.

include/simics/dmllib.h
lwaern-intel: This has already existed previously with `_memoized_recursion`, I just moved it up. If it didn't cause any issues then... :P

But sure, I can get rid of the `DML_fault` hack

syssimics:
Verification [#11722109](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11722109): pass

py/dml/ctree.py
```diff
@@ -1418,6 +1413,31 @@ def mkEquals(site, lh, rh):
...
+    def __init__(self, site, lh, rh): pass
+
+    def __str__(self):
+        lh = str(self.lh)
+        rh = str(self.rh)
+        if self.lh.priority <= Equals.priority:
+            lh = '('+lh+')'
+        if self.rh.priority <= Equals.priority:
+            rh = '('+rh+')'
+        return lh + ' == ' + rh
```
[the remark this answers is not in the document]
lwaern-intel: Ehh... there's no real benefit to using f-strings here other than doctrine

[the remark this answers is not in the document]
lwaern-intel: @mandolaerik Do you insist upon this?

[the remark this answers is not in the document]
mandolaerik: The style errors (space around `+`) need fixing, there is no f-string doctrine AFAIK, it's a matter of taste so if you prefer to fix by space then that's fine. (IIRC f-strings are often faster, but not significantly)

syssimics:
Verification [#11832114](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11832114): pass

PR #190, opened 2023-06-15
Test distributed API files only

mandolaerik:
Unclear if this makes a difference in practice, but makes logic clearer


syssimics:
Verification [#11726503](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11726503): pass

PR #191, opened 2023-06-16
Remove use of simicsutils.internal.is_same_file

jhbaarnh:
It is being removed in 7.


py/port_dml.py
```diff
@@ -14,9 +14,13 @@
...
 from pathlib import Path
 import ply.lex
-from simicsutils.internal import is_same_file
 from simicsutils.host import host_type
 
+def is_same_file(p1, p2):
+    return (not (os.path.exists(p1) or os.path.exists(p2))
+            or ((os.path.exists(p1) and os.path.exists(p2)
+                 and os.path.samefile(p1, p2))))
+
```
mandolaerik, line 23: if you look at the usage, both files are expected to be present, so better to replace usage with samefile right away instead.

jhbaarnh: @mandolaerik Fixed

mandolaerik (approved):

syssimics:
Verification [#11727694](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11727694): pass

PR #192, opened 2023-06-28
Warn if log groups are used as log level, or vice versa -- SIMICS-21138

lwaern-intel:
Depends on #186 

The verification of this is gonna be interesting. It will complain about what https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/4184 addresses, but what else?

### DIFF
--- a/RELEASENOTES.docu
+++ b/RELEASENOTES.docu
@@ -1724,4 +1724,10 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
   <build-id value="6244"><add-note> Added an error if a device declares
       more than 63 log groups (including the two built-in log groups.)
   </add-note></build-id>
+  <build-id value="next"><add-note> Added a warning for if a specified log
+      level of a log statement is likely intended to instead specify the log
+      groups, and/or vice versa <bug number="22018374443"/>.
+      This warning is only enabled by default with Simics API version 7 or
+      above. With version 6 and below it must be explicitly enabled by passing
+      <tt>--warn=WLOGMIXUP</tt> to DMLC.</add-note></build-id>
 </rn>
--- a/py/dml/dmlc.py
+++ b/py/dml/dmlc.py
@@ -481,6 +481,25 @@ def main(argv):
         else:
             defs[name] = value
 
+    if options.api_version not in api_versions():
+        prerr("dmlc: the version '%s' is not a valid API version" % (
+                options.api_version))
+        sys.exit(1)
+
+    if options.full_module and options.api_version not in ['4.8']:
+        prerr("dmlc: the -m option is only valid together with --api=4.8"
+              " or older")
+        sys.exit(1)
+
+    # This warning is disabled by default below Simics 7 due to sheer
+    # prominence of the issue it warns about in existing code.
+    # By only enabling the warning in Simics 7, we allow existing codebases
+    # to handle the bugs as part of migration, instead of suddenly
+    # overwhelming them with a truly massive amount of warnings in an
+    # intermediate release.
+    if options.api_version in {'4.8', '5', '6'}:
+        ignore_warning('WLOGMIXUP')
+
     for w in options.disabled_warnings:
         if not is_warning_tag(w):
             prerr("dmlc: the tag '%s' is not a valid warning tag" % w)
@@ -527,16 +546,6 @@ def main(argv):
               + "form. They are COMPLETELY UNSUPPORTED. "
               + "The DMLC developers WILL NOT respect their use. "
               + "NEVER enable this flag for any kind of production code!!!***")
-
-    if options.api_version not in api_versions():
-        prerr("dmlc: the version '%s' is not a valid API version" % (
-                options.api_version))
-        sys.exit(1)
-
-    if options.full_module and options.api_version not in ['4.8']:
-        prerr("dmlc: the -m option is only valid together with --api=4.8"
-              " or older")
-        sys.exit(1)
 
     dml.globals.api_version = options.api_version
 
--- a/py/dml/messages.py
+++ b/py/dml/messages.py
@@ -1895,6 +1895,80 @@ class WPCAST(DMLWarning):
                       if maybe_intended else '')
         DMLWarning.__init__(self, site, old, new, suggestion)
 
+class WLOGMIXUP(DMLWarning):
+    """
+    A specified log level of a `log` looks as though you meant to specify the
+    log groups instead, and/or vice versa. For example:
+    ```
+    // Log group used as log level, when the intention is instead to
+    // specify log groups and implicitly use log level 1
+    log spec_viol, some_log_group: ...;
+
+    // Log groups and log level mistakenly specified in reverse order
+    log info, (some_log_group | another_log_group), 2: ...;
+
+    // Log level used as log groups, when the intention is instead to
+    // specify the subsequent log level
+    log info, 2, 3: ...;
+    ```
+    If you want to specify log groups, make sure to (explicitly) specify the
+    log level beforehand. If you want to specify the subsequent log level, use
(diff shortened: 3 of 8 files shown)


lwaern-intel:
Oh there's already plenty... yeah we should have implemented these sooner.

syssimics:
Verification [#11785943](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11785943): fail

lwaern-intel:
The warnings are now only enabled by default on Simics 7.

I will run a VP pretest against this to check if the `as_int` use reveals an existing bug like it did for `trainings`.

syssimics:
Verification [#12157070](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12157070): fail

RELEASENOTES.docu
```diff
@@ -1724,4 +1724,11 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
   <build-id value="next"><add-note> Added an error if a device declares
       more than 63 log groups (including the two built-in log groups.)
   </add-note></build-id>
+  <build-id value="next"><add-note> Added a pair of warnings for if the
+      specified log level of a log statement is likely intended to instead
+      specify the log groups (<tt>WLGROUPSASLEV</tt>), or vice versa
+      (<tt>WLLEVASGROUPS</tt>) <bug number="22018374443"/>.
+      These warnings will only be enabled by default in Simics 7. In Simics 6
+      and below they must be explicitly enabled through the use of the
+      <tt>--warn</tt> flag. </add-note></build-id>
```
mandolaerik, line 1733: We can consider merging the warnings to one? I doubt that anyone ever will only want one of them, and easier for the user to pass a single warning flag (and easier for us to explain it: "..through the use of the <tt>--warn=WLOGGROUP</tt> flag"). We can still parameterize the warning to give descriptive messages in both cases.

mandolaerik: Technically, it's enabled by default with SIMICS_API=7. But your wording is sufficient for a releasenote.

py/dml/messages.py
```diff
@@ -1895,6 +1895,44 @@ def __init__(self, site, old, new, maybe_intended):
                       if maybe_intended else '')
         DMLWarning.__init__(self, site, old, new, suggestion)
 
+class WLGROUPSASLEV(DMLWarning):
+    """
+    The specified log level of a `log` looks as though you meant to specify the
+    log groups instead. For example:
+    ```
+    log info, some_log_group: ...;
+    log info, (some_log_group | another_log_group), 1: ...;
```
mandolaerik, line 1904: Note that we generally discourage `log info` on level 1 in production code (it's good for printf debugging only). So better to use e.g. `spec_viol` for the level 1 cases.

py/dml/messages.py
```diff
@@ -1895,6 +1895,44 @@ def __init__(self, site, old, new, maybe_intended):
...
+    explicitly specify the log level, and specify it first.
+    ```
+    log info, 1, some_log_group: ...;
+    log info, 1, (some_log_group | another_log_group): ...;
+    ```
+    """
+    fmt = ("the specified log level is likely intended to instead be a log "
+           + "groups specification. "
+           + "In order to specify log groups, the log level must be "
+           + "(explicitly) specified beforehand.")
```
mandolaerik, line 1916: You should perhaps mention "disabled by default when SIMICS_API <= 6"

py/dml/dmlc.py
```diff
@@ -481,6 +481,26 @@ def show_illegal_attributes(option, opt, value, parser):
...
+              " or older")
+        sys.exit(1)
+
+    # This pair of warnings are disabled by default below Simics 7 due to sheer
+    # prominence of the issues they warn about in existing code.
+    # By only enabling the warnings in Simics 7, we allow existing codebases
+    # to handle the bugs as part of migration, instead of suddenly
+    # overwhelming them with a truly massive amount of warnings in an
+    # intermediary release.
+    if options.api_version < '7':
```
mandolaerik, line 500: I prefer to phrase it as `not in {'4.8', '5', '6'}`; both for subtle subjective reasons and because `'10'` < `'7'`.

lwaern-intel: Oh yeah, true. This is not the only place in DMLC we do checks like this -- we should do a once-over.

py/dml/messages.py
lwaern-intel: In the docstring, that is? Having it be in the error message makes no sense.

RELEASENOTES.docu
lwaern-intel: True. How about `WLLEVGROUPSMIXUP` as the unified warning? Or maybe just `WLOGMIXUP`?

py/dml/messages.py
```diff
@@ -1895,43 +1895,79 @@ def __init__(self, site, old, new, maybe_intended):
...
-    fmt = ("the specified log groups are likely intended to instead be a log "
-           + "level specification. "
-           + "If so, reorder the specifications: the log level must be "
-           + "specified before the log groups.")
+
+    This warning is only enabled by default with Simics API version 7 or above.
+    """
+    fmt = ("log statement with likely misspecified log level(s) and log "
+           + "groups: %s")
+    def __init__(self, site, kind, level, later_level, groups):
```
lwaern-intel, line 1927: What a mess. Chime in if you think this is too complex.

test/tests.py
```diff
@@ -1070,30 +1070,22 @@ def run_cc(self, cc_extraargs):
...
-        join(testdir, "1.4", "errors", "WLLEVASGROUPS.dml"),
-        dmlc_extraargs = ["--warn=WLLEVASGROUPS"]))
+        ["1.4", "errors", "WLOGMIXUP"],
+        join(testdir, "1.4", "errors", "WLOGMIXUP.dml"),
+        dmlc_extraargs = ["--warn=WLOGMIXUP"]))
 
 if get_simics_major() == "7":
     all_tests.append(CTestCase(
-        ["1.4", "errors", "T_ETYPE_integer_t"],
+        ["1.4", "errors", "ETYPE_integer_t"],
```
lwaern-intel, line 1083: The `T_` prefix in subtest names are just redundant.

syssimics:
Verification [#12174402](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12174402): fail

syssimics:
Verification [#12211754](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12211754): fail

syssimics:
Verification [#12221962](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12221962): fail

syssimics:
Verification [#12222398](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12222398): pass

mandolaerik (approved):

test/tests.py
```diff
@@ -1070,15 +1070,23 @@ def run_cc(self, cc_extraargs):
...
         join(testdir, "1.4", "errors", "ETYPE_integer_t.dml"),
         api_version="7"))
+    all_tests.append(CTestCase(
+        ["1.4", "errors", "WLOGMIXUP"],
+        join(testdir, "1.4", "errors", "WLOGMIXUP.dml"),
+        api_version="7"))
```
mandolaerik, line 1089: This is fine, but one alternative to consider is `/// DMLC-FLAG --api-version=7` in the test. Deserves an explanatory comment too ("only enabled by default in 7 and newer"): in a couple of years, api=7 will indicate it tests some legacy behaviour.

lwaern-intel: `--api-version=7` won't be accepted in simics-base/6, so that's not an option.

PR #194, opened 2023-07-03
Avoid deprecated option forms

mandolaerik:
Not supported with Simics 7


syssimics:
Verification [#11813361](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11813361): pass

PR #196, opened 2023-07-04
Avoid internal function simics_root_path

mandolaerik:
simics_base_path() is more well-defined


syssimics:
Verification [#11821435](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=11821435): pass

PR #199, opened 2023-08-25
Run the DML tests from GitHub Action intead of TeamCity build

andreashedstrom-intel:


.github/workflows/build-and-test.sh
```diff
@@ -0,0 +1,30 @@
...
+pushd modules/
+ln -s ../.. dmlc
+popd
+
+# Local build of dmlc
+CPUS=$(cat /proc/cpuinfo | grep ^processor | wc -l)
+make dmlc -j $CPUS || exit $?
+
+# Run the tests
+T126_JOBS=$CPUS DMLC_CC=/usr/itm/gcc/13.1.0/bin/gcc ./bin/test-runner --moduledirs modules/dmlc --suite modules/dmlc/test || exit $?
```
mandolaerik, line 30: Some paths are hard-coded to our env. Would it make sense to start with some of the paths overrideable, like
```
if [ -n "$(DMLC_CC)" ] ; then export DMLC_CC=/usr/itm/... ; fi
```
... and probably similar for `PACKAGE_REPO`?
We can do the same later on request, of course.

andreashedstrom-intel: Yes, this is just a quick hack to get the box moved from TC into GH ; we`re still running (or I was assuming we would still be running) on our local runners where we have simicsmp adn /usr/itm/gcc . 

If we want to run outside on generic runners, we must somehow package up a docker image (or an image diff that you apply to some standard ubuntu image) that provides a decent GCC and Simics Base. I don't know how to do that, so that will have to be done and resolved later.

andreashedstrom-intel:
and there's a 3rd option:
3) We run DML from ci-tools repo. The ci-tools repo is internal and already runs on our local self-hosted runners. So the flow, with hard-coded ITM paths, will just work out-of-the-box. If we create a workflow_dispatch triggered workflow, then you as a user on the inside can trigger it using the `gh workflow run XXX` syntax, similar to what you now do (invoking the verify-dml-pr.sh script from ci-tools).

PR #200, opened 2023-08-30
Read the previous register value before a partial inquiry write

mandolaerik:
Fixes SIMICS-21382


syssimics:
Verification [#12107533](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12107533): pass

lib/1.4/dml-builtins.dml
```diff
@@ -2157,7 +2157,8 @@ template bank is (object, shown_desc) {
             (start, end) = this._intersect(
                 offset, size, hits[r].offset, hits[r]._size(), true);
 
-            local uint64 r_value = 0;
+            local uint64 r_value = (r_start == 0 && r_end == hits[r]._size())
+                ? 0 : hits[r].get();
```
lwaern-intel, line 2161: We sure we want `.get()` here rather than `.get_val()`? Are we ok with the semantics that an inquiry set may incur an inquiry get -- that that's what people would expect?
I agree with the choice though. If we pick `.get_val()` we risk people overriding `get()` with their own super special semantics, and then that _silently_ won't get reflected in a partial inquiry set, leading to confusion.

test/1.4/registers/T_inquiry.py
```diff
@@ -2,45 +2,37 @@
...
+def test(offs, data, expected_r, expected_overflow=bbbb, get_called=True):
+    obj.bank.b.r = aaaa
+    obj.bank.b.overflow = bbbb
+    r = Register_LE(obj.bank.b, offs, size=len(data))
+    r.read_transaction.inquiry = True
+    r.write_transaction.inquiry = True
+    val = int.from_bytes(data, 'little')
+    r.write(val)
+    expect_true(obj.set_called)
+    expect_equal(obj.get_called, get_called)
```
lwaern-intel, line 21: Using saved variables for introspection is naughty, but is fine _specifically_ in a DMLC test. It's not something we want to encourage elsewhere, though.

test/1.4/registers/T_inquiry.py
```diff
@@ -2,45 +2,37 @@
...
+    expect_equal(r.read(), val)
+    expect_false(obj.set_called)
+    expect_true(obj.get_called)
+    obj.get_called = False
+
+test(0, b'\x11',     0xaaaaaaaaaaaaaa11)
+test(7, b'\x11',     0x11aaaaaaaaaaaaaa)
+test(1, b'\x11\x22', 0xaaaaaaaaaa2211aa)
+test(7, b'\x11\x22', 0x11aaaaaaaaaaaaaa, 0xbbbbbbbbbbbbbb22)
+test(7, b'\x11\x22', 0x11aaaaaaaaaaaaaa, 0xbbbbbbbbbbbbbb22)
```
lwaern-intel, line 34: Two identical invocations of `test`.

lwaern-intel (changes requested):
Looks good, expect for a duplicated line in T_inquiry.py for which you probably meant to test something different.

test/1.4/registers/T_inquiry.py
mandolaerik: thanks, fixed

lwaern-intel (approved):

PR #201, opened 2023-09-04
Unrolled Device Info (draft)

JonatanWaern:
This was a req from ~1 year ago, which was used from the branch

Not exactly intended to be merged yet, mostly in a PR for testing


MagnusIntel:
@AleksandrDanilovIntel this might help you with the ARM register validation

PR #202, opened 2023-09-07
Fix ICE when serializing `long` or `size_t` -- SIMICS-21424

lwaern-intel:


py/dml/serialize.py
```diff
@@ -118,28 +118,27 @@ def construct_assign_apply(funname, intype):
                                  [current_expr], attr_value_t)
         return ctree.mkAssignStatement(current_site, target_expr,
                                        ctree.ExpressionInitializer(apply_expr))
-    if isinstance(real_type, TInt):
-        if real_type.signed:
-            return construct_assign_apply("SIM_make_attr_int64", real_type)
-        else:
-            return construct_assign_apply("SIM_make_attr_uint64", real_type)
-    if isinstance(real_type, TEndianInt):
```
lwaern-intel, line 126: Turns out `deserialize` already _does_ do this properly! It's just `serialize` that is bad.

lwaern-intel:
The only suspect part of this is `type_signature`:
```
    if isinstance(dmltype, IntegerType):
        return 'I%d%s%s' % (dmltype.bits,
                            dmltype.byte_order[0] if dmltype.is_endian else "h",
                            "s" if dmltype.signed else "u")
```
We may want to add dedicated cases for `long` and `size_t`: `IL` and `IS`.

lwaern-intel:
> The only suspect part of this is `type_signature`:
> 
> ```
>     if isinstance(dmltype, IntegerType):
>         return 'I%d%s%s' % (dmltype.bits,
>                             dmltype.byte_order[0] if dmltype.is_endian else "h",
>                             "s" if dmltype.signed else "u")
> ```
> 
> We may want to add dedicated cases for `long` and `size_t`: `IL`(`s`) and `IS`(`s`).

We do. Our assumption about the bit widths of `long` and `size_t` feel very insufficient to me: conflating serialization of `int64` and `long` on Windows -- and having the generated de/serialization function for one be used for the other -- risks loading and storing `long`s as `int64`s (and vice versa), which is certainly UB, and may very well be _dangerous_ UB that leads to stack/heap corruption.

syssimics:
Verification [#12147475](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12147475): fail

syssimics:
Verification [#12147878](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12147878): fail

lwaern-intel:
GH is being mean to the TC job and us as a result :(

eventually it'll allow it to run

syssimics:
Verification [#12148332](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12148332): pass

mandolaerik:
> risks loading and storing `long`s as `int64`/`int32`s (and vice versa), which is certainly UB,

One option might be a memcpy indirection, so that we technically serialize the byte representations of all integers that may share serializers with `size_t`/`long`. But your solution is clearer, and only at the cost of minor duplication, most of which GCC -O2 probably will be able to collapse.

mandolaerik (approved):

PR #203, opened 2023-09-07
Add script to find globally dead DML methods

mandolaerik:
- Correct some line directives, to avoid noise in coverage reports
- Add script to find globally dead DML methods

### DIFF
--- a/py/dml/c_backend.py
+++ b/py/dml/c_backend.py
@@ -1657,7 +1657,7 @@ def generate_deinit(device):
                 # any exceptions. But we don't want to force them to insert
                 # try-catch in the init method.
                 with LogFailure(device.site, event, indices):
-                    codegen_inline(device.site, method, indices, [], []).toc()
+                    codegen_inline(method.site, method, indices, [], []).toc()
             for i in range(len(dims)):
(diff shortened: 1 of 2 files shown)


mandolaerik:
Wrote this to cover Kalen's use cases with coverage on dead code. The idea would be to deploy this script (or similar) as a postprocessing step of CI after *all* devices have been compiled, and use that (in conjunction with filters) to make sure no model contains dead code. That way, we can know that any method that is still locally unused in a device, is in fact a library method used by someone else, and thus included in the aggregated coverage report.

Filters will have to be applied (both for coverage and for the dead-method scan) to avoid complaints on DML code from outside your repo; it is up to the respective code owner to ensure coverage for such code.

mandolaerik:
I have tested the script manually on a few devices (sample-pci-device, sample-pcie-sriov-device), and it seems to work OK. What remains is:
* ensure this is actually wanted by stakeholders
* write proper tests for the script
* try the script on a larger scale, e.g. SSC git or the DML test suite. Both to smoke out problems and to dogfood the tool, there are probably dead methods out there that we can clean out.

MagnusIntel (approved):

PR #205, opened 2023-09-12
Add _compat_proxy_port_ifaces param

mandolaerik:
- Factor out params for API version ranges
- Add _compat_proxy_port_ifaces param
- Rename function
- Refactor: Re-use common function
- Add _compat_proxy_port_ifaces param


mandolaerik:
Will add tests and releasenote.

syssimics:
Verification [#12174833](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12174833): fail

PR #206, opened 2023-09-13
Update docs for signal_port and signal_connect

MagnusIntel:
In the documentation, these were described in the midst of "Templates for reset"

### DIFF
--- a/lib/1.4/utility.dml
+++ b/lib/1.4/utility.dml
@@ -196,108 +196,6 @@ template _init_val_power_on_reset is (init_val, power_on_reset) {
     }
 }
 
-/**
-### signal_port
-
-<section>
-
-#### Description
-
-Implements a signal interface with saved state. The current state of the signal
-is stored in the saved boolean `high`, and a spec-violation message is logged
-on level 2 if the signal is raised or lowered when already high or low. The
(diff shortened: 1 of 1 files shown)


mandolaerik:
We have a section [5.4 Connect related templates](https://simics-download.pdx.intel.com/simics-6/docs/html/dml-1.4-reference-manual/utility.html#connect-related-templates), I suppose we should move the `signal_connect` template there, or alternatively rethink the categories.

Docs are extracted in order, so moving around docs requires moving around the corresponding code, which is OK.

MagnusIntel:
> We have a section [5.4 Connect related templates](https://simics-download.pdx.intel.com/simics-6/docs/html/dml-1.4-reference-manual/utility.html#connect-related-templates), I suppose we should move the `signal_connect` template there, or alternatively rethink the categories.
> 
> Docs are extracted in order, so moving around docs requires moving around the corresponding code, which is OK.

Can I put signal_port under "Connect related templates" as well? It fits better there than the current place in any case

mandolaerik:
Having a separate one for signals makes sense, but perhaps signal should come before reset (which is based on signal)?

MagnusIntel:
Can we just get this in and then debate new sections etc? The current state is awful and this PR should be a distinct improvement

PR #208, opened 2023-09-21
Migrate from optparse to argparse

mandolaerik:
- Shut up linter
- Remove the --illegal-attrs flag
- Remove --illegal-attrs flag
- Remove --version, which was ignored
- Migrate from optparse to argparse

### DIFF
--- a/py/dml/dmlc.py
+++ b/py/dml/dmlc.py
@@ -2,14 +2,14 @@
 # SPDX-License-Identifier: MPL-2.0
 
 import sys, os, traceback, re
-import optparse
+import argparse
 import tempfile
 import shutil
 import time
 from pathlib import Path
 
 from . import structure, logging, messages, ctree, ast, expr_util, toplevel
-from . import codegen, serialize
+from . import serialize
 from . import dmlparse
 from . import output
 
@@ -265,9 +265,7 @@ def main(argv):
     # DML files must be utf8, but are generally opened without specifying
     # the 'encoding' arg. This works only if utf8_mode is enabled.
     assert sys.flags.utf8_mode
-    optpar = optparse.OptionParser(
-        """
-  dmlc [flags] <file> [output-base]""")
+    parser = argparse.ArgumentParser()
 
     # <add id="dmlc options">
     # <name>Command Line Options</name>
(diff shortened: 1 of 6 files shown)


syssimics:
Verification [#12224557](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12224557): pass

syssimics:
Verification [#12227474](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12227474): pass

py/dml/dmlc.py
```diff
@@ -281,102 +300,98 @@ def main(argv):
...
+        help='generate makefile dependencies')
 
     # <dt>--no-dep-phony</dt>
     # <dd>With --dep, avoid addition of a phony target for each dependency
     # other than the main file.</dd>
-    optpar.add_option(
-        '--no-dep-phony', dest = "no_dep_phony", action = 'store_true',
-        help = 'With --dep, avoid addition of a phony target for each'
+    parser.add_argument(
+        '--no-dep-phony', action='store_true',
```
lwaern-intel, line 330: So with no explicit `dest` the parser will parse `--no-dep-phony`, and turn it into `no_dep_phony` when genning the options? Huh.

lwaern-intel (approved):
Undeniably uncontroversial

py/dml/messages.py
```diff
@@ -2392,3 +2392,7 @@ class PINT1(PortingMessage):
     types, then the value of the variable becomes 1, whereas for
     `int1` in DML 1.4 the value is -1."""
     fmt = "Change int1 to uint1"
+
+warnings = {name: cls for (name, cls) in globals().items()
```
lwaern-intel, line 2396: Just to double-check that I remember the semantics of `globals()` -- does it return the accessible vars _defined in the current module only_, or does it return accessible vars accessible at the top-level of the current module, no matter where they are defined?

mandolaerik: Somewhere inbetween, it seems -- globals() returns a dict that includes some builtin symbols like `__name__`,  `__loader__` and `__builtins__`, but not others like `__import__`.

py/dml/dmlc.py
mandolaerik: Yes, that's standard practice for argparse. Adn `add_argument('-f', '--foo-bar', ...)` stores it in `foo_bar`.

syssimics:
Verification [#12247784](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12247784): pass

PR #204, opened 2023-09-08
Soothe static analysis on Windows

lwaern-intel:
See [this](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12154207). Hoping this will fix it, will make a Test W run before merging.

PR #210, opened 2023-09-22
Trait trampolines: declare `_flat_index` as uint32

lwaern-intel:
This is to silence GCC when it decides to make warnings based on bad assumptions about the values that an argument may take.

Or at least, that is what I think is going on. Original issue was reported by @gwikland. I have not been able to create a minimal reproducer, which is why this lacks a test.

The seemingly stochastic nature of the issue _does_ terrify me -- making me doubt myself and wondering if the warning is actually based on a concrete usage rather than GCC assuming something to be possible. However, based on an inspection of the generated code, experimentation, and my previous experience of a similarly stupid error (see #204) I've thoroughly convinced myself that no, GCC really _is_ just deciding to be dumb.

### DIFF
--- a/py/dml/c_backend.py
+++ b/py/dml/c_backend.py
@@ -2703,11 +2703,11 @@ def generate_trait_trampoline(method, vtable_trait):
         site = method.site
         obj = method.parent
         if obj.dimensions:
-            out(f'int _flat_index = {tname}.id.encoded_index;\n')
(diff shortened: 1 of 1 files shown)


mandolaerik (approved):

syssimics:
Verification [#12228952](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12228952): pass

PR #212, opened 2023-09-28
Add porting rule for `assert 0;`

mandolaerik:
This tends to be fairly common in DML 1.2 code

### DIFF
--- a/test/1.2/misc/porting.dml
+++ b/test/1.2/misc/porting.dml
@@ -393,6 +393,8 @@ event ev2 is (evt) {
     parameter timebase = "seconds";
 }
 
+parameter zero = 0;
+
 method init() {
     local int1 i1;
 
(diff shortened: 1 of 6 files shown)


syssimics:
Verification [#12261190](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12261190): pass

test/1.2/misc/porting.dml
```diff
@@ -428,4 +430,9 @@ method init() {
             $c.signal.signal_raise();
         }
     }
+    if (1) {
+        assert $zero;
```
lwaern-intel, line 434: Since `PZEROCOND` covers both `assert 0;` and `assert 1;`, test `assert 1;` too. Also, because of that, you may want to consider renaming it to `PILITCOND` -- as in, Integer Literal Condition.

PR #213, opened 2023-09-28
RAII Types and more

lwaern-intel:
Besides the obvious (RAII architecture, vectors, strings), this includes:
1. `cast(initializer, ...)` syntax, and rework of associated logic
3. As a result of the above, compound literals. `cast({1,2,3}, int[3])` is now supported.
4. `a += initializer` syntax, and rework of associated logic. This means complex method calls can be used as RHS of `+=`/`-=`/etc.
5. Reworked statements' `toc`/`toc_inline` family
6. ExpressionInitializer.assign_to and Expression.write rework
7. The addition of "addressable" and "c_lval" to Expression, and an associated rework of lvalue-ness
8. Strict `DMLType.cmp`. Old implementation is still used in a few places and is called `cmp_fuzzy`
9. The ability to hash DML Types in a way compatible with the new strict cmp.
10. `extern cident(blabla) as dmlident;`.
11. Improved deserialization logic for arrays; doesn't need a stack frame per dimension

This is a draft PR as it is _way_ too massive to be merged in as-is. We should decide upon what we should split and how.
In addition, the tests in this PR are too basic -- they are touch-ups of the ad-hoc tests I made while working on this. They are not sufficient to e.g. exhaustively test the semantics of strings and vectors -- vectors in particular have some very complicated logic attached to them.

You'll spot a number of `TODO(RAII)`s in the implementation -- these represent particular question marks where I feel the implementation could potentially be improved, rather than unimplemented components -- this PR is feature complete.

### DIFF
--- a/include/simics/dml-raii-types.h
+++ b/include/simics/dml-raii-types.h
@@ -103,11 +103,12 @@ _dml_string_realloc_for_len(_dml_string_t *s, uint32 new_len) {
 UNUSED static void
 _dml_string_resize(_dml_string_t *s, uint32 new_len) {
     uint32 prev_len = s->len;
+    uint32 prev_size = s->size;
     if (prev_len == new_len) return;
     _dml_string_realloc_for_len(s, new_len);
     s->len = new_len;
     if (s->size) {
-        if (prev_len > new_len) {
+        if (prev_size == 0 || prev_len > new_len) {
             s->s[new_len] = '\0';
         } else {
             memset(s->s + prev_len + 1, 0, new_len - prev_len);
@@ -640,7 +641,7 @@ _dml_vect_realloc(size_t elem_size, _dml_vect_t *v, uint32 new_size) {
     if (new_size > v->size) {
         ASSERT(new_size >= v->start + v->len);
         v->elements = MM_REALLOC_SZ(v->elements, new_size*elem_size, char);
-        if (v->start + v->len > v->size)
+        if (v->size && v->start + v->len > v->size)
             memcpy(v->elements + v->size*elem_size,
                    v->elements,
                    (v->len - (v->size - v->start))*elem_size);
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,2 +1,3 @@
 *~
 *.pyc
+flycheck_*
```
lwaern-intel, line 3: This addition was because I got _real_ tired of my `rg` taking me into files fisketur flycheck created, and adding a `flycheck_*` filter was the easiest way to avoid that. We may or may not want this in.

test/1.4/syntax/T_assign.dml
```diff
@@ -6,6 +6,16 @@ dml 1.4;
...
 
+method m(int i) -> (int) throws {
+    if (i > 4) throw;
+    return i;
+}
+
+method ms(int i) -> (string) throws {
+    if (i > 4) throw;
+    return mk_string_f("%d", i);
+}
```
lwaern-intel, line 17: Artifact, I had moved this method to `1.4/types/raii/T_various`.

syssimics:
Verification [#12261598](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12261598): fail

py/dml/expr.py
```diff
@@ -292,6 +335,32 @@ def copy(self, site):
 
 mkNullConstant = NullConstant
 
+class Orphan(Expression):
+    """Expressions that evaluate to a value that is allocated on the stack, but
+    not belonging to any local variable. Archetypical example are function
+    applications."""
```
lwaern-intel, line 341: This description was written in a better time when orphans were directly tied to C concepts. What they are nowadays are not so simple (`Expression.orphan` has the better description)

lwaern-intel:
Test W run: https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=12292208

lwaern-intel:
Verification run featuring incompatibility patch (https://github.com/intel/device-modeling-language/pull/217): https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=12292226

lwaern-intel:
Verification without incompatibility: https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=12292197

PR #215, opened 2023-09-28
Validate that index args are within bounds.

mandolaerik:
- Avoid wrapping constants in multiple layers of InlineParam
- Validate that index args are within bounds.


mandolaerik:
Apparently I opened #219 as a continuation of this one, don't know why. Closing.

PR #218, opened 2023-10-04
Introduce `Site.linemarked` to suppress line directives for non-method C code

lwaern-intel:
This is still somewhat of a proof-of-concept -- I haven't looked too closely at where we want to apply this outside of c_backend.


MagnusIntel:
I discussed this with Erik today and he thought it was a bit of work, but then realized that the only place we should put line markers is within methods ... which supposedly made the problem much easier.

syssimics:
Verification [#12294889](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12294889): pass

mandolaerik:
I like #221 much better, I hope we can close this one

PR #219, opened 2023-10-05
Add bound checks for method indices

mandolaerik:


mandolaerik:
No tests yet, and I'm not sure this is the right approach. Opens:
* Need a cost analysis: how much does this bloat generated code?
  * In particular: I tried to design a fallback behaviour that minimizes the cost in code bloat, but at the cost of functionality (clearing *all* indices to zero is cheap but more rude than strictly necessary). Need some real numbers to assess if it's worth it.
* With this patch, the callee checks the bound; should it be the caller? That would give a better error message, but probably also more code bloat.
* How well does this generalize to other bound checks? e.g., accessing params or session variables in object arrays, or creating traitrefs to object arrays.

PR #221, opened 2023-10-05
Selectively allow generation of line directives

lwaern-intel:
An alternate (and by all indications superior) approach to the problem #218 solves.

Closes #218


py/dml/c_backend.py
```diff
@@ -1870,23 +1855,23 @@ def generate_init_data_objs(device):
...
+                                node.site, node._type, node.astinit,
+                                Location(node.parent, index_exprs),
+                                global_scope, True)
+                        except DMLError as e:
+                            report(e)
+                        else:
+                            markers = ([('store_writes_const_field', 'FALSE')]
+                                       if deep_const(node._type) else [])
+                            coverity_markers(markers, node.site)
+                            init.assign_to(nref, node._type)
```
mandolaerik, line 1874: why linemark initializers? they are not interesting while debugging, and just add noise to coverage.

If you need to enable linemarks just for coverity, then I suppose we can change `coverity_markers` to go an easy path if linemarks are disabled?

py/dml/expr.py
```diff
@@ -80,8 +83,28 @@ def coverity_markers(markers, site=None):
...
+# and the current output is FileOutput
+# This context manager generates a line directive reset when left unless
+# linemarks were already allowed, or manual_reset is True
+@contextmanager
+def allow_linemarks(manual_reset=False):
+    prev_linemarks = dml.globals.linemarks
+    curr_output = output.current()
+    is_file_output = isinstance(curr_output, FileOutput)
+    assert is_file_output >= prev_linemarks
+    assert dml.globals.linemarks_enabled >= prev_linemarks
```
mandolaerik, line 96: can you plese express this in first order logic instead?

lwaern-intel: implications are first-order logic! It's not my fault that "X implies Y" is spelled `X <= Y`!

py/dml/expr.py
```diff
@@ -80,8 +83,28 @@ def coverity_markers(markers, site=None):
         site_linemark_nocoverity(site)
 
 
+# Allow linemarks to be generated if dml.globals.linemarks_enabled is True
+# and the current output is FileOutput
+# This context manager generates a line directive reset when left unless
+# linemarks were already allowed, or manual_reset is True
+@contextmanager
+def allow_linemarks(manual_reset=False):
```
mandolaerik, line 91: Can we find a better word than `allow` here? to me, 'allow' something indicates that it's an error to do that otherwise. I can understand your reasoning that 'if linemarks are enabled, they will be output unless disallowed", but it's really a symmetric conjunction (linemarks happen whenever they are both "enabled" and "allowed"), so you could just as well swap the two words.

(though I realize I'm starting a discussion on the naming of an internal identifier, so feel free to close the thread with no comment)

py/dml/c_backend.py
lwaern-intel: 1. I don't need linemarks for coverity support; and 2. with this PR `coverity_markers` has such a path.

I could just as easily remove the `allow_linemarkers` and this will work. Like I said in R.C., linemarking initializers is a question if we want it for _other_ analysis tools than gcov or gdb. I mentioned Coverity as an example because if coverity finds issue with a session/saved initializer, then we would rather it point to the DML declaration, wouldn't we?

py/dml/expr.py
lwaern-intel: i have no idea how to best express this concept. I found "allow" and decided that were good enough.

syssimics:
Verification [#12298426](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12298426): pass

py/dml/ctree.py
```diff
@@ -475,16 +475,26 @@ def toc(self):
...
+        with allow_linemarks():
+            for stmt in self.body:
+                stmt.toc()
+        for stmt in self.post:
+            stmt.toc()
     def control_flow(self):
         return ControlFlow(fallthrough=True, throw=self.method.throws)
 
+# TODO: should this rewrite `body=[Compound(statements)]` to `body=statements`?
 mkInlinedMethod = InlinedMethod
```
lwaern-intel, line 498: Yeah it should, as long `statements` as it doesn't have any declaration. Previously `InlinedMethod` was just a wrapper around a `mkCompound(_; pre + body + post)`, and `mkCompound` collapses any `Compound` of the list is passed if it has no declaration.

py/dml/c_backend.py
```diff
@@ -3357,7 +3345,6 @@ def generate_cfile_body(device, footers, full_module, filename_prefix):
                 raise ICE(e.site, 'error during late compile stage')
             eom_linemark(func.method.rbrace_site)
             out('}\n\n', preindent = -1)
-        reset_line_directive()
```
lwaern-intel, line 3360: This does the wrong thing when `gather_size_statistics` is on. That should be solvable though.

syssimics:
Verification [#12299345](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12299345): pass

py/dml/c_backend.py
```diff
@@ -3220,6 +3206,7 @@ def splitting_point():
     assert_global_c_scope()
     if c_split_threshold and c_file.tell() > c_split_threshold:
         c_file.advance()
+        # TODO is this needed?
         reset_line_directive()
```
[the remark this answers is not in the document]
mandolaerik: AFAICS, it is redundant. A self-referencing line directive is output near the top of a newly created file. The few lines in the top of the new file includes no `#line` directive.

py/dml/expr.py
mandolaerik: I find it more readable to use conventional boolean operators, preferably on DNF or CNF.  like `is_file_output or not prev_linemarks`. And in this case, I think you can collapse both lines to `dml.globals.linemarks or not prev_linemarks`.

lwaern-intel: I've fixed this in the latest commit, forgot to resolve it.

py/dml/c_backend.py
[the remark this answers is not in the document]
lwaern-intel: Yeah. I'll be rid of it.

syssimics:
Verification [#12303125](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12303125): pass

py/dml/expr.py
```diff
@@ -80,8 +84,38 @@ def coverity_markers(markers, site=None):
...
```
mandolaerik, line 92: looks like `manual_reset` is never used, so that can be removed.

lwaern-intel: Yeah; I thought I'd need it once I began writing the PR, but nope. Now it just exists as a "just in case" but in this case I agree we can add it later if need be.

syssimics:
Verification [#12303345](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12303345): pass

syssimics:
Verification [#12316100](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12316100): fail

syssimics:
Verification [#12341554](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12341554): fail

syssimics:
Verification [#12341827](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12341827): pass

test/1.4/errors/T_ESWITCH.dml
```diff
@@ -11,12 +11,6 @@ method init() {
         /// ERROR ESWITCH
     {
     }
-    switch (1)
-        // must start with a case
-        /// ERROR ESWITCH
-    {
-        ;
-    }
```
lwaern-intel, line 19: Artifact from when the RAII parser changes were present. I thought I had unstaged this.

syssimics:
Verification [#12342942](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12342942): pass

py/dml/codegen.py
```diff
@@ -3219,12 +3225,10 @@ def stmt_switch(stmt, location, scope):
...
             assert body_ast.kind == 'compound'
-            [stmt_asts] = body_ast.args
+            [stmt_asts, rbrace_site] = body_ast.args
             stmts = codegen_statements(stmt_asts, location, scope)
-            if (not stmts
-                or not isinstance(stmts[0], (ctree.Case, ctree.Default))):
-                raise ESWITCH(
-                    body_ast.site, "statement before first case label")
+            if not stmts:
+                raise ESWITCH(body_ast.site, "empty switch statement")
```
lwaern-intel, line 3231: Oops. I'm getting rid of _this_ diff.

test/1.4/misc/linemarks.dml
```diff
@@ -0,0 +1,218 @@
...
+    // END-LINEMARKS
+}
+
+// destroy() is the only overridable method in DML 1.4 which is ad-hoc invoked
+// by DMLC via a codegen_inline_byname, and can possibly never be called
+// otherwise.
+method destroy() {
+    // BEGIN-LINEMARKS-FUNC test_deinit
+    local int x;
+    x = inline_m(x);
```
mandolaerik, line 210: we discussed elsewhere that
```
x = inline_m(
   x + 3 // this line should get a linemark
   );
```
... though it's arguably unimportant to test this, as we want to eventually retire inline methods.

lwaern-intel: Already tested elsewhere, in `init()`.

test/tests.py
```diff
@@ -1508,6 +1509,154 @@ def test(self):
...
+                        for func in curr_span[1]:
+                            func_spans.setdefault(func, set()).add(
+                                (curr_span[2], lineno - 1))
+                    else:
+                        cblock_spans[curr_span[1]] = lineno - 1
+                    curr_span = None
+                elif line == '// NOT-LINEMARKED':
+                    assert (curr_span is not None
+                            and curr_span[0] is func_spans), \
+                            ("{lineno}: // NOT-LINEMARKED used outside of "
```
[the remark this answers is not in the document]
lwaern-intel: oops

test/tests.py
```diff
@@ -1508,6 +1509,154 @@ def test(self):
...
+            with open(curr_file) as file_handle:
+                curr_cblock_expected = []
+                for (dml_lineno, c_lineno, curr_func) in self.dml_lines_iter(
+                        curr_file, file_handle.readlines()):
+                    if curr_cblock_expected:
+                        (expected_dml_lineno,
+                         expected_c_lineno) = curr_cblock_expected.pop()
+                        assert ((dml_lineno, c_lineno)
+                                == (expected_dml_lineno, expected_c_lineno)), \
+                                (f'In {curr_file}: Expected DML line '
```
mandolaerik, line 1619: I think the style guide wants us to avoid `\` line continuations, which *can* be done by movingthe `(` of the message on the previous line. This stuff is dangerous though, it's way too easy to accidentally write `assert (x, y)` instead of `assert x, y`. I don't know how to make this easy to parse, perhaps something like this?
```
assert (
    ...
), (
    ...
)
```

lwaern-intel: errgh... maybe.

I hate Python's breed of syntactically significant whitespace.

test/tests.py
```diff
@@ -1508,6 +1509,154 @@ def test(self):
...
+
+    def test(self):
+        super().test()
+        c_file = Path(f'{self.cfilename}.c').resolve()
+        h_file = Path(f'{self.cfilename}.h').resolve()
+        base_file = Path(self.filename).resolve()
+
+        func_spans = {}
+        cblock_spans = {}
+        not_linemarked = set()
```
mandolaerik, line 1571: Please add comment and/or type annotations.

test/tests.py
```diff
@@ -1508,6 +1509,154 @@ def test(self):
...
+                elif line == '// BEGIN-LINEMARKS-CBLOCK':
+                    curr_span = (cblock_spans, lineno + 1)
+                elif line == '// END-LINEMARKS':
+                    assert curr_span is not None
+                    if curr_span[0] is func_spans:
+                        for func in curr_span[1]:
+                            func_spans.setdefault(func, set()).add(
+                                (curr_span[2], lineno - 1))
+                    else:
+                        cblock_spans[curr_span[1]] = lineno - 1
```
mandolaerik, line 1589: Consider extracting this DML analysis to a function (and maybe the same for the C code analysis), to reduces the number of mutated variables simultaneously in scope.

test/tests.py
lwaern-intel: oo! I can hear Python gurus all across the world becoming _absolutely infuriated_ by just the gall of _demanding_ people declare their types ;)

I've mentioned previously how much I hate type annotations due to the half-measure they represent, as they are nothing more but glorified comments, and so have an instinctive, irrational repulsion to their use. But since the question here is the choice between comments or glorified comments, I suppose I could indulge in the latter.

test/tests.py
lwaern-intel: Good idea.

test/tests.py
mandolaerik: Hence 'and/or'. I also don't quite like how type annotations turned out, but type annotations are often more concise than a comment in expressing what you want. Expressing `func_span: dict[str, set[(int, int)]] = {}` in text is much more  verbose without really adding anything essential.

test/1.4/misc/linemarks.dml
```diff
@@ -0,0 +1,218 @@
...
+    {
+        assert true;
+    // NOT-LINEMARKED
+    }
+
+    // ... Unless it contains a declaration.
+    {
+        local int y;
+    }
+
```
mandolaerik, line 98: As discussed elsewhere, the way we use NOT-LINEMARKED to test how compound statement collapse is somewhat unfocused and smokey, and may possibly cause unnecessary disruption. Ideally it might be expressed as a ctree_test, but not today. Still, these tests are of some use so discarding them right away would be a waste -- how about moving the less important tests out to a separate method, with a comment on this status? That will make it easier to take proper action if they ever get in the way.

doc/1.4/running-dmlc.md
```diff
@@ -115,6 +115,11 @@ causes generated C code to follow the DML code more closely.
 
 Adds Synopsys® Coverity® analysis annotations to suppress common false positives
 in generated C code created from DML 1.4 device models.
+
+Analysis annotation generation impacts the generation of line directives in a
+way that may cause debugging or coverage tools besides Coverity to display
+unexpected behavior. Because of this, it's recommended that `--coverity` is
+only used when needed.
```
lwaern-intel, line 122: So this disclaimer specifically refers to this inside `coverity_markers`:
```
out('#ifndef __COVERITY__\n')
site_linemark_nocoverity(site, adjust=-(len(markers) + 1))
out('#endif\n')
```
The linemark we generate is adjusted such that the next line that the upcoming line after the `#endif` and the `/* coverity ... */ markers will be redirected to the correct DML line; however, that means that the `#endif` and the markers will be redirected to the _preceding_ DML lines. This makes no difference if you're only interested in Coverity and GCC error reporting working properly; they will never complain about the `#endif` or comments. However, GDB may perhaps consider breaking on an `#endif` or a comment (however that works out), and GCOV may consider the redirected DML lines covered.

test/tests.py
lwaern-intel: Done both.

test/1.4/misc/linemarks.dml
lwaern-intel: Done.

doc/1.4/running-dmlc.md
lwaern-intel: And no, there is no way around this save by omitting the `#ifndef` entirely, in which case the preceding line directive reset will still be in effect which would _also_ be unexpected,  and so this disclaimer would still needed save it would also need to say that GCC error reporting may display unexpected behavior.

lwaern-intel: This is why I hate Coverity analysis annotations. You hate them because their implementation necessitates pragmas. I hate them because their implementation necessitates _this._

lwaern-intel: And no, there is no superior option to analysis annotations either. Believe me, I've tried to find one.

syssimics:
Verification [#12402635](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12402635): pass

mandolaerik (approved):

PR #222, opened 2023-10-10
1.4:utility.dml: template miss_pattern_bank: Allow method overrides

gwikland:
Allow method overrides so models could for instance print out info messages when an unmapped access occurs.

### DIFF
--- a/lib/1.4/utility.dml
+++ b/lib/1.4/utility.dml
@@ -1176,9 +1176,10 @@ template function_io_memory {
 
 #### Description
 
-Only valid in `bank` objects. Handles unmapped accesses by
-ignoring write accesses, and returning a given value for each unmapped
-byte.
+Only valid in `bank` objects. Handles unmapped accesses by ignoring write
+accesses, and returning a given value for each unmapped byte. If you want to
+customize this behaviour, overriding `unmapped_get` is sufficient to also
+customize `unmapped_read`.
 
 #### Parameters
 
@@ -1190,13 +1191,14 @@ template miss_pattern_bank is bank {
     param miss_pattern : uint8;
     shared method unmapped_read(uint64 offset, uint64 bits, void *aux)
         -> (uint64) throws default {
-        return this.get(offset, bits);
+        return this.unmapped_get(offset, bits);
     }
-    shared method unmapped_get(uint64 offset, uint64 bits) -> (uint64) throws default {
+    shared method unmapped_get(uint64 offset, uint64 bits)
+        -> (uint64) throws default {
         return miss_pattern * 0x0101010101010101;
     }
     shared method unmapped_write(uint64 offset, uint64 value, uint64 bits,
-                          void *aux) throws default {
+                                 void *aux) throws default {
     }
 }
 
--- a/lib/1.4/utility.dml
+++ b/lib/1.4/utility.dml
@@ -1189,14 +1189,14 @@ miss\_pattern: each missed byte in a miss read is set to this value
 template miss_pattern_bank is bank {
     param miss_pattern : uint8;
     shared method unmapped_read(uint64 offset, uint64 bits, void *aux)
-        -> (uint64) throws {
-        return miss_pattern * 0x0101010101010101;
+        -> (uint64) throws default {
+        return this.get(offset, bits);
     }
-    shared method unmapped_get(uint64 offset, uint64 bits) -> (uint64) throws {
+    shared method unmapped_get(uint64 offset, uint64 bits) -> (uint64) throws default {
         return miss_pattern * 0x0101010101010101;
     }
     shared method unmapped_write(uint64 offset, uint64 value, uint64 bits,
-                          void *aux) throws {
+                          void *aux) throws default {
     }
 }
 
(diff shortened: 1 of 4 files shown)


lib/1.4/utility.dml
```diff
@@ -1189,14 +1189,14 @@ miss\_pattern: each missed byte in a miss read is set to this value
 template miss_pattern_bank is bank {
     param miss_pattern : uint8;
     shared method unmapped_read(uint64 offset, uint64 bits, void *aux)
-        -> (uint64) throws {
-        return miss_pattern * 0x0101010101010101;
+        -> (uint64) throws default {
+        return this.get(offset, bits);
```
lwaern-intel, line 1193:
```
return this.unmapped_get(offset, bits);
```
... right?

lwaern-intel (changes requested):

syssimics:
Verification [#12324175](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12324175): fail

lib/1.4/utility.dml
lwaern-intel: Considering how badly verification broke, probably.

lib/1.4/utility.dml
```diff
@@ -1189,14 +1189,14 @@ miss\_pattern: each missed byte in a miss read is set to this value
...
     }
-    shared method unmapped_get(uint64 offset, uint64 bits) -> (uint64) throws {
+    shared method unmapped_get(uint64 offset, uint64 bits) -> (uint64) throws default {
         return miss_pattern * 0x0101010101010101;
     }
     shared method unmapped_write(uint64 offset, uint64 value, uint64 bits,
-                          void *aux) throws {
+                          void *aux) throws default {
```
mandolaerik, line 1199: The corresponding documentation update is missing.

In particular: With your current impl, a user who overrides the return value of `unmapped_get` will also see the corresponding change from unmapped_read. If they want to rely on this, then they need that it is documented as a stable behaviour. (Whether or not your proposed default chain is desired is a separate question; I think your proposed API makes sense but my stance is not firm).

What's painful here is that we have a template with in total 2 lines of method body, and when this `default` marking is added, the template will expose an API that will require many words to describe. This is exactly why so much of our APIs have started out as non-`default`. To an end-user, It may make more sense to override unmapped_read/unmapped_get directly, without the miss_pattern_bank indirection; the only cost is really that you have to write `return 0xffffffffffffffff;` instead of `return default(offset, bits);`.

mandolaerik: Also, you need to add tests for all override points.

lwaern-intel:
Erik's demands as well as debate about the merits of this change has made the effort needed to advocate significantly greater than I expected, so I'll hijack this PR and argue on its behalf.

lib/1.4/utility.dml
lwaern-intel: > The template will expose an API that will require many words to describe.

Not really.

> If you wish to customize this behaviour, overriding `unmapped_get` is sufficient to also customize `unmapped_read`, and overriding `unmapped_set` is also sufficient to customize `unmapped_write`.

jakob-engblom-i:
I have a use case that could be helped by this. In my "transaction forwarder from Simics PCIe to other simulator probably RTL". 

The PCIe config bank in the forwarder device contains the bare minimum set of registers needed to make PCIe functional on the Simics side, basically the BARs.  No attempt is made to mirror all PCIe registers from the RTL.

For memory transactions that reach the PCIe config bank I do this:
* Send the operation to `default()` for the `transaction_access` method - this means writes to BARs have desired effects in SImics, etc. I save the return value of this call. 
* Create a new transaction that is sent to the "forwarder" device, separate object in Simics. This produces another transaction.issue return value, and for a read, comes back with some value in the transaction. 
* Next, I need to decide how to reconcile the results. 

Ideally, for an operation that misses all defined registers in the PCIe config bank, I use the result from RTL. Assumption is that the RTL has more stuff in it... but to do this I need to know if I did indeed hit "empty space" in the PCIe config bank. Right now, all locations in the PCIe config space returns Sim_PE_no_exception since something is handling the accesses. 

In case both RTL and the Simics model returned values on reads, I would like to compare them to flag inconsistencies.  

Right now, I cannot do that since I cannot tell apart hits in registers and hits in the fallback. 

I would like to know if we did hit a register defined in my model or the fallback "return 0xFF" logic that I understand to be there. Basically, either by looking at the return value from default or from some other data point in the model that tells me what happened behind the scenes.

mandolaerik:
@jakob-engblom-i can you please share some (pseudo-)code for what this looks like with vs without this PR?

jakob-engblom-i:
> @jakob-engblom-i can you please share some (pseudo-)code for what this looks like with vs without this PR?

There is simply no way to do it. 

Code such as it is is here: https://github.com/intel-innersource/applications.simulators.simics.contrib.contrib/blob/main/pci-device-forwarding-memory-operations/modules/forward-pcie-new-main/forward-pcie-new-main.dml 

What I would like to be able to do is to have a return value in some shape that can tell me if I hit or missed the actual defined registers in the bank. 
```
    method transaction_access(transaction_t *t, uint64 offset, void *aux) -> (exception_type_t) {
    ...
        local exception_type_t ex;
        ex=default(t,offset,aux);
        if(ex!=Sim_PE_No_Exception) {
            log info, 2, PCIe : "(%s) Exception from default: 0x%x", this.qname, ex;
        }
        // Check result of operation - did we hit a register or not?
        // -- Should be Sim_PE_No_Exception in case we hit a register - actually, always is
        //     due to how missing registers is implemented in PCIe
```

syssimics:
Verification [#13281440](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13281440): pass

PR #225, opened 2023-10-17
Fix logging always using the device object within shared methods -- SIMICS-11346

lwaern-intel:
The `STATIC_ASSERT` warrants a run against VP. Just to be _sure._


test/1.4/lib/T_map_target_connect.py
```diff
@@ -53,13 +53,13 @@
...
+with stest.expect_log_mgr(obj.bank.z, log_type="spec-viol"):
     rex = re.compile("^failed to read 8 bytes @ 0x0")
     with stest.expect_log_mgr(obj, log_type="info", regex=rex):
         _ = obj.x_value
 
     rex = re.compile("^failed to write 8 bytes @ 0x0")
-    with stest.expect_exception_mgr(simics.SimExc_Attribute):
-        with stest.expect_log_mgr(obj, log_type="info", regex=rex):
+    with stest.expect_log_mgr(obj, log_type="info", regex=rex):
+        with stest.expect_exception_mgr(simics.SimExc_Attribute):
```
lwaern-intel, line 63: I took the liberty of correcting this. Before, the `stest.expect_log_mgr` didn't do what it was supposed to do due to the thrown exception.

syssimics:
Verification [#12361980](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12361980): pass

RELEASENOTES-1.4.docu
```diff
@@ -513,4 +513,12 @@
       within a port, bank or subdevice that is itself declared within an object
       array would lead to the type of that attribute being incorrectly
       registered <bug number="SIMICS-21474"/>.</add-note></build-id>
+  <build-id value="next"><add-note> Fixed a bug where log statements within
+      <tt>shared</tt> methods would always use the device's Simics object as the
```
mandolaerik, line 517: I think the term is "configuration object" or "Simics configuration object"

include/simics/dmllib.h
```diff
@@ -3130,10 +3130,24 @@ _DML_get_single_hook_attr(void *_hook, uintptr_t _aux) {
...
+    uint32 index_divisor;
+} _dml_log_object_assoc_t;
+
+UNUSED static inline conf_object_t *
+_identity_to_logobj(const _dml_log_object_assoc_t *object_assocs,
+                    conf_object_t *dev, _identity_t id) {
+    _dml_log_object_assoc_t obj_assoc = object_assocs[id.id - 1];
+    if (!obj_assoc.index_divisor) return dev;
+    return ((conf_object_t **)((char *)dev + obj_assoc.port_obj_offset))[
+        id.encoded_index/obj_assoc.index_divisor];
```
mandolaerik, line 3148: space around `/`

py/dml/c_backend.py
```diff
@@ -349,6 +351,8 @@ def generate_hfile(device, headers, filename):
 
     print_device_substruct(device)
 
+    out(f'STATIC_ASSERT(sizeof(struct {crep.cname(device)}) <= UINT32_MAX);\n')
```
mandolaerik, line 354: We could also `out("// allows us to store device struct offsets in uint32, which saves space\n")` both for ourselves and for the perpetrator.

py/dml/c_backend.py
```diff
@@ -2137,7 +2141,32 @@ def generate_object_vtables_array():
...
+def generate_log_object_assoc_array():
+    items = []
+    for node in dml.globals.objects:
+        object_node = (node if node.objtype in {'device', 'bank',
+                                                'subdevice', 'port'}
+                       else node.object_parent)
+        if (object_node is not dml.globals.device
+            # Anonymous banks, OR banks never given a name due to compile
+            # errors
+            and not object_node.name is None):
```
mandolaerik, line 2155: We don't need this precondition: Anonymous banks don't instantiate templates with shared methods (and it's OK to ICE if they do). And after a compile error, we can output whatever (and I don't understand how this could give `name is None` -- I think we have never-triggered assertions that `name is None` implies `objkind in {'field','bank'}`.

py/dml/c_backend.py
```diff
@@ -3291,6 +3320,8 @@ def generate_cfile_body(device, footers, full_module, filename_prefix):
     generate_events(device)
     generate_identity_data_decls()
     generate_object_vtables_array()
+    if compat.shared_logs_on_device not in dml.globals.enabled_compat:
+        generate_log_object_assoc_array()
```
mandolaerik, line 3324: naming: this is fine, but makes me relate to the unrelated term "associative array", often used for the dict datatype (in particular in the scheme/lisp world)

py/dml/compat.py
```diff
@@ -105,6 +105,19 @@ class port_obj_param(CompatFeature):
...
+@feature
+class shared_logs_on_device(CompatFeature):
+    '''This compatibility feature changes the semantics of log statements
+    inside shared methods so that they always log on the device object, instead
+    of the nearest enclosing Simics object like with non-shared methods.
+    This behaviour was a bug present since the very introduction of shared
+    methods, which has lead to plenty of script code having become reliant
+    on it, especially in regards to how banks log. This feature preserves the
+    bugged behaviour.'''
+    short = "logs inside shared methods always log on the device object"
```
mandolaerik, line 117: Other `short` strings use imperative mood. OK to break the pattern if it reads better. (try --help-no-compat)

py/dml/ctree.py
```diff
@@ -3534,6 +3536,19 @@ def ctype(self):
...
         return "(%s).id" % (self.traitref.read(),)
 
+class LogObjectFromObjIdentity(Expression):
+    priority = dml.expr.Apply.priority
+    @auto_init
+    def __init__(self, site, identity):
+        crep.require_dev(site)
+
+    def ctype(self):
+        return TNamed('_conf_object_t *')
```
mandolaerik, line 3546: what's a `_conf_object_t`?

test/1.4/legacy/T_shared_logs_on_device_disabled.dml
```diff
@@ -0,0 +1,44 @@
+/*
+  © 2023 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+dml 1.4;
+
+device test;
+
+/// DMLC-FLAG --simics-api=6
```
mandolaerik, line 9: Note that this will give a compile error when building DMLC for simics 8; someday we'll need to add an annotation to handle that. Today is not that day, though.

test/1.4/legacy/T_shared_logs_on_device_disabled.dml
```diff
@@ -0,0 +1,44 @@
...
+    hello();
+    /// GREP \[obj info\] hello from g
+    g.hello();
+    /// GREP \[obj info\] hello from g.c\[1\]
+    g.c[1].hello();
+    /// GREP \[obj\.g\.s\[2\] info\] hello from g\.s\[2\]
+    g.s[2].hello();
+    /// GREP \[obj\.g\.s\[2\] info\] hello from g\.s\[2\]\.c\[3\]\[5\]
+    g.s[2].c[3][5].hello();
+    /// GREP \[obj\.g\.s\[2\]\.c\[3\]\[5\]\.s info\] hello from g\.s\[2\]\.c\[3\]\[5\]\.s
```
mandolaerik, line 42: wonderful escapes

test/1.4/legacy/T_shared_logs_on_device_disabled.dml
mandolaerik: in fact this line is redundant; what we need is a test that --simics-api=6 *without* passing `--no-compat` gives the buggy behaviour. It is that test that will be problematic for simics 8.

mandolaerik: and now I see you wrote that test already, [T_shared_logs_on_device_enabled.dml](https://github.com/intel/device-modeling-language/pull/225/files#diff-1d79db2582c21fd49607b71ae51a13e6c26b798a14380455525b3abe780dc58b).

test/1.4/registers/instrumentation_test.py
```diff
@@ -42,11 +42,11 @@ def test(obj):
     # the instrumentation tests were disabled
     instrumentation_access_inquire.test(obj, subscribe_b2)
     with stest.allow_log_mgr(None, 'spec-viol'):
-        instrumentation_access_set_missed.test(obj, subscribe_b1, obj)
+        instrumentation_access_set_missed.test(obj, subscribe_b1, obj.bank.b1)
     instrumentation_access_set_offset.test(obj, subscribe_b1)
     instrumentation_access_set_value.test(obj, subscribe_b1)
     instrumentation_access_suppress.test(obj, subscribe_b1)
-    instrumentation_access_value_at_miss.test(obj, subscribe_b1, obj)
+    instrumentation_access_value_at_miss.test(obj, subscribe_b1, obj.bank.b1)
```
mandolaerik, line 49: I think we can remove the last arg, all callers pass obj.bank.b1 now

mandolaerik:
Looks great, only nitpicks

py/dml/c_backend.py
lwaern-intel: > We don't need this precondition
We need _a_ precondition. Because this _will_ run into anonymous banks, and the code under this `if` _will_ ICE if so. If not this check, then it'd be a check if the bank instantiates any template with a shared method.
> Anonymous banks don't instantiate templates with shared methods
It's certainly possible thanks to in-each; but I don't disagree that it's such a niche case we can simply ICE if we run into it.
> (and I don't understand how this could give `name is None`)
Not sure myself. `1.4/errors/EAMBINH` and `1.4/errors/EIDENT` ICEd when I restricted this precondition to DML 1.2, and went away when I didn't.

lwaern-intel: 1. and 2. are WRONG!!!!!! This check is not only relevant for anonymous banks themselves, they are relevant for any object _under_ an anonymous bank! I don't feel comfortable restricting all such objects to never instantiate any template with a shared method, and there I don't see why that would be better anyway, since the check I already have in place is so simple and does the right thing. Like, do we want to go out of our way just to _spite_ DML 1.2 users? Not to mention it's a breaking change.

py/dml/c_backend.py
lwaern-intel: Ah, true, didn't consider that... Intended meaning was "log object assocations." Will think over if I want to change it.

py/dml/ctree.py
lwaern-intel: _very_ good question.

I'm to used to dmllib's naming scheme.

test/1.4/legacy/T_shared_logs_on_device_disabled.dml
lwaern-intel: Hah! Yes. You think we should add an option to `GREP` so it doesn't operate on a regex?

test/1.4/registers/instrumentation_test.py
lwaern-intel: I think so too. I'll try it.

test/1.4/legacy/T_shared_logs_on_device_disabled.dml
lwaern-intel: You're complaining about this line, but I just followed the pattern of the legacy tests _you_ created!

py/dml/c_backend.py
lwaern-intel: `generate_log_object_assocs_array`, ez

lwaern-intel:
You may notice that there are a lot more commits than previous. They speak for themselves. Those with the same name as before have been left untouched.

test/1.4/registers/instrumentation_test.py
lwaern-intel: yup.

syssimics:
Verification [#12366594](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12366594): pass

lwaern-intel:
jenkins testing is broken _again._ No VP run it seems

test/1.4/legacy/T_shared_logs_on_device_disabled.dml
mandolaerik: no it's fine! simple is good. I just couldn't resist admiring the beauty.

test/1.4/legacy/T_shared_logs_on_device_disabled.dml
mandolaerik: Right! so whose fault is it that this mistake of mine wasn't caught in YOUR review of that code? :)

py/dml/c_backend.py
lwaern-intel: > BTW, my python linter (ruff) says X is not None is preferred over not X is None.
Yup, that was a goof. I'll correct if once you've done your quick look.

syssimics:
Verification [#12371660](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12371660): fail

mandolaerik (approved):

syssimics:
Verification [#12389503](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12389503): pass

PR #226, opened 2023-10-19
Make releasenotes display consistently

sgronlun:
In particular, use prefix for the tracker-system, always a space before the
bug-tag, as well as no dot before the tag but after - SIMICS-21408.


lwaern-intel:
We have `RELEASENOTES` and `RELEASENOTES-1.4` too; they also need to be patched.

mandolaerik (approved):

syssimics:
Verification [#12371038](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12371038): pass

syssimics:
Verification [#12371273](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12371273): pass

PR #227, opened 2023-10-26
Move bank instrumentation to bank.{read/write}

mandolaerik:


mandolaerik:
This makes sure instrumentation works also for device-internal accesses, which allows a clean migration of sideband banks to the transaction interface. This also moves the feature further away from the memop, which  means:
* the reported size is deduced from byte enables, rather than taken from the transaction
* the reported initiator is harder to reach; our workaround is to convey it in a session variable back-door, which in some cases will cause an incorrect initiator to be reported in instrumentation callbacks after `read`/`write` if `SIM_transaction_wait` interacts with reentrancy.

syssimics:
Verification [#12409052](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12409052): fail

syssimics:
Verification [#12409305](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12409305): pass

mandolaerik:
[Jenins](https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=12411079&tab=queuedBuildOverviewTab)

mandolaerik:
Jenkins looks ok

lib/1.4/dml-builtins.dml
```diff
@@ -2262,53 +2311,33 @@ template bank is (object, shown_desc) {
...
 
+        dev._current_initiator = prev_initiator;
         return success;
     }
 
     shared method transaction_access(transaction_t *t, uint64 offset,
                                      void *aux) -> (exception_type_t) default {
         local uint64 size = SIM_transaction_size(t);
-        local conf_object_t *ini = SIM_transaction_initiator(t);
+        local conf_object_t *ini = dev._current_initiator;
```
lwaern-intel, line 2340: This surprises me. We don't need `SIM_transaction_initiator` nor `SIM_get_mem_op_initiator` at all anymore? Their results will always be equal to the object we get in `_transaction_access`?

mandolaerik: Oh, I think this hunk should be dropped, apparently our tests don't seem to cover initiator.

mandolaerik:
VP passing is probably due to their tests not relying much on bank instrumentation (in part because bank instrumentation don't work for sideband accesses before this PR).

I think their copied overrides will cause callbacks to be called twice, but things that do rely on bank instrumentation are likely on the form "wait for next access", where a double callback probably isn't disruptive.

mandolaerik:
[Here](https://jira.devtools.intel.com/browse/SMD-3715) is the issue for fixing this in VP.

mandolaerik:
Before merging this, I should also add a release note.

mandolaerik:
I can see that the VP duplication issue is scheduled for fixing before nov 17, so we can consider waiting for that.

RELEASENOTES-1.4.docu
```diff
@@ -521,4 +521,12 @@
...
       With version 6 or below it must be explicitly enabled by passing
       <tt>--no-compat=shared_logs_on_device</tt> to DMLC.</add-note></build-id>
+  <build-id value="next"><add-note> Bank instrumentation callbacks are now
+      issued from the <tt>read</tt> and <tt>write</tt> methods instead of the
+      <tt>transaction_access</tt>/<tt>io_memory_access</tt> method of
+      banks. This means that breakpoints on banks will trigger also on
+      reads or writes initiated within the device. This also means
+      that if <tt>read</tt> or <tt>write</tt> is overridden in a bank,
+      then instrumentation callbacks will only be called if they
+      call <tt>default()</tt>.</add-note></build-id>
```
mandolaerik, line 531: This perhaps deserves a discussion, there are probably valid use cases for read/write overrides where this change will be a regression. I don't know what to do about that.

syssimics:
Verification [#12413771](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12413771): pass

PR #228, opened 2023-10-27
Make global device/hook lists available only during code generation

mandolaerik:
- Fix spelling
- Remove redundant flag
- Make global device list available only during code generation
- Make global hook list available only during code generation
- Only worry about None names in obsolete DML

### DIFF
--- a/py/dml/c_backend.py
+++ b/py/dml/c_backend.py
@@ -2124,14 +2124,8 @@ def generate_port_object_assocs_array():
                        else node.object_parent)
         if (object_node is not dml.globals.device
             # Anonymous banks
-            and not (compat.dml12_misc in dml.globals.enabled_compat
-                     and object_node.ident is None)
-            # HACK compile errors may lead to .name of nodes never being set.
-            # This issue has proven difficult to fix due to multiple technical
-            # debts, most egregiously that the various uses of
-            # dml.globals.objects and dml.globals.hooks don't properly take
-            # into account the possibility of objects becoming rejected
-            and object_node.name is not None):
+            and (compat.dml12_misc not in dml.globals.enabled_compat
+                 or object_node.ident is not None)):
             port_obj_offset = (
                 f'offsetof({crep.structtype(dml.globals.device)}, '
                 + f'{crep.cref_portobj(object_node, ())})')
--- a/py/dml/c_backend.py
+++ b/py/dml/c_backend.py
@@ -1473,7 +1473,7 @@ def generate_identity_data_decls():
         add_variable_declaration(f'ht_str_table_t {prefix}_id_info_ht',
                                  'HT_STR_NULL(false)')
 
-    declare_id_infos(dml.globals.objects, '')
+    declare_id_infos(objects.Device.objects, '')
     if dml.globals.hooks:
         declare_id_infos(dml.globals.hooks, '_hook')
 
@@ -1488,8 +1488,8 @@ def generate_init_identity_hashtable():
     # This gets overridden if there's a top-level object that shares its name
     # with the device.
     out(f'ht_insert_str(&_id_info_ht, "{dml.globals.device.name}", '
-        + f'&_id_infos[{dml.globals.device.uniq} - 1]);\n')
-    out('for (uint64 i = 0; i < %d; ++i) {\n' % (len(dml.globals.objects),),
+        + f'&_id_infos[{dml.globals.device.uniq - 1}]);\n')
+    out('for (uint64 i = 0; i < %d; ++i) {\n' % (len(objects.Device.objects),),
         postindent=1)
     out('ht_insert_str(&_id_info_ht, _id_infos[i].logname, &_id_infos[i]);\n')
     out('}\n', preindent=-1)
@@ -2102,7 +2102,7 @@ def generate_trait_deserialization_hashtables(device):
 
 def generate_object_vtables_array():
     items = []
-    for node in dml.globals.objects:
+    for node in objects.Device.objects:
         if node.traits:
             node.traits.mark_referenced(dml.globals.object_trait)
             ancestry_path = node.traits.ancestry_paths[dml.globals.object_trait][0]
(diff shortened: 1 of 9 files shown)


py/dml/objects.py
```diff
@@ -216,8 +215,12 @@ def __init__(self, ident, site, parent, arraylens = (), idxvars = ()):
...
         self._confidential = False
 
-        self.uniq = len(dml.globals.objects) + 1
-        dml.globals.objects.append(self)
+        # during code generation, set to a unique positive integer,
+        # used as object ID, e.g. in the id_infos array.  This integer
+        # number is the index into object.Device.objects *plus one*;
+        # the plus one is because object ID 0 is reserved in runtime
+        # for an invalid reference.
+        self.uniq = None
```
mandolaerik, line 223: since uniq is not just any unique integer, but *the* object ID of the object used in runtime, perhaps we should also rename it, e.g. what about `object_id` (and `hook_id`)?

py/dml/c_backend.py
```diff
@@ -1238,7 +1238,7 @@ def generate_simple_events_control_methods(device):
 
     site = logging.SimpleSite('<_cancel_simple_events>')
     by_dims = {}
-    for hook in dml.globals.hooks:
+    for hook in objects.Device.hooks:
```
lwaern-intel, line 1241: You've earlier pushed for global variables being exclusively inside `dml.globals` for the potential future development of a DMLC daemon. Have you changed your mind?

py/dml/objects.py
```diff
@@ -319,6 +325,31 @@ def add_init_data(self, node):
     def idxvars(self):
         return ()
 
+    # A device is a reusable, but not reentrant, context manager, entered
+    # during code generation.
```
lwaern-intel: Interesting, but what does it give us? It doesn't help with the potential future development of a DMLC daemon, because we still have the other global variables which are not tied to any context manager, and I'm not convinced a context manager is a correct call if not placed across DMLC's entire operation -- right now, for example, it excludes `process`, and thus `mkobj` of the device object. That does not sit right with me; are we sure nothing would require `uniq`s or the object array in those contexts?

py/dml/objects.py
```diff
@@ -461,8 +492,10 @@ def __init__(self, name, site, parent, msg_types, arraylens = ()):
         self.msg_types = msg_types
         self._arraylens = arraylens
         self.dimensions += self.local_dimensions()
-        self.uniq = len(dml.globals.hooks) + 1
-        dml.globals.hooks.append(self)
+        # during code generation, set to a unique positive integer
+        # (index in object.Device.hooks plus one, see
+        # CompositeObject.uniq)
+        self.uniq = None
```
lwaern-intel, line 502: redundant

py/dml/objects.py
lwaern-intel: I see your point. I would rather have a standardized name than just tacking a prefix onto `id`. The "array of things and index of thing used for identification" pattern is one I've used a lot, just search for `uniq` and you'll find tons of applications. Even more in RAII. I don't have a good name for it beyound `uniq`. `idx`? That can be confused with other things.

lwaern-intel, line 224: This belongs under `CompositeObject`, not `DMLObject`.

syssimics:
Verification [#12412426](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12412426): fail

lwaern-intel (changes requested):
To solve the issue of anonymous banks and fields we still need to introduce a variant of `logname_anonymized` that gives a no-really-i-guarantee-its-unique string for use in our various machinery, and which gives anonymous banks and fields placeholder names via counting. That method would also incorporate the `'dev' if node is dml.globlals.device` logic that `generate_identity_data_decls` has. Unfortunately it has to be separate as `QName.fmt` leverages `logname_anonymized` and for it we don't want to give dummy names to anonymous banks and fields.
There's also the backwards checkpointing compatibility issue -- the new variant of `logname_anonymized` would e.g. cause events inside anonymous banks to receive new event names and thus break checkpoints.

py/dml/c_backend.py
mandolaerik: The idea with moving stuff to `globals` is essentially to eventually collect that to only be accessible through some context manager, because that resolves most of the problems with globality. Now I lock stuff behind a different context manager with a narrower scope.

py/dml/objects.py
mandolaerik: If they do, then we have to handle that somehow. The context manager gives us a clear rule, which seems reasonable.

However, perhaps we should have some fallback value with broken `__str__` instead of `None`, to capture violations earlier; EventClassParamExpr violated the rule which gave some confusing late errors when C code delcared `_evclass_None`.

py/dml/objects.py
mandolaerik: This is `CompositeObject.__init__`.

py/dml/objects.py
mandolaerik: No, hook is not a CompositeObject, and it has a `uniq` slot

py/dml/objects.py
lwaern-intel: So it is, I failed at GitHub.

py/dml/objects.py
lwaern-intel: So it was, I failed at GitHub and thought it under `Group`.

syssimics:
Verification [#12462948](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12462948): fail

mandolaerik:
> To solve the issue of anonymous banks and fields we still need to introduce a variant of `logname_anonymized` that gives a no-really-i-guarantee-its-unique string for use in our various machinery,

Discussed OOGH, choosing to ignore this for now: AFAICS, these bugs are mostly pre-existing, and relate to the use of certain combinations of niche 1.4 features within 1.2 devices. Some reasonable corners like `in each field { event e; }` may give you explicit errors and thus is no big problem (people don't have to debug, they can just adapt or report). There are other intricate corners which can lead to more subtle bugs, like object identity not working properly:
```
register r is foo{
  in each field { is foo; }
  method m() {
    foreach x in (each foo in (r)) {
      assert cast(r, foo) != x;
    }
  }
}
```
which indeed has the potential for hard-to-debug errors where a proper fix would save time for users. But all such use cases I can think of are combinations of such obscure circumstances (whole-field register AND in-each:ed template AND reliance on identity AND dml 1.2) that they are unlikely to ever happen.

syssimics:
Verification [#12466810](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12466810): pass

syssimics:
Verification [#12466834](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12466834): pass

lwaern-intel (approved):
You may or may not want to squash some commits together

syssimics:
Verification [#12467579](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12467579): pass

PR #231, opened 2023-11-01
Add test for `if`-`else` ambiguity

lwaern-intel:
To test both our parser's behavior and that generated C doesn't provoke `Wparantheses`.

### DIFF
--- a/test/1.2/statements/T_if_ambiguity.dml
+++ b/test/1.2/statements/T_if_ambiguity.dml
@@ -1,0 +1,84 @@
+/*
+  © 2023 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+dml 1.2;
+
+device test;
+
+// This tests both that GCC doesn't emit Wparentheses and the behavior of
+// the DML parser.
+// Although GCC emitting Wparentheses could be considered desirable for the
+// patterns this test exemplifies, even if braces were inserted to disambiguate
+// the generated C would be identical due to `mkCompound` collapse.
+// Thus Wparentheses must be avoided in general.
+
+import "testing.dml";
+
+data bool t = true;
+data bool f = false;
+
+method test() -> (bool ok) {
+    // Each of these test cases tests that the DML parser associates the 'else'
+    // with the inner 'if', and not the outer 'if' (this happens no matter how
+    // the code is indented).
+    // In addition, we expect DMLC to avoid Wparentheses by inserting braces
+    // around the body of the outer if. E.g. the first test case should
+    // generate:
+    //
+    // if (_dev->t)
+    // {
+    //     if (_dev->f)
+    //     DML_ASSERT(...)
+    //     else
+    //     v10_ok = 1;
+    // }
+    //
+    // instead of:
+    //
+    // if (_dev->t)
+    // if (_dev->f)
+    // DML_ASSERT(...)
+    // else
+    // v10_ok = 1;
+    //
+    // ... which would lead to GCC emitting Wparentheses due to the ambiguity
+    ok = false;
+    if ($t)
+        if ($f)
+            assert false;
+        else ok = true;
+
+    assert ok;
+    ok = false;
+
+    if ($t)
+        while (!ok)
+            if ($f)
+                assert false;
+            else ok = true;
+
+    assert ok;
+    ok = false;
+
+    if ($t)
+        for (;!ok;)
+            if ($f)
+                assert false;
+            else ok = true;
+
+    assert ok;
+    ok = false;
+
+    local int vect v;
+    VINIT(v);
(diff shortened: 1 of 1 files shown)


test/1.2/statements/T_if_ambiguity.dml
```diff
@@ -0,0 +1,60 @@
...
+dml 1.2;
+
+device test;
+
+// This tests both that GCC doesn't emit Wparantheses and the behavior of
+// the DML parser.
+// Although GCC emitting Wparantheses could be considered desirable for the
+// patterns this test exemplifies, even if braces were inserted to disambiguate
+// the generated C would be identical due to `mkCompound` collapse.
+// Thus Wparantheses must be avoided in general.
```
mandolaerik, line 14: if this mainly covers that the test looks good to GCC, then `/// COMPILE-ONLY` should suffice?

Also, looks a bit like a smoke test; did you consider testing on a ctree_test level? E.g., `toc()` difficult corners into a StrOutput and compare the result character-by-character. The upside would be a test case where you can see the brace generation difficulties visually; the downside would be that the test would go more technical and brittle.

lwaern-intel: The runtime component is to test the behavior of the DML parser -- to which `if` each `else` is associated with. If we're only interested testing for `Wparantheses` then `/// COMPILE-ONLY` would suffice, yes.

This is not really a smoke test: what we care about, explicitly, is that GCC doesn't emit Wparantheses for uses of if-else, and that can only occur in very specific circumstances. See https://github.com/gcc-mirror/gcc/blob/master/gcc/testsuite/gcc.dg/Wparentheses-12.c and https://github.com/gcc-mirror/gcc/blob/master/gcc/testsuite/gcc.dg/Wparentheses-13.c . In our case, this can only possibly occur with DML statements for which:
1. The definition of `toc_stmt()` itself uses `toc_stmt()` of some arbitrary parameter statement
2. The definition of `toc_stmt()` doesn't enclose the inner use of `toc_stmt()` in braces

The relevant statements for this is applicable are rather easy to find. The only one which satisfies both and isn't covered here is `DoWhile`, and that's because it eliminates ambiguity by virtue of the `while (...);`.

lwaern-intel: In some sense, leveraging `ctree_test` would be _more_ of a smoke test because then we wouldn't run it past GCC.

lwaern-intel: @mandolaerik Reminder. Are you convinced by my arguments?

test/1.2/statements/T_if_ambiguity.dml
```diff
@@ -0,0 +1,60 @@
...
+  © 2023 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+dml 1.2;
+
+device test;
+
+// This tests both that GCC doesn't emit Wparantheses and the behavior of
+// the DML parser.
+// Although GCC emitting Wparantheses could be considered desirable for the
```
mandolaerik, line 11: `s/Wparantheses/Wparentheses/`

Perhaps warrants adding "parentheses" to the wordlist

test/1.2/statements/T_if_ambiguity.dml
mandolaerik: Sorry I forgot this. I just wanted to say there's something about the test that itches a bit, and I gave my best guess why it does. But itchy tests with good coverage are OK; we can improve the test later if it ever becomes a real problem.

test/1.2/statements/T_if_ambiguity.dml
```diff
@@ -0,0 +1,60 @@
...
+
+data bool t = true;
+data bool f = false;
+
+method test() -> (bool ok) {
+    ok = false;
+    if ($t)
+        if ($f)
+            assert false;
+        else ok = true;
```
mandolaerik, line 26: would be good with some comment to show what this tests and how, like "This generates `if ...`, rather than `if ...` which gcc would dislike"

test/1.2/statements/T_if_ambiguity.dml
lwaern-intel: Iiii suspect it would blow up half of Simics base ;)

or maybe it's just me who makes that mistake

test/1.2/statements/T_if_ambiguity.dml
lwaern-intel: Done

syssimics:
Verification [#12511694](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12511694): pass

mandolaerik (approved):

PR #233, opened 2023-11-07
Survive ENAMEID better

mandolaerik:


syssimics:
Verification [#12466826](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12466826): pass

lwaern-intel (approved):
I like this.

PR #234, opened 2023-11-07
Add a warning for large unrolled loops

lwaern-intel:
This is both a proper PR and also just a Jenkins Pretest target to figure out how bad the issue is on VP


test/1.2/errors/T_WBIGUNROLL.dml
```diff
@@ -0,0 +1,80 @@
...
+    foreach x in ($zeroes_51) {
+        // As fiftieth bit is constant 0, DML 1.2 semantics eliminates this if
+        // entirely, and subsequently causes that iteration to be omitted from
+        // codegen entirely; reducing the total count to 50
+        if (x != 0) assert true;
+    }
+
+    // As fiftieth bit is constant 0, select can cut short at it, reducing the
+    // total number of iterations to 50
+    select x in ($zeroes_50) where (x == 0) {
```
mandolaerik, line 62: did you mean zeroes_51?

test/1.2/errors/T_WBIGUNROLL.dml
```diff
@@ -0,0 +1,80 @@
...
+
+data int zero = 0;
+
+parameter zeroes_50 = [$zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero,
+                       $zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero,
+                       $zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero,
+                       $zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero,
+                       $zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero,
+                       $zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero,
+                       $zero, $zero];
```
mandolaerik, line 20: for readability, consider
```
param zeroes_10 = [$zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero]
param zeroes_50 = $zeroes_10 + $zeroes_10 + $zeroes_10 + $zeroes_10 + $zeroes_10;
..
assert zeroes_50.len == 50;
```

py/dml/codegen.py
```diff
@@ -3048,12 +3048,19 @@ def stmt_select(stmt, location, scope):
...
                     break
 
+            iterations = len(clauses)
             if else_dead:
                 (last_cond, last_stmt) = clauses.pop(-1)
                 assert last_cond.constant and last_cond.value
                 if_chain = last_stmt
             else:
                 if_chain = codegen_statement(else_ast, location, scope)
+            if iterations > 50 and isinstance(lst, List):
```
mandolaerik, line 3058: I would bump the limit to at least 65 because of "bits in an integer" superstition.

syssimics:
Verification [#12467586](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12467586): pass

test/1.2/errors/T_WBIGUNROLL.dml
lwaern-intel: ... you can concatenate compile-time lists?

test/1.2/errors/T_WBIGUNROLL.dml
lwaern-intel: yes

py/dml/codegen.py
lwaern-intel: Sure. The number's arbitrary anyway.

test/1.2/errors/T_WBIGUNROLL.dml
lwaern-intel: No you can't.

test/1.2/errors/T_WBIGUNROLL.dml
```diff
@@ -0,0 +1,80 @@
...
+
+// Fiftieth zero is constant
+parameter zeroes_51 = [$zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero,
+                       $zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero,
+                       $zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero,
+                       $zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero,
+                       $zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero,
+                       $zero, $zero, $zero, $zero, $zero, $zero, $zero, $zero,
+                       $zero, 0, $zero];
```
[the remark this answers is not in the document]
lwaern-intel: Nah.

syssimics:
Verification [#13501484](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13501484): ✅ pass

RELEASENOTES.docu
```diff
@@ -1795,4 +1795,11 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
   <build-id _6="next" _7="next"><add-note>Multiple `extern` declarations for the
       same identifier are now allowed as long as they all declare the same type
       for the identifier <bug number="SIMICS-22472"/>.</add-note></build-id>
+  <build-id _6="next" _7="next"><add-note> Added the warning <tt>WBIGUNROLL</tt>
+      which is emitted when a <tt>#select</tt> or <tt>#foreach</tt> statement is
+      compiled to a large unrolled loop.
+      This warning is only enabled by default with Simics API version 8 or
+      above. With version 7 and below it must be explicitly enabledby passing
```
mandolaerik, line 1802: `enabled by`

py/dml/codegen.py
```diff
@@ -3031,12 +3031,19 @@ def stmt_select(stmt, location, scope):
...
+            iterations = len(clauses)
             if else_dead:
                 (last_cond, last_stmt) = clauses.pop(-1)
                 assert last_cond.constant and last_cond.value
                 if_chain = last_stmt
             else:
                 if_chain = codegen_statement(else_ast, location, scope)
+            if iterations > WBIGUNROLL.limit and isinstance(lst, List):
+                report(WBIGUNROLL(stmt.site,
+                                  '#'*(stmt.site.dml_version() != (1, 2))
```
mandolaerik, line 3043: space around binop

py/dml/codegen.py
```diff
@@ -3162,6 +3169,10 @@ def foreach_constant_list(site, itername, lst, statement, location, scope):
                     stmt)
             spec.append(mkCompound(site, decls + [stmt]))
 
+        if len(spec) > WBIGUNROLL.limit and isinstance(lst, List):
+            report(WBIGUNROLL(site,
+                              '#'*(site.dml_version() != (1, 2)) + 'foreach',
```
mandolaerik, line 3174: space around binop

py/dml/compat.py
```diff
@@ -198,6 +198,30 @@ class suppress_WLOGMIXUP(CompatFeature):
     short = "Suppress the warning 'WLOGMIXUP' by default"
     last_api_version = api_6
 
+@feature
+class suppress_WBIGUNROLL(CompatFeature):
+    '''This compatibility feature makes it so the warning `WBIGUNROLL` is
+    suppressed by default. `WBIGUNROLL` warns about `#foreach` and `#select`
+    statements that compile to large unrolled loops &mdash; for more
+    information, see the documentation of `WBIGUNROLL` in the
+    [Messages](messages.html) section.
```
mandolaerik, line 207: better with a direct link, add `<a id=` on the destination if needed

py/dml/compat.py
```diff
@@ -198,6 +198,30 @@ class suppress_WLOGMIXUP(CompatFeature):
...
+    version 8.
+
+    Passing `--no-compat=suppress_WBUGUNROLL` to DMLC has almost the same effect
+    as passing `--warn=WBIGUNROLL`; either will cause DMLC to report the warning
+    even when the Simics API version in use is below 8. The only difference
+    between these two options is that if `--no-compat=suppress_WBIGUNROLL` is
+    used (and `--warn=WBIGUNROLL` is not), then `WBIGUNROLL` may still be
+    explicitly suppressed via `--no-warn=WBIGUNROLL`. In contrast,
+    `--warn=WBIGUNROLL` doesn't allow for `WBIGUNROLL` to be suppressed at
+    all.'''
```
mandolaerik, line 221: that's somewhat weird, when you pass `--X --no-X` to a compiler, you expect that the one that appears last takes precedence.

In any case, I think we can drop this paragraph entirely; there is no essential difference and the main reason for a compat feature is for consistency with other things that change with API 8.

py/dml/compat.py
mandolaerik: Oh, seems that `messages.html#dt:wbigunroll` would work out-of-the-box

test/1.2/legacy/suppress_WBIGUNROLL.dml
```diff
@@ -0,0 +1,80 @@
+/*
+  © 2024 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+dml 1.2;
+
```
mandolaerik, line 6: I think we should make the warning specific to 1.4, big unroll is a valid idiom in 1.2 (as demonstrated by dml-builtins). (Unsure how common valid non-collapsed unrolls are in practice, though)

py/dml/compat.py
lwaern-intel: whaaat
That's a thing? Wow.

test/1.2/legacy/suppress_WBIGUNROLL.dml
lwaern-intel: The thing you're talking about is already, intentionally, not covered by `WBIGUNROLL`: this _only_ warns about big unrolls when performed on compile-time lists specified by the model, and does not warn about `foreach` on `ObjectList`.

py/dml/codegen.py
lwaern-intel: This is a stylistic choice we can quibble about: when combining `+` and `*`, I actually prefer omitting spaces around `*`; I feel that increases readability re. operator precedence. I see it as similar to implicit multiplication in math via concatenation (`2(3)`).

py/dml/compat.py
lwaern-intel: I agree with dropping the explanation of the difference; that was inherited from the description of `suppress_WLOGMIXUP`, but it doesn't bear repeating. Though, I think I should keep the mention that you can either use `--warn=WBIGUNROLL` or `--no-compat=suppress_WBIGUNROLL`.

test/1.2/legacy/suppress_WBIGUNROLL.dml
lwaern-intel: > (Unsure how common valid non-collapsed unrolls are in practice, though)
_Too common._ Don't you remember the context behind this PR's creation? And delay to merging?

py/dml/compat.py
mandolaerik: >  I should keep the mention that you can either use --warn=WBIGUNROLL or --no-compat=suppress_WBIGUNROLL.

Agreed

py/dml/compat.py
lwaern-intel: Though `validate_md_links.py` hates it. I guess we could patch it to tolerate `#dt:blabla` if any line in the target file starts with `* blabla`?

py/dml/codegen.py
mandolaerik: Here we don't combine `+` and `*`. but we do combine `*` and ` != `.

And I agree that PEP-8 is not always optimal; in particular, I find `f(foo=bar == 2)` a bit weird. But I just accept and obey the style for consistency, unless there's strong reasons not to.

test/1.2/legacy/suppress_WBIGUNROLL.dml
mandolaerik: Discussed on IM, dropping this request.

py/dml/codegen.py
lwaern-intel: > Here we don't combine `+` and `*`

I mean... we are, though the `+` _is_ on a new line. That's good enough for readability, I admit. Though that doesn't apply to the below `'#'*(site.dml_version() != (1, 2)) + 'foreach',`

> we do combine `* and `!=`.

Not on the same syntactic level, the `!=` is within an `()`; I'm talking about `*` and `+`, which is on the same syntactic level, and where operator precedence determines how associations are made. This is where I'm arguing that omitting spaces around `*` such that such parts are more obviously whole terms helps readability. To exemplify, compare
```
   bla_bla_blau * blib * blub_blub + glabradahadaba + blara * dabara
+ canyoutellthat * ihavegivenup + on * variable * naming
```
vs.
```
   bla_bla_blau*blib*blub_blub + glabradahadaba + blara*dabara
+ canyoutellthat*ihavegivenupon + on*variable*naming
```

And that does apply to `'#'*(site.dml_version() != (1, 2)) + 'foreach'`, but not as strongly. If you write out:
```
'#' * (site.dml_version() != (1, 2)) + 'foreach'
```
It's not as obviously clear at a glance that the `(site.dml_version() != (1, 2))` is used for controlling a term of the string concatenation sequence, as opposed to a term in of itself. Of course, the counterargument to that would be parentheses...:
```
('#' * (site.dml_version() != (1, 2))) + 'foreach'
```
If you insist on spacing, I would do the above. even though i kinda hate it.

py/dml/compat.py
lwaern-intel: > Oh, seems that messages.html#dt:wbigunroll would work out-of-the-box
wait. What experiment did you do, that made you say this?
If we look at what `messages_to_md.py` does...
```
        f.write(f"  <dt><b>\n\n{fmt_message(m)} [{m.__name__}]</b></dt>\n")
```
Looking at this, I don't understand why the hypelink syntax would pick out and work specifically with `m__name__`.

But! Maybe the better approach is to just tweak `messages_to_md.py` to generate those anchors, so we don't have to hand-write them.

mandolaerik: > What experiment did you do, that made you say this?
I looked at the generated messages.html, and also typed `file:///path/to/messages.html#dt:wsystemc` into my web browser

lwaern-intel: ... huh!

PR #236, opened 2023-11-07
Fix incorrect linemark tracking with `--coverity`

lwaern-intel:
This improves defect reporting. Admittedly a rather lazy solution.

### DIFF
--- a/py/dml/output.py
+++ b/py/dml/output.py
@@ -64,6 +64,12 @@ class Output(object):
             self.out('#line %d "%s"\n'
                      % (self.lineno + 1, quote_filename(self.filename)))
 
+    class IndeterminateFilename: pass
+
+    def indeterminate_line_directive(self):
+        self.redirected_filename = Output.IndeterminateFilename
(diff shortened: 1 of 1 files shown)


lwaern-intel:
@mandolaerik NAG

py/dml/output.py
```diff
@@ -64,6 +64,12 @@ def reset_line_directive(self):
             self.out('#line %d "%s"\n'
                      % (self.lineno + 1, quote_filename(self.filename)))
 
+    class IndeterminateFilename: pass
```
mandolaerik, line 67: I don't understand -- no code checks for this, so what's the desired property? that it's `!=` all paths and `None`? Is it ever `__str__`ed?

Consider singletoning the class, to mark the intent of an object without properties, as opposed to a class without properties (I sometimes use a decorator `@lambda cls: cls()`; I miss the `@apply` construct from py2).

mandolaerik (approved):

PR #237, opened 2023-11-09
Discard reference -- SIMICS-21584

lwaern-intel:
I don't like how the way the compatibility feature is implemented, but doing it at AST building level was the best I could do. Any attempt to try to do it at parser level fell apart.


syssimics:
Verification [#12479725](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12479725): fail

syssimics:
Verification [#12484192](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12484192): pass

lwaern-intel:
We have segfaults, by the way! This may or may not be due to the thing I complained about, regardless, you'll want to track them down.

syssimics:
Verification [#12510780](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12510780): pass

py/dml/c_backend.py
```diff
@@ -1859,37 +1859,39 @@ def generate_init_data_objs(device):
...
+                        init = eval_initializer(
+                            node.site, node._type, node.astinit,
+                            Location(node.parent, index_exprs),
+                            global_scope, True)
+                    except DMLError as e:
+                        report(e)
+                    else:
+                        markers = ([('store_writes_const_field', 'FALSE')]
+                                   if deep_const(node._type) else [])
+                        coverity_markers(markers, init.site)
```
mandolaerik, line 1876: why not
```
if deep_const(...):
  coverity_markers(..)
```
?

lwaern-intel: To generate linemarks even if `deep_const` is not True; `coverity_markers([], site)` is equivalent to `site_linemark(site)`. As for the validity of generating linemarks for session/saved initializers, we've covered that in #221.

py/dml/codegen.py
```diff
@@ -1207,6 +1207,13 @@ def expr_variable(tree, location, scope):
         if in_dev_tree:
             e = in_dev_tree
     if e is None:
+        # TODO/HACK: The discard ref is exposed like this to allow it to be as
+        # keyword-like as possible while still allowing it to be shadowed.
+        # Once we remove support for discard_ref_shadowing the discard ref
+        # should become a proper keyword and its codegen be done via dedicated
+        # dispatch
+        if name == '_' and tree.site.dml_version() != (1, 2):
```
mandolaerik, line 1215: This may not suffice as a reminder, since the comment is not adjacent to the code that will be touched when the compat is removed. One may argue that you mention discard_ref_shadowing in the comment so grep will find it, but you may consider a `# when removing, remember to also fix ..` comment near the enabled_compat-checking code.

py/dml/codegen.py
```diff
@@ -2399,53 +2430,66 @@ def stmt_assign(stmt, location, scope):
...
-            for (tgt, sym) in zip(tgts, syms))
-        return stmts
+            if isinstance(tgt, NonValue):
+                expr = codegen_src_for_nonvalue_target(site, tgt, src_ast,
+                                                       location, scope)
+                stmt_pairs.append((mkCopyData(tgt.site, expr, tgt), None))
+            else:
+                init = eval_initializer(site, tgt.ctype(), src_ast, location,
+                                        scope, False)
+                name = 'tmp%d' % (i,)
```
mandolaerik, line 2442: `_tmp%d`? (because what if `extern int tmp0(void);` exists and the initializer of a subsequent assignment calls it)

mandolaerik: oh, the problem was not introduced by you. Still valid but not blocking then.

py/dml/compat.py
```diff
@@ -117,6 +117,13 @@ class shared_logs_on_device(CompatFeature):
     short = "Make logs inside shared methods always log on the device object"
     last_api_version = api_6
 
+@feature
+class discard_ref_shadowing(CompatFeature):
+    '''This compatibility feature allows declarations (within methods or
+    objects) to be named '_'. This will cause the discard reference `_` to be
```
mandolaerik, line 123: `s/'_'/'\_'/`, or backtick. Unescaped underscores in markdown are scary.

py/dml/dmlparse.py
```diff
@@ -2632,6 +2659,12 @@ def ident(t):
 def ident(t):
     t[0] = t[1]
 
+def ident_enforce_not_discard_ref(site, ident):
+    if (str(ident) == '_'
```
mandolaerik, line 2663: `str` should be redundant?

py/dml/dmlparse.py
```diff
@@ -2350,23 +2371,27 @@ def hashselect(t):
...
 @prod_dml14
 def foreach(t):
     'statement_except_hashif : FOREACH ident IN LPAREN expression RPAREN statement'
+    ident_enforce_not_discard_ref(site(t, 2), t[2])
     t[0] = ast.foreach(site(t), t[2], t[5], t[7])
 
 @prod_dml14
 def hashforeach(t):
     'statement_except_hashif : HASHFOREACH ident IN LPAREN expression RPAREN statement'
+    ident_enforce_not_discard_ref(site(t, 2), t[2])
```
mandolaerik, line 2394: Lots of these calls. Perhaps easier to make a single call in the `ident` production rule? And create a separate `ident_or_underscore` production rule, which would be like `ident` but without that call?

mandolaerik (approved):
Approved if you fix or dismiss remaining comments.

py/dml/dmlparse.py
lwaern-intel: Nope. I tried, but ran into reduce/reduce conflicts.

py/dml/dmlparse.py
lwaern-intel: _Probably._ I did it because `reserved` also does it.

py/dml/compat.py
lwaern-intel: oh right

py/dml/dmlparse.py
mandolaerik: Oh, but there should be some middle path: There should be no reduce/reduce after `HASHFOREACH`, so there you can use `ident`, and in other cases (the start token of `statement`, I suppose) there will be reduce/reduce so for those can do `ident_or_underscore` with a manual `ident_enforce_not_discard_ref`. Could make sense if it's not a vast majority of all cases?

py/dml/dmlparse.py
mandolaerik: oh, that's probably python2 legacy, the `unicode` vs `str` vs `bytes` discrepancy. Everything is strings now (unicode errors are caught in `toplevel`), so this can be removed in `reserved` too.

py/dml/dmlparse.py
lwaern-intel: Such ad-hoc approaches would be very fragile to parser changes (i.e. a change could force rewriting an existing `ident` to `ident_or_underscore`) not to mention would introduce an inconsistency (_either_ `ident_or_underscore` is used or a manual `ident_enforce_not_discard_ref` is.) This would only serve to make the parser even more difficult to understand. So I feel it'd be worse.

I agree that the `ident_enforce_not_discard_ref` is sub-par, and would welcome any superior approach _which is at least as general and consistently applicable._ `ident_decl` (i.e. `ident` and not `ident_or_underscore`) would be preferrable, if LALR didn't ruin our day.

py/dml/codegen.py
lwaern-intel: I just want to note that any issue this could only possibly is incredibly niche. In particular, this could only present an issue if `-g` is switched on, because otherwise the C variable name will get uniquified. (Notice that all codegen involved here uses `scope` and not `lscope`, so DML resolution will never take these temporary variables into account; the only possible issue is if the wrong thing happens at C level).

But yes; if `-g` is switched on, and the initializer references some C literal `tmp0` (made accessible via `extern`), then this could present an issue. But it's rather far-fetched.

py/dml/dmlparse.py
lwaern-intel: Nice

py/dml/ctree.py
```diff
@@ -4847,15 +4906,35 @@ def assign_to(self, dest, typ):
...
-        if deep_const(typ):
-            out('memcpy((void *)&%s, (%s){%s}, sizeof %s);\n'
-                % (dest.read(),
-                   TArray(typ, mkIntegerLiteral(site, 1)).declaration(''),
-                   mkCast(site, self.expr, typ).read(),
-                   dest.read()))
+        rt = safe_realtype_shallow(typ)
+        # There is a reasonable implementation for this case (memcpy), but it
+        # never occurs today
+        assert not isinstance(typ, TArray)
```
lwaern-intel, line 4912: Sloppy. should use `rt` instead, as should `isinstance(typ, TExternStruct)`.

py/dml/dmlparse.py
mandolaerik: I would argue the opposite wrt fragility: your current approach requires that any new rule that uses `ident` remembers to call `ident_enforce_not_discard_ref`, if not the deprecation will silently be incomplete.

But we probably just misunderstand each other. More efficient if I try to code up what I mean, perhaps I will understand better when I see the R/R conflicts myself.

mandolaerik: I made an attempt, now pushed to this PR.

test/1.4/expressions/T_discard_ref.dml
```diff
@@ -5,6 +5,8 @@
 dml 1.4;
 device test;
 
+/// DMLC-FLAG --no-compat=discard_ref_shadowing
```
mandolaerik, line 8: Without this, the test passed in t126 before I added objident_or_underscore.

However, I suppose the error would still have been captured in the 7 build, so one may argue that adding the flag reduces test coverage in CI (the file is never compiled *without* the flag). So perhaps this commit should be dropped.

syssimics:
Verification [#12518623](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12518623): fail

syssimics:
Verification [#12518616](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12518616): pass

py/dml/dmlparse.py
```diff
@@ -2670,7 +2670,9 @@ def reserved_(t):
 @prod_dml14
 @lex.TOKEN(ident_rule('ident', ['_']))
 def ident_underscore(t):
-    if compat.discard_ref_shadowing not in dml.globals.enabled_compat:
+    if (compat.discard_ref_shadowing not in dml.globals.enabled_compat
+        # forgive the `param _ auto` declaration
+        and not site(t).filename().endswith('dml-builtins.dml'):
```
lwaern-intel, line 2675: we don't have a `param _ auto` declarqtion!

py/dml/dmlparse.py
```diff
@@ -2674,15 +2685,27 @@ def ident_enforce_not_discard_ref(site, ident):
...
 def reserved(t):
     raise ESYNTAX(site(t, 1), str(t[1]), "reserved word")
 
+@prod_dml14
+@lex.TOKEN(ident_rule('ident_or_underscore', reserved_words_14))
+def reserved_(t):
+    raise ESYNTAX(site(t, 1), str(t[1]), "reserved word")
+
+@prod_dml14
+@lex.TOKEN(ident_rule('ident', ['_']))
```
lwaern-intel, line 2671: just do
```
ident : _
```
also in my WIP I called the token `DISCARDREF`.

py/dml/dmlparse.py
lwaern-intel: also do please actually go through the effort to build and test your changes before pushing them to to another's branch. That's just rude.

py/dml/dmlparse.py
mandolaerik: `_` follows the convention that identifier-formed tokens use that as their name.

mandolaerik: And thanks for spotting my stupid copy-paste

py/dml/codegen.py
mandolaerik: I was thinking of the code in dmlparse that checks enabled_compat, but I suppose compat.py makes even more sense.

py/dml/dmlparse.py
mandolaerik: oh, sorry, I was so sure we had that I didn't even bother to check. So the top commit is rubbish, now removed.

syssimics:
Verification [#12598423](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12598423): fail

doc/1.4/language.md
```diff
@@ -3098,6 +3107,29 @@ method m() -> (bool, int) {
 local (bool a, int i) = m();
 ```
 
+
+"\_" may be used as an identifier for local variables, as well as other
+method-local bindings such as the method parameters, and the bound identifier
+in `foreach`/`#foreach`/`#select` statements, and message component paramters of
```
mandolaerik, line 3113: s/paramters/parameters

mandolaerik (approved):
Looks good. Two remarks:
* I would vote for also adding struct memebers named `_`. This can just as well be added in a separate PR, though.
* Looks like you preserve old behaviour in 1.2, thus introducing a discrepancy between 1.2 and 1.4. Doing the same in 1.2 and 1.4 would possibly give less code, but OTOH this doesn't really matter (dying code) and the code is already written.

syssimics:
[DML Verification : 6](https://github.com/intel-innersource/applications.simulators.simics.infra.ci-tools/actions/runs/13628169858): ✅ success

syssimics:
[DML Verification : 6](https://github.com/intel-innersource/applications.simulators.simics.infra.ci-tools/actions/runs/13628167973): ✅ success

syssimics:
[DML Verification : 6](https://github.com/intel-innersource/applications.simulators.simics.infra.ci-tools/actions/runs/13628323674): ✅ success

syssimics:
[DML Verification : 6](https://github.com/intel-innersource/applications.simulators.simics.infra.ci-tools/actions/runs/13628323635): ✅ success

lwaern-intel:
@mandolaerik [Verification passes](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/13628321989), this is ready for review.

lwaern-intel:
[Jenkins run](https://cje-fm-owrp-prod06.devtools.intel.com/satg-simics-main/job/sre-vp/job/_pretest/203396). No failures seem related.

doc/1.4/language.md
```diff
@@ -1505,6 +1516,24 @@ integer members (and arrays of similar) are translated
 to endian integers (or arrays of such) of similar size,
 with endianness matching the layout. Layout and endian integer
 members are accessed normally.
+
+The *discard identifer* "\_" may be used as the name of any number of members
```
mandolaerik, line 1520: Inconsistent naming "\_" vs `_`, seems that you prefer the former in the context of declarations and the latter in the context of lvalues? why?

lwaern-intel: i was inconsistent because i was inconsistent. and cause i wrote the docs for the discard identifier after i wrote them for the discard ref.

Really, using "`_`" should always be preferred, as depending on the setting a ticked underscore may not appear much different from surrounding text.

mandolaerik: I think plain `_` should be ok. If the user cares about seeing differences between identifiers and other words, then they should set things up accordingly.

doc/1.4/language.md
```diff
@@ -3175,6 +3204,24 @@ local (bool a, int i) = m();
 In the absence of explicit initializer expressions, a default
 "all zero" initializer will be applied to each declared object.
 
+The *discard identifier* "\_" may be used as an identifier for local variables,
+as well as other method-local bindings such as the method parameters, the bound
+identifier in `foreach`/`#foreach`/`#select` statements, and message component
+parameters of [hook-bound after statements](#hook-bound-after-statements).
+Any method-local binding named "\_" *will not be added to scope.* This is useful
```
mandolaerik: nitpick `s/.*/*.`, no need to fix.

PR #239, opened 2023-11-14
Store all permutations in a global shared array, so that one 32-bit index per register is sufficient.

mandolaerik:
- remove unused method
- Avoid re-calculating read/write field order
- Store field order as permutation indices instead of field references
- Store all permutations in a global shared array, so that one 32-bit index per register is sufficient.

### DIFF
--- a/include/simics/dmllib.h
+++ b/include/simics/dmllib.h
@@ -3129,9 +3129,91 @@ _DML_get_single_hook_attr(void *_hook, uintptr_t _aux) {
                                      &hook->queue);
 }
 
-
 UNUSED static inline uint64 _identity_to_key(_identity_t id) {
         return (uint64)id.id << 32 | (uint64)id.encoded_index;
+}
+
+typedef struct {
+        ht_int_table_t id_to_idx_read;
+        ht_int_table_t id_to_idx_write;
+        ht_str_table_t perm_set;
+        VECT(uint8) table;
+} _permutation_helper_t;
+
+UNUSED static _permutation_helper_t *
+_DML_new_permutation_helper(void)
+{
+        _permutation_helper_t *ret = MM_MALLOC(1, _permutation_helper_t);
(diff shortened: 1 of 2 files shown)


syssimics:
Verification [#12507032](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12507032): fail

mandolaerik:
This is yet unpolished, and meant as a basis for discussions (but at least 1.4/lib/fields passes).
* does this approach make sense, or is it a dead end?
* The memoization of `_get_read_fields` adds some ugliness -- just look at `_cleanup_permutation_table_helper` -- we can consider skipping this and keep hash tables instead, doing an identity_t lookup every time. Probably somewhat slower in runtime, and may use more or less mem, but the diff should not be dramatic.

syssimics:
Verification [#12507039](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12507039): fail

lwaern-intel (changes requested):
I think this approach is... though not _pretty_, it is fine when adjusted. My big peeve is `_permutation_table_helper()`, but that's fixable (see comment.)

The guarantee we offer that the implicit calls of independent startups happen before independent startup memoized was to offer a degree of control when startup memoizeds had dependencies, so it's not like we're operating outside their boundary of what they were intended for (_except_ `_permutation_table_helper()`, which, again, is fixable.)

lwaern-intel:
We have segfaults, by the way! This may or may not be due to the thing I complained about, regardless, you'll want to track them down.

johanhogberg (approved):
Great work!
We e-mailed a new performance test which we didn't have earlier.
When we started with granite rapids we saw that it took very long time to enable-probes and we got down the time from over a minute to about 10 seconds.

We think that your refactoring actually should be slightly faster, but lets find out first.

mandolaerik:
This PR does demonstrate an interesting memoization technique, but after #243 it is no longer needed.

PR #240, opened 2023-11-15
Add DMLC option to not suppress GCC's Wunused for local variables -- SIMICS-21585

lwaern-intel:
Implemented clumsily in a way that doesn't necessarily only affect local variables.
This PR is more of a way to investigate the impact this has, and serve as a talking
point. We likely want to report unused variables in DMLC directly instead.

Depends on #237


syssimics:
Verification [#12510790](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12510790): pass

PR #241, opened 2023-11-17
Fix VP issues

lwaern-intel:


py/dml/ctree.py
```diff
@@ -4707,6 +4712,10 @@ def ctype(self):
     def read(self):
         return self.expr.read()
     def discard(self): pass
+    def incref(self):
+        self.expr.incref()
+    def decref(self):
+        self.expr.decref()
```
lwaern-intel, line 4718: This kinda stayed. It lack of these implementations _were_ a bug, but it doesn't really matter now.

syssimics:
Verification [#12524458](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12524458): pass

PR #243, opened 2023-11-24
Use counting sort to sort in linear instead of quadratic time

mandolaerik:
- remove unused method
- Use counting sort to sort in linear instead of quadratic time

### DIFF
--- a/lib/1.4/dml-builtins.dml
+++ b/lib/1.4/dml-builtins.dml
@@ -2825,33 +2825,6 @@ template register is (_conf_attribute, get, set, shown_desc,
 
     // fills in the passed array with all fields in this register, and returns
     // the number of elements and a bitmask showing bits not covered by fields.
-    shared method _get_all_fields(field *fields) -> (int, uint64) /* n, unmapped */ {
-        local uint64 unmapped;
-        unmapped[this.bitsize-1:0] = -1;
-        local int n = 0;
-        // TODO: it is inefficient to re-generate the field list each call
-        foreach f in (this.fields) {
-            local int lsb = f.lsb;
-            // sort 'fields' by lsb
-            local int i;
-            for (i = n; i > 0; --i) {
-                local int next_lsb = fields[i - 1].lsb;
-                if (next_lsb > lsb) {
-                    fields[i] = fields[i - 1];
-                } else {
-                    break;
-                }
-            }
-            fields[i] = f;
-            unmapped[lsb + f.bitsize - 1 : lsb] = 0;
-            ++n;
-        }
-        assert n <= 64;
-        return (n, unmapped);
-    }
-
-    // fills in the passed array with all fields in this register, and returns
-    // the number of elements and a bitmask showing bits not covered by fields.
     shared method _get_get_fields(_get_field *fields) -> (int, uint64) /* n, unmapped */ {
         local uint64 unmapped;
         unmapped[this.bitsize-1:0] = -1;
--- a/lib/1.4/dml-builtins.dml
+++ b/lib/1.4/dml-builtins.dml
@@ -21,6 +21,7 @@ import "simics/model-iface/register-view.dml";
 import "simics/model-iface/register-view-read-only.dml";
 import "simics/model-iface/transaction.dml";
 import "simics/util/hashtab.dml";
+import "simics/util/bitcount.dml";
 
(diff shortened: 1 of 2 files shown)


mandolaerik:
This is a simpler alternative to #239: almost as fast (4ns instead of 2ns per field for getting the sorted fields), and no memory footprint. Thanks to Love for suggesting this approach.

syssimics:
Verification [#12567253](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12567253): fail

syssimics:
Verification [#12570225](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12570225): pass

lib/1.4/dml-builtins.dml
```diff
@@ -2825,136 +2826,89 @@ template register is (_conf_attribute, get, set, shown_desc,
...
-            for (i = n; i > 0; --i) {
-                local int next_lsb = fields[i - 1].lsb;
-                if (next_lsb > lsb) {
-                    fields[i] = fields[i - 1];
-                } else {
-                    break;
-                }
-            }
-            fields[i] = f;
+            lsbs[lsb] = 1;
```
lwaern-intel, line 2860: note the microoptimization:
```C
lsbs |= (1 << lsb);
```
I'm actually going to speak in favor of it since the entire point of this PR is to optimize, and it doesn't really change anything in clarity.

lwaern-intel: You can, however, disregard it if it makes no difference on your benches.

lib/1.4/dml-builtins.dml
```diff
@@ -2825,136 +2826,89 @@ template register is (_conf_attribute, get, set, shown_desc,
...
             unmapped[lsb + f.bitsize - 1 : lsb] = 0;
-            ++n;
+            n += 1;
```
lwaern-intel, line 2862: Instead of accumulating `n` in a loop, use `_set_fields.len`, instead OR `bit_count64(lsbs)`. The latter is only a single cycle, rather than some cycles per number of fields. Again, since the entire point of this PR is to optimize, I'm going to speak in favor of it unless you can show through benches that it makes no difference.

lib/1.4/dml-builtins.dml
lwaern-intel: Wow! Turns out i don't understand crap and compilers do whatever the hell they want!

lib/1.4/dml-builtins.dml
lwaern-intel: Nope! Accumulating `n` is faster! Slightly faster than returning `bit_count64(lsbs)`, and much faster than `_set_fields.len`. My guess is that it gets ILP:d away.

lwaern-intel (approved):
see my comments for my follies. Microoptimization is hard, who knew.

lib/1.4/dml-builtins.dml
lwaern-intel: Surprisingly `n++` is about as fast as `n += 1` even though the generation of `n += 1` takes the address of `n` and `n++` doesn't.

lwaern-intel: Oo! If you get replace `n += 1` with `++n` you'll cut down on the diff since `++n` was the original code!

lwaern-intel: Yeah remove this diff

lib/1.4/dml-builtins.dml
```diff
@@ -2825,136 +2826,89 @@ template register is (_conf_attribute, get, set, shown_desc,
...
```
lwaern-intel, line 2884: Same here, reestablish `++n;`

lib/1.4/dml-builtins.dml
```diff
@@ -2825,136 +2826,89 @@ template register is (_conf_attribute, get, set, shown_desc,
...
```
lwaern-intel, line 2906: and here

lwaern-intel (changes requested):

lwaern-intel (approved):

syssimics:
Verification [#12592898](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12592898): pass

PR #244, opened 2023-11-30
Don't test reverse execution in Simics 7

aragners:
The reverse exectuion feature is getting removed in 7.

### DIFF
--- a/test/1.4/lib/T_signal_templates.py
+++ b/test/1.4/lib/T_signal_templates.py
@@ -53,12 +53,14 @@ stest.expect_equal(obj.outsig_signal_high, True)
 obj.outsig = stub
 stest.expect_equal(stub.level, 1)
 
-# reverse execution doesn't cause signal pulsing
-transitions = stub.transitions
-simics.SIM_run_command('enable-reverse-execution')
-simics.SIM_run_command('continue 1000')
-simics.SIM_run_command('reverse')
-stest.expect_equal(stub.transitions, transitions)
+# Reverse execution is removed in Simics 7
+if 'rev-execution' in simics.SIM_get_all_classes():
+    # reverse execution doesn't cause signal pulsing
+    transitions = stub.transitions
+    simics.SIM_run_command('enable-reverse-execution')
+    simics.SIM_run_command('continue 1000')
+    simics.SIM_run_command('reverse')
+    stest.expect_equal(stub.transitions, transitions)
 
 obj.test_outsig = False
 stest.expect_equal(stub.level, 0)


test/1.4/lib/T_signal_templates.py
```diff
@@ -53,12 +53,14 @@ def signal_lower(self):
...
 stest.expect_equal(stub.level, 1)
 
-# reverse execution doesn't cause signal pulsing
-transitions = stub.transitions
-simics.SIM_run_command('enable-reverse-execution')
-simics.SIM_run_command('continue 1000')
-simics.SIM_run_command('reverse')
-stest.expect_equal(stub.transitions, transitions)
+# Reverse execution is removed in Simics 7
+if int(simics.SIM_version_major()) < 7:
```
lwaern-intel, line 57: This'll crash if the version in play is 4.8, won't it? Not that I expect that ever happens, but still, for the sake of principle I'd suggest:
`if simics.SIM_version_major() == '4.8' or int(simics.SIM_version_major()) < 7:`

lwaern-intel (changes requested):

test/1.4/lib/T_signal_templates.py
aragners: Would  `int(simics.SIM_version_major().split(".")[0]) < 7` be okay? Less clean but works with outdated Simics versions.

aragners: When I think more about it I think DML 1.4 was added in 6, maybe with some tech preview in 5. I don't think it's worth adding the extra ugliness for supporting this test to be allowed to run in old Simics versions with "x.y" major versions.

lwaern-intel: DMLC in principle still supports 4.8. But yes, this is nit-picky, and I only brought it out of principle's sake. It's not a blocker.

lwaern-intel: Actually... rather, what we support is _API_ version 4.8. I'm not sure what majors of Simics proper we support. This makes this more shaky.

aragners: I changed the check to see if rev-execution class is available or not.

syssimics:
Verification [#12603417](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12603417): fail

syssimics:
Verification [#12603867](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12603867): pass

lwaern-intel (approved):
The only concern was if this was somehow very slow, but it's not.

test/1.4/lib/T_signal_templates.py
```diff
@@ -53,12 +53,14 @@ def signal_lower(self):
...
-simics.SIM_run_command('reverse')
-stest.expect_equal(stub.transitions, transitions)
+# Reverse execution is removed in Simics 7
+if 'rev-execution' in simics.SIM_get_all_classes():
+    # reverse execution doesn't cause signal pulsing
+    transitions = stub.transitions
+    simics.SIM_run_command('enable-reverse-execution')
+    simics.SIM_run_command('continue 1000')
+    simics.SIM_run_command('reverse')
+    stest.expect_equal(stub.transitions, transitions)
```
mandolaerik, line 63: Better to rewrite the test to use the 7 equivalent of revexec (snapshots, right?)

FWIW, I think the purpose of the test is to exercise the following logic in `signal_connect.set`, in `1.4/utility.dml`:
```
        local bool hotplug = SIM_object_is_configured(dev.obj)
            && !SIM_is_restoring_state(dev.obj);
```

aragners: I agree, but as snapshots API is not finalized yet I think we can wait for that. Posted https://jira.devtools.intel.com/browse/SIMICS-21789 to fix this.

PR #245, opened 2023-12-01
Resolved paths in generated C-file's sources list

TSonono:
I did not write any tests for this since I'm not familiar with the repo. I tested it by "manually at least". Let me know if I should change something


MagnusIntel (approved):

py/dml/c_backend.py
```diff
@@ -3247,7 +3247,7 @@ def generate_cfile(device, footers,
         ' * Generated by dmlc, do not edit!',
         ' *',
         ' * Source files:',
-        '\n'.join([' *   ' + f for f in source_files]),
+        '\n'.join([' *   ' + str(Path(f).resolve()) for f in source_files]),
```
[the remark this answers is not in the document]
TSonono: Thanks for the suggestion, fixed

lwaern-intel (changes requested):
As stated in R.C., the `dead_dml_methods` script should be changed to assert on that the path is absolute. Something like:
```python
def dml_sources(c_file: Path) -> set[Path]:
    '''Given a DMLC-generated C file, return the set of DML files that
    were used to generate it'''
    def paths():
        for p in dml_sources_from_body(c_file.read_text()):
            p = Path(p)
            assert p.is_absolute(), p
            yield p
    return {p for p in paths()}
```

You're welcome to shorten that. Building a set by leveraging `.add()` is OK, I just wrote a generator adjustor for principle's sake.

Also, to test that the change and assertion works, modify the import in `tests/1.4/misc/dead_methods.dml` to
`import "../misc/dead_methods_imported.dml";`, with a comment stating why the import's like that (test that DMLC resolves the path in the source lists).

py/dead_dml_methods.py
```diff
@@ -56,10 +56,14 @@ def dml_sources_from_body(body):
...
 def dml_sources(c_file: Path) -> set[Path]:
-    '''Given a DMLC-generated C file, return the set of DML files that
-    were used to generate it'''
-    return {(c_file.parent / Path(p)).resolve()
-            for p in dml_sources_from_body(c_file.read_text())}
+    """Given a DMLC-generated C file, return the set of DML files that
+    were used to generate it"""
+    def paths():
+        for p in dml_sources_from_body(c_file.read_text()):
+            p = Path(p)
```
lwaern-intel, line 63: Cleaner:
```python
        for p in map(Path, dml_sources_from_body(c_file.read_text())):
            assert p.is_absolute(), f'{p} is not an absolute path'
            yield p
```

(thanks Erik!)

lwaern-intel (changes requested):

py/dead_dml_methods.py
TSonono: Nice

syssimics:
Verification [#12619624](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12619624): pass

lwaern-intel (approved):
👍

PR #249, opened 2024-01-02
Remove end-year test

mandolaerik:
I suppose this should be fine, since the rest of Simics no longer requires
the end-year to be up to date.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-# © 2021-2024 Intel Corporation
+# © 2021 Intel Corporation
 # SPDX-License-Identifier: MPL-2.0
 
 #
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <!--
-  © 2021-2024 Intel Corporation
+  © 2021 Intel Corporation
   SPDX-License-Identifier: MPL-2.0
 -->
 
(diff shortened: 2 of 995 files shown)


syssimics:
Verification [#12769216](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12769216): pass

andreashedstrom-intel (commented):
Not sure if this is the correct change or not. The DML repo has already been updated with the correct year on all files (not sure if that was the correct play either but it was the only thing we could do by the people working that day and it can easily be reverted).

The Base repo did not skip the check, they changed format from NNNN-MMMM to just MMMM so perhaps you want to do this too?

See base repo commit ace8e9a9d6e6e0c9bbdc1559270db63179feb492 for more details.

mandolaerik:
> Not sure if this is the correct change or not. The DML repo has already been updated with the correct year on all files (not sure if that was the correct play either but it was the only thing we could do by the people working that day and it can easily be reverted).

Great, my first patch was an emergency quick-fix but that's not needed then.
 
> The Base repo did not skip the check, they changed format from NNNN-MMMM to just MMMM so perhaps you want to do this too?

Done now.

syssimics:
Verification [#12797586](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12797586): fail

andreashedstrom-intel (approved):

PR #251, opened 2024-01-17
update ignorelist

mandolaerik:


syssimics:
Verification [#12841542](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12841542): pass

PR #252, opened 2024-01-23
fix typo

mandolaerik:


lwaern-intel (approved):
If you're sure this is the only one in the section ;)

PR #254, opened 2024-01-25
Make the suppression of `WLOGMIXUP` a proper compatibility feature

lwaern-intel:
@MagnusIntel Adding you as reviewer as I'm interested whether you think the documentation for `suppress_WLOGMIXUP` is sufficient or needs something more.

### DIFF
--- a/RELEASENOTES.docu
+++ b/RELEASENOTES.docu
@@ -1748,4 +1748,13 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
       are themselves constant inlined method parameters (i.e. an inline method
       call propagates an inline parameter to an inline method call of its own.)
   </add-note></build-id>
+  <build-id _6="next" _7="next"><add-note> The DMLC behaviour of suppressing the
+      warning <tt>WLOGMIXUP</tt> by default below Simics API version 7 is now
+      controlled by the compatibility feature <tt>suppress_WLOGMIXUP</tt>.
+      This is in order to increase visibility of the API-dependent behaviour,
+      and to provide migration documentation about it.
+      As a consequence, <tt>WLOGMIXUP</tt> can be enabled below Simics API
+      version 7 by passing either <tt>--warn=WLOGMIXUP</tt> (as before) or
+      <tt>--no-compat=suppress_WLOGMIXUP</tt> to DMLC.</add-note>
+  </build-id>
 </rn>
--- a/py/dml/compat.py
+++ b/py/dml/compat.py
@@ -119,6 +119,32 @@ class shared_logs_on_device(CompatFeature):
 
 
 @feature
+class suppress_WLOGMIXUP(CompatFeature):
+    '''This compatibility feature makes it so the warning `WLOGMIXUP` is
+    suppressed by default. `WLOGMIXUP` warns about usages of a common faulty
+    pattern which results in broken log statements &mdash; for more
+    information, see the documentation of `WLOGMIXUP` in the
+    [Messages](messages.html) section.
+
+    `WLOGMIXUP` is suppressed by default below Simics API version 7 in order
+    to avoid overwhelming users with warnings, as the faulty pattern that
+    `WLOGMIXUP` reports is very prevalent within existing code. Addressing
+    applications of the faulty pattern should be done before or as part of
+    migration to Simics API version 7.
+
+    Passing `--no-compat=suppress_WLOGMIXUP` to DMLC has almost the same effect
+    as passing `--warn=WLOGMIXUP`; either will cause DMLC to report the warning
+    even when the Simics API version in use is below 7. The only difference
+    between these two options is that if `--no-compat=suppress_WLOGMIXUP` is
+    used (and `--warn=WLOGMIXUP` is not), then `WLOGMIXUP` may still be
+    explicitly suppressed via `--no-warn=WLOGMIXUP`. In contrast,
+    `--warn=WLOGMIXUP` doesn't allow for `WLOGMIXUP` to be suppressed at
+    all.'''
+    short = "Suppress the warning 'WLOGMIXUP' by default"
+    last_api_version = api_6
+
+
+@feature
 class dml12_inline(CompatFeature):
     '''When using `inline` to inline a method in a DML 1.2 device,
     constant arguments passed in typed parameters are inlined as
--- a/lib/1.2/dml-builtins.dml
+++ b/lib/1.2/dml-builtins.dml
@@ -211,6 +211,7 @@ template device {
     parameter _compat_port_obj_param auto;
     parameter _compat_io_memory auto;
     parameter _compat_shared_logs_on_device auto;
+    parameter _compat_suppress_WLOGMIXUP auto;
     parameter _compat_dml12_inline auto;
     parameter _compat_dml12_not auto;
     parameter _compat_dml12_goto auto;
--- a/lib/1.4/dml-builtins.dml
+++ b/lib/1.4/dml-builtins.dml
@@ -546,6 +546,7 @@ template device {
     param _compat_port_obj_param auto;
     param _compat_io_memory auto;
     param _compat_shared_logs_on_device auto;
+    param _compat_suppress_WLOGMIXUP auto;
     param _compat_dml12_inline auto;
     param _compat_dml12_not auto;
     param _compat_dml12_goto auto;
--- a/py/dml/dmlc.py
+++ b/py/dml/dmlc.py
@@ -534,27 +534,6 @@ def main(argv):
         parser.error("dmlc: the -m option is only valid together with --api=4.8"
                      " or older")
 
-    # This warning is disabled by default below Simics 7 due to sheer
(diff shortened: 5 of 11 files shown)


py/dml/compat.py
```diff
@@ -118,6 +118,23 @@ class shared_logs_on_device(CompatFeature):
...
+    '''This compatibility feature makes it so the warning `WLOGMIXUP` is
+    suppressed by default. `WLOGMIXUP` warns about usages of a common faulty
+    pattern which results in broken log statements &mdash; for more
+    information, see the documentation of `WLOGMIXUP` in the
+    [Messages](messages.html) section of the DML 1.4 Reference Manual.
+
+    `WLOGMIXUP` is suppressed by default below Simics API version 7 in order
+    to avoid overwhelming users with warnings, as the faulty pattern that
+    `WLOGMIXUP` reports is very prevalent within existing code. Addressing
+    applications of the faulty pattern should be done as part of migration to
```
lwaern-intel, line 132: before or as part of

syssimics:
Verification [#12885718](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12885718): pass

MagnusIntel (approved):

MagnusIntel:
This piece of the release note
> As a consequence, WLOGMIXUP can be enabled below Simics API version 7 by passing either --warn=WLOGMIXUP or --no-compat=suppress_WLOGMIXUP to DMLC.
would actually be nice to have in https://github.com/intel/device-modeling-language/wiki/B.-Managing-deprecated-language-features

RELEASENOTES.docu
```diff
@@ -1748,4 +1748,12 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
       are themselves constant inlined method parameters (i.e. an inline method
       call propagates an inline parameter to an inline method call of its own.)
   </add-note></build-id>
+  <build-id _6="next" _7="next"><add-note> The behaviour of suppressing the
+      warning <tt>WLOGMIXUP</tt> by default below Simics API version 7 is now
+      controlled by the compatibility feature <tt>suppress_WLOGMIXUP</tt>
+      rather than being built into DMLC. This is in order to increase visibility
```
mandolaerik, line 1754: Two notes, neither blocks a merge:

* "rather than": The compat feature _is_ still built into DMLC
* The text can be significantly shortened, since the change itself is not so significant to an end-user. I suppose it would suffice with smth like "added a compat feature that controls the default setting of WLOGMIXUP. This was mainly done to raise the visibility of its API-dependent behaviour".

mandolaerik (approved):

py/dml/compat.py
```diff
@@ -118,6 +118,23 @@ class shared_logs_on_device(CompatFeature):
     last_api_version = api_6
 
 
+@feature
+class suppress_WLOGMIXUP(CompatFeature):
+    '''This compatibility feature makes it so the warning `WLOGMIXUP` is
+    suppressed by default. `WLOGMIXUP` warns about usages of a common faulty
+    pattern which results in broken log statements &mdash; for more
+    information, see the documentation of `WLOGMIXUP` in the
+    [Messages](messages.html) section of the DML 1.4 Reference Manual.
```
mandolaerik, line 127: "of the ... manual" can be omitted, since this text is within an appendix in the same manual.

py/dml/compat.py
```diff
@@ -118,6 +118,23 @@ class shared_logs_on_device(CompatFeature):
...
+
+    `WLOGMIXUP` is suppressed by default below Simics API version 7 in order
+    to avoid overwhelming users with warnings, as the faulty pattern that
+    `WLOGMIXUP` reports is very prevalent within existing code. Addressing
+    applications of the faulty pattern should be done before or as part of
+    migration to Simics API version 7.'''
```
mandolaerik, line 133: @MagnusIntel so your suggestion is to mention `--no-warn=WLOGMIXUP` here?

RELEASENOTES.docu
lwaern-intel: I don't really agree with point (2). In particular, I want to keep the sentence "This is in order to increase visibility of that API-dependent behaviour, and to provide migration documentation about it" completely. This will draw the attention of those who are undergoing migration efforts.
The "as a consequence" part could be stripped away, I suppose; I added it to provide more clarity about the behavior. In particular, I wanted to reassure people that were already aware of `WLOGMIXUP` that "don't worry, `--warn=WLOGMIXUP` will still work as before, but now `--no-compat` is also an alternative."

Point 1: I could strip everything past "rather than", but I want to be clear that the mentioned behavior is not new. Maybe just rewrite it to be more accurate: "... rather than being implemented ad-hoc"?

py/dml/compat.py
lwaern-intel: Sure.

py/dml/compat.py
lwaern-intel: I'm not sure. He could mean that it should mention that `--warn=WLOGMIXUP` can be used as an alternative to `--no-compat=suppress_WLOGMIXUP`. @MagnusIntel you'll have to clarify what you meant.

If so, it could be worth to also note there are some subtle differences between the two; `--warn=WLOGMIXUP` guarantees that the warning gets reported no matter what, while `--no-compat=suppress_WLOGMIXUP` just disables the _default_ suppression, but still allows for the warning to get suppressed explicitly via `--no-warn`.
I.e. `--warn=WLOGMIXUP --no-warn=WLOGMIXUP` will result in `WLOGMIXUP` being reported; in contrast, `--no-compat=suppress_WLOGMIXUP --no-warn=WLOGMIXUP` results in `WLOGMIXUP` being suppressed.

RELEASENOTES.docu
lwaern-intel: Actually point 1 can be addressed by tweaking the final statement of the releasenote.

lwaern-intel: Done. Is the newer one better?

py/dml/compat.py
mandolaerik: I don't think we need to document these subtleties in detail, but there is a point in cross-referencing the corresponding --warn flag as an alternative. People will tweak flags until it works, and the migration logic will likely stay stable until it's deprecated out so whatever tweaks they do will continue to work.

RELEASENOTES.docu
mandolaerik: I will drop this debate. If you can find formulations that *you* think are improvements based on by comments then please do, but the current note is also good enough.

lwaern-intel: well aright

py/dml/compat.py
MagnusIntel: I think --warn=<TAG> is drastically easier to understand than --no-compat=suppress_<TAG>. And if I got the latter flags wrong, it just shows that the --no-compat flag is incomprehensible or unnecessarily hard to understand.

lwaern-intel: So your complaint is rather that `--no-compat`, and compatibility features in general (like their naming scheme describing the legacy behavior rather than what the effect of disabling it is) is difficult to understand in general? 
I think both me and Erik agree... we've discussed previously about how we should present compatibility features, and if we should present them as "features" at all.
If a particular point of confusion is just the name of the `--no-compat` option, would it help if the option was called `--disable-compat` instead?

Regardless, addressing the presentation of `--no-compat` is orthogonal to this PR.

I can add in the documentation that `--warn=WLOGMIXUP` can serve as an alternative to `--no-compat=suppress_WLOGMIXUP`, but that's **only** applicable to `suppress_WLOGMIXUP`, as no other compatibility feature is about warnings. We are not going to add `--no-compat=suppress_<TAG>` in general (though that naming may become a pattern for future compatibility features, if we later want to introduce other warnings that we want to bind to API versions.)

py/dml/compat.py
```diff
@@ -130,7 +130,16 @@ class suppress_WLOGMIXUP(CompatFeature):
...
+    migration to Simics API version 7.
+
+    Passing `--no-compat=suppress_WLOGMIXUP` to DMLC has almost the same effect
+    as passing `--warn=WLOGMIXUP`; either will cause DMLC to report the warning
+    even when the Simics API version in use is below 7. The only difference
+    between these two options is that if `--no-compat=suppress_WLOGMIXUP` is
+    used (and `--warn=WLOGMIXUP` is not), then `WLOGMIXUP` may still be
+    explicitly suppressed via `--no-warn=WLOGMIXUP`. In contrast,
+    `--warn=WLOGMIXUP` doesn't allow for `WLOGMIXUP` to be suppressed at
+    all.'''
```
lwaern-intel, line 142: @MagnusIntel @mandolaerik Does this look fine?

MagnusIntel: I find it unlikely that anyone would consume such fine details, but at least it is complete

MagnusIntel (approved):

syssimics:
Verification [#12892647](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12892647): pass

PR #256, opened 2024-01-31
Make test work with sym-linked sources

rogue4242:
More recent MinGW versions resolve symlinks when embedded source references in dependecy files.
Hence, the test should also ensure it always uses the resolved locations for both source and target (older MinGW might still not resolve locations, so only resolving one of the two would then confuse these).

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -1203,7 +1203,7 @@ class DmlDep(DmlDepBase):
             join(testdir, '1.2', 'misc', 'rel', 'a', 'x.h'),
             join(testdir, '1.2', 'misc', 'rel', 'a', 'y.h')]
         for exp in expected_subset:
-            if all(os.path.normpath(p) != os.path.normpath(exp)
+            if all(Path(p).resolve() != Path(exp).resolve()
                    for p in prereqs):
                 raise TestFail('missing prerequisite: ' + exp)
 


test/tests.py
```diff
@@ -1203,7 +1203,8 @@ def test(self):
             join(testdir, '1.2', 'misc', 'rel', 'a', 'x.h'),
             join(testdir, '1.2', 'misc', 'rel', 'a', 'y.h')]
         for exp in expected_subset:
-            if all(os.path.normpath(p) != os.path.normpath(exp)
+            if all(os.path.normpath(os.path.realpath(p))
+                   != os.path.normpath(os.path.realpath(exp))
```
lwaern-intel, line 1207: Is `.normpath` even needed with `.realpath`?
Also, we're moving away `os.path.realpath` to `Path(p).resolve()`. If realpath is sufficient, please use `Path(p).resolve()` instead.

lwaern-intel (changes requested):

test/tests.py
rogue4242: Honestly, I do not know if `normpath` is still needed. I kept it in a "better safe than sorry" mindset. I'll test with `Path(p).resolve()`

rogue4242: Changed. Test succeeds.

lwaern-intel (approved):
Looks good. Gonna run validation and then merge.

syssimics:
Verification [#12917316](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12917316): pass

PR #257, opened 2024-02-05
Use project mini-python

jhbaarnh:
This is in preparation for moving Python to an addon package.

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -65,8 +65,8 @@ else:
 bat_sfx = batch_suffix()
 exe_sfx = '.exe' if is_windows() else ""
 
-mini_python = [join(simics_base_path(), host_type(), "bin",
-                    "mini-python" + exe_sfx)]
+mini_python = [join(testparams.project_path(), "bin",
+                    f"mini-python{batch_suffix()}")]
 dmlc_py = [join(project_host_path(), "bin", "dml", "python")]
 main_dmlc = mini_python + dmlc_py
 # Alternative python interpreter for DMLC including args,


syssimics:
Verification [#12943968](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12943968): pass

lwaern-intel:
Hmm... this breaks the DML job because it works by _not_ leveraging a Simics repo, but by using prebuilt packages. In order for the DML job to work, we'd need to patch tests.py in order to leverage that package (and then @andreashedstrom-intel can update the DML job to ensure that package is available). @mandolaerik What should be the plan of action, here? Is there any way we can reformulate this patch such that the DML job can be updated to work without needing any future changes in the DMLC repo? (And ideally, in such a way that would allow the DML job to be updated to be functional immediately; however, it's OK if it needs to wait until the addon package becomes available.)

test/tests.py
```diff
@@ -65,8 +65,7 @@ def active_test(fullname):
 bat_sfx = batch_suffix()
 exe_sfx = '.exe' if is_windows() else ""
 
-mini_python = [join(simics_base_path(), host_type(), "bin",
-                    "mini-python" + exe_sfx)]
+mini_python = [join(simics_repo_root(), "bin", f"mini-python{batch_suffix()}")]
```
mandolaerik, line 68: simics_repo_root is only available when running from the  base repo, and we want the test to be runnable from a normal project. So better to use `testparams.project_path` here. Also, we should check if this change gives a significant performance penalty on Windows; if so, we should consider adding something to t126 that makes it use mini-python.exe directly.

mandolaerik (changes requested):

lwaern-intel (approved):
Oh hey that just made the DML job Just Work, huh. That's great.
Approval is pending verification (not that I see why it would fail.)

test/tests.py
```diff
@@ -65,8 +65,7 @@ def active_test(fullname):
 bat_sfx = batch_suffix()
 exe_sfx = '.exe' if is_windows() else ""
 
-mini_python = [join(simics_base_path(), host_type(), "bin",
-                    "mini-python" + exe_sfx)]
+mini_python = [join(project_path(), "bin", f"mini-python{batch_suffix()}")]
```
lwaern-intel, line 68: Actually, either use `testparams.project_path`, or update existing usages of  `testparams.project_path` to use `project_path`

lwaern-intel (changes requested):

syssimics:
Verification [#12947213](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12947213): pass

test/tests.py
```diff
@@ -65,8 +65,8 @@ def active_test(fullname):
 bat_sfx = batch_suffix()
 exe_sfx = '.exe' if is_windows() else ""
 
-mini_python = [join(simics_base_path(), host_type(), "bin",
-                    "mini-python" + exe_sfx)]
+mini_python = [join(testparams.project_path(), "bin",
+                    f"mini-python{batch_suffix()}")]
```
lwaern-intel, line 69: I missed the `{batch_suffix()}` part. I'll run a quick Test W just to be sure it works OK there.

lwaern-intel: ... or so I say yet TC decided to rebuild Build W for no reason. bah.

lwaern-intel: [Here's the run](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=12947896&buildTypeId=SimicsCiSimics_MonoTests_TestWFast). All failures are unrelated, and the t126 suite is no slower, so I'd deem this good.

lwaern-intel (approved):

jhbaarnh:
@mandolaerik I need your approval too, since you clicked "request changes".

mandolaerik (approved):

PR #261, opened 2024-02-27
Leverage location when typechecking method overrides

lwaern-intel:
This fixes an issue when leveraging `typeof`: that a method signature using it would lead to incorrect compile-time errors if the method is overridden.

Ran into this while doing simple regs work. This PR alone should be an uncontroversial fix, however, I am looking warily at other uses of `eval_type` where `None` is passed as the location. Most of it is trait-related, which should really only be able to leverage top-level stuff -- however, today, they can't even do that, as they have no access to even `dev`. That question however should be left to a different PR.

### DIFF
--- a/RELEASENOTES.docu
+++ b/RELEASENOTES.docu
@@ -1757,4 +1757,8 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
       version 7 by passing either <tt>--warn=WLOGMIXUP</tt> (as before) or
       <tt>--no-compat=suppress_WLOGMIXUP</tt> to DMLC.</add-note>
   </build-id>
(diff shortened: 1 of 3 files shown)


test/1.4/expressions/T_typeof.dml
mandolaerik, line 1: `expressions/typeof` feels like wrong naming, `typeof` is not an expression. I suggest either move to `types/` or `methods/`, or something like `expressions/method_signature` (because I suppose the problem is with expressions that appear in method signatures; I suppose array sizes have the same issue, for instance)

Other than that, looks flawless to me.

syssimics:
Verification [#13057064](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13057064): pass

mandolaerik (approved):

PR #262, opened 2024-03-05
Create SECURITY.md

MagnusIntel:


mandolaerik (changes requested):
oh! I bet a copyright blurb is missing

lwaern-intel:
Yup. Our MD copyright blurb is:
```
<!--
  © 2024 Intel Corporation
  SPDX-License-Identifier: MPL-2.0
-->
```

syssimics:
Verification [#13093599](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13093599): fail

MagnusIntel:
@mandolaerik @lwaern-intel Why do you think we need a copyright blurb in this file? Please remove the "Changes requested" so we don't have spend more cycles in this.

mandolaerik:
> Why do you think we need a copyright blurb in this file? 
Quick answer: Because the test fails otherwise.

Longer: The blurb is needed because it's easy to write a test that requires it for all files, and in almost all files it's easier to just add a blurb than to add a passlist entry with an explanatory comment.

mandolaerik (approved):

PR #263, opened 2024-03-08
Fix EREGOL not being reported for registers of banks in groups/subdevices

lwaern-intel:


syssimics:
Verification [#13108782](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13108782): pass

test/1.4/errors/T_EREGOL.dml
```diff
@@ -69,3 +69,50 @@ bank b8[i < 2] {
...
+            register b size 4 @ 11;
+            // no error
+            register c size 4 @ 18;
+            register d size 4 @ 22;
+        }
+
+        bank b2 {
+            /// ERROR EIDXVAR
+            register r size 4 @ i;
+        }
```
mandolaerik, line 116: many lines for little coverage. The "no error" cases are known from before, and if `group { subdev { group { bank` works then it should be safe to assume the others work too.

Also, I think you can merge b1 and b2 now, after your try/except thingy.

lwaern-intel: Yeah alright. I'll use my discretion though, probably won't reduce it as much as you figure it deserves.

syssimics:
Verification [#13116684](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13116684): pass

mandolaerik (approved):

PR #265, opened 2024-03-11
Don't pass any data for zero getter or setter

pshishpo-intel:
The data has no meaning for absent getters and setters.

### DIFF
--- a/include/simics/dmllib.h
+++ b/include/simics/dmllib.h
@@ -1976,8 +1976,8 @@ _register_port_legacy_proxy_attr(conf_class_t *devcls, conf_class_t *portcls,
                 is_bank ? "bank" : "port", portname, attrname);
         SIM_register_typed_attribute(
                 devcls, name,
-                getter ? _get_legacy_proxy_attr : NULL, data,
-                setter ? _set_legacy_proxy_attr : NULL, data,
+                getter ? _get_legacy_proxy_attr : NULL, getter ? data : NULL,
+                setter ? _set_legacy_proxy_attr : NULL, setter ? data : NULL,
                 (attr_attr_t)(Sim_Attr_Pseudo | Sim_Attr_Internal),
                 type, NULL, sb_str(&proxy_desc));
         sb_free(&proxy_desc);


pshishpo-intel:
@mandolaerik Could you please review this pull request?

include/simics/dmllib.h
```diff
@@ -1976,8 +1976,8 @@ _register_port_legacy_proxy_attr(conf_class_t *devcls, conf_class_t *portcls,
                 is_bank ? "bank" : "port", portname, attrname);
         SIM_register_typed_attribute(
                 devcls, name,
-                getter ? _get_legacy_proxy_attr : NULL, data,
-                setter ? _set_legacy_proxy_attr : NULL, data,
+                getter ? _get_legacy_proxy_attr : NULL, getter ? data : NULL,
+                setter ? _set_legacy_proxy_attr : NULL, setter ? data : NULL,
```
lwaern-intel, line 1980: Does this make any observable difference? Or does the current state of affairs violate `SIM_register_type_attribute`'s API? If the answers to both are "no", then I must question why we should bother.

lwaern-intel: To clarify: this the answer to both questions are "no", then this is an unnecessary complication. It'd be meaningless code, which is of negative value.

syssimics:
Verification [#13116912](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13116912): pass

include/simics/dmllib.h
pshishpo-intel: @lwaern-intel AFAIK there is no observable difference. I came across this place when I decided to check if there are any attributes registered that have `NULL` getter or setter but non-`NULL` data pointer. My intention was to find possible errors where a `NULL` getter or setter was specified by mistake. I don't remember if any errors were found.
I think that passing NULL for data would be more consistent.
If you still think that this is an unnecessary complication, please just reject the pull request. No worries here.

lwaern-intel:
Rejected as unnecessary.

PR #266, opened 2024-03-11
Explain default initializer for local variables

nemkoveu:
Explicitly document that `local` variables are initialized with zeroes by default, similar to `saved` and `session`.


nemkoveu:
@mandolaerik, please review.

syssimics:
Verification [#13121001](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13121001): pass

PR #267, opened 2024-03-26
Add uint64_t and int64_t types

mandolaerik:
- Repair test
- Add uint64_t and int64_t types

### DIFF
--- a/doc/1.4/language.md
+++ b/doc/1.4/language.md
@@ -1314,11 +1314,15 @@ and `layout`.
 
 The familiar integer types `char` and `int` are
 available as aliases for `int8` and `int32`,
-respectively. The C keywords `short`, `signed`, `unsigned`
-and `long` are reserved words in DML and not allowed in type
-declarations.  The types `size_t` and `uintptr_t` are
-defined as in C.
-</dd><dt>
+respectively. The C keywords `short`, `signed` and `unsigned`
+are reserved words in DML and not allowed in type
+declarations.
+
+The types `size_t` and `uintptr_t`, `long`, `uint64_t`, `int64_t`, are defined
+as in C. The types `long`, `uint64_t` and `int64_t` are provided mainly for
+compatibility with third party libraries; they are needed because they are
+incompatible with the corresponding Simics types (`uint64`, etc) on some
+platforms.  </dd><dt>
 
 Endian integers
 </dt><dd>
--- a/py/dml/types.py
+++ b/py/dml/types.py
@@ -482,6 +482,8 @@ class IntegerType(DMLType):
             return NotImplemented
         if isinstance(self, TSize) != isinstance(other, TSize):
             return NotImplemented
+        if isinstance(self, TInt64_t) != isinstance(other, TInt64_t):
+            return NotImplemented
         if (dml.globals.dml_version == (1, 2)
             and compat.dml12_int in dml.globals.enabled_compat):
             # Ignore signedness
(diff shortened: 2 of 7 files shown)


syssimics:
Verification [#13186013](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13186013): pass

py/dml/types.py
```diff
@@ -610,6 +612,29 @@ def clone(self):
...
+    `uint64_t *` arg. We find `uint64` a generally more useful type
+    primarily because it works better with format strings; "%lld" works
+    for `uint64` on both linux64 and win64.
+    '''
+    __slots__ = ()
+    def __init__(self, signed, const=False):
+        IntegerType.__init__(self, 64, signed, const=const)
+
+    def describe(self):
+        return self.declaration('').rstrip()
```
lwaern-intel, line 627: ew. I'd vote for repetition instead (`return 'int64_t' if self.signed else 'uint64_t`), or, if you want to be particularly snazzy, leverage my "favorite" trick: `return 'u'*(not self.signed) + 'int64_t'`.

py/dml/types.py
```diff
@@ -1271,6 +1296,13 @@ def compatible_types(type1, type2):
         ('size_t', TSize(False))]:
     typedefs[name] = typ
 
+if dml.globals.dml_version != (1, 2):
```
lwaern-intel, line 1299: You shouldn't do this at top-level! `dml.globals.dml_version` can't be well-defined here, can it? Even if it is, that feels like that would _very_ easily break from any change we do re. when dml version is determined.

test/1.2/operators/T_sizeof.dml
```diff
@@ -39,8 +39,8 @@ method test -> (bool ok) {
           (sizeoftype size_t == sizeoftype void *));
 
     // check that all sizeof invocations return a size_t
-    log "info": "%zd %zd %zd %zd", sizeof x, sizeof $x,
-        sizeoftype uint32, sizeoftype (t);
+    SIM_log_info(1, $dev.obj, 0, "%zd %zd %zd %zd", sizeof x, sizeof $x,
```
lwaern-intel, line 42: What's this change about?

lwaern-intel (changes requested):

test/1.2/operators/T_sizeof.dml
mandolaerik: Oh, I just found it on the way. IIRC, `log` adjust format strings to make qualifiers match args, so passing it as a `%zd` arg to `log` doesn't necessarily check size_t-ness, but passing it to SIM_log_info does. (format string conversions happen correctly in 1.4; I don't remember what limitations there are in 1.2 but this change is arguably an improvement)

py/dml/types.py
mandolaerik: Oh, crap. Will fix it somehow.

test/1.2/operators/T_sizeof.dml
lwaern-intel: Alright.

doc/1.4/language.md
```diff
@@ -1316,7 +1316,7 @@ The familiar integer types `char` and `int` are
 available as aliases for `int8` and `int32`,
 respectively. The C keywords `short`, `signed`, `unsigned`
 and `long` are reserved words in DML and not allowed in type
-declarations.  The types `size_t` and `uintptr_t` are
+declarations.  The types `uint64_t`, `int64_t`, `size_t` and `uintptr_t` are
 defined as in C.
```
lwaern-intel, line 1320: This deserves some kind of clarifying disclaimer. How about:
> The types `uint64_t`, `int64_t`, `size_t` and `uintptr_t` are
defined as in C. (`uint64_t` and `int64_t` are guaranteed compatible with the types offered by `stdlib.h`, in contrast to `uint64` and `int64` which are defined by Simics.)

py/dml/types.py
```diff
@@ -610,6 +612,33 @@ def clone(self):
...
+    primarily because it works better with format strings; "%lld" works
+    for `uint64` on both linux64 and win64.
+    '''
+    __slots__ = ()
+    def __init__(self, signed, const=False):
+        IntegerType.__init__(self, 64, signed, const=const)
+
+    def c_name(self):
+        name = 'int64_t' if self.signed else 'uint64_t'
+        return f'const {name}' if self.const else name
```
lwaern-intel, line 628: Consider using `self.const_str()` like everything else (unless you're doing away with that function)

lwaern-intel (changes requested):

py/dml/types.py
mandolaerik: done

doc/1.4/language.md
mandolaerik: changed

py/dml/types.py
mandolaerik: factored out a function

syssimics:
Verification [#13268602](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13268602): pass

syssimics:
Verification [#13268624](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13268624): pass

syssimics:
Verification [#13268743](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13268743): pass

PR #268, opened 2024-04-04
Fix immediate after callbacks not notifying about state change

lwaern-intel:
I kinda sorta forgot about this when I wrote immediate after.

After-on-hooks doesn't have the same problem as messages can only be sent through a hook from within DML itself.

### DIFF
--- a/RELEASENOTES-1.4.docu
+++ b/RELEASENOTES-1.4.docu
@@ -526,4 +526,7 @@
   <build-id _6="6287" _7="7012"><add-note> Fixed a bug where overlapping
       registers would not be rejected if they were part of a bank within a group
       or subdevice.</add-note></build-id>
+  <build-id _6="next" _7="next"><add-note> Fixed a bug where executed callbacks
+      posted via immediate after statements would not trigger the device state
+      change notifier.</add-note></build-id>
 </rn>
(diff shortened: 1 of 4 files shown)


syssimics:
Verification [#13215078](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13215078): pass

lwaern-intel:
Right, @mandolaerik is on vacation. Requesting @JonatanWaern as reviewer.

RELEASENOTES-1.4.docu
```diff
@@ -526,4 +526,7 @@
   <build-id _6="6287" _7="7012"><add-note> Fixed a bug where overlapping
       registers would not be rejected if they were part of a bank within a group
       or subdevice.</add-note></build-id>
+  <build-id _6="next" _7="next"><add-note> Fixed a bug where executed callbacks
+      posted via immediate after statements would not trigger the device state
+      change notifier.</add-note></build-id>
```
mandolaerik, line 531: `state-change`?

mandolaerik (approved):

RELEASENOTES-1.4.docu
lwaern-intel: Not according to a releasenote in RELEASENOTES-1.2:
>  <build-id value="6122"><add-note> Added the <tt>--state-change-dml12</tt> flag, allowing the use of an incomplete version of state change notifiers in DML 1.2 devices <bug number="SIMICS-17952"/>.</add-note></build-id>

PR #271, opened 2024-04-17
Document order semantics of `each`-`in` and `init`

lwaern-intel:
I thought we already did explicitly say the order of `each`-`in`s was not defined, but nope.

### DIFF
--- a/doc/1.4/language.md
+++ b/doc/1.4/language.md
@@ -3897,6 +3897,13 @@ template itself contain a method that descends into subobjects; the
 implementation of `hard_reset` in `utility.dml`
 demonstrates how this can be done.
 
+The order in which objects are given by a specific `each`-`in` expression is not
+defined, except for that it is deterministic. That is, for a particular choice
+of template `X` and object `Y` in an `each X in (Y)` expression, for a
+particular iteration of the device model, and for the particular DMLC build
+used, the order in which objects are given by that expression is guaranteed to
+be consistent.
+
 ### List Expressions
 
 <pre>
--- a/lib/1.4/dml-builtins.dml
+++ b/lib/1.4/dml-builtins.dml
@@ -365,7 +365,12 @@ device is created, *before* any attributes have been
 initialized. Typically used to initialize a default value, or to set up data
 structures.
 
-The method `init` is automatically called on all objects that implement the `init` template.
+The method `init` is automatically called on all objects that implement the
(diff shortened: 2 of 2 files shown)


doc/1.4/language.md
```diff
@@ -3897,6 +3897,12 @@ template itself contain a method that descends into subobjects; the
 implementation of `hard_reset` in `utility.dml`
 demonstrates how this can be done.
 
+The order in which objects are given by a specific `each`-`in` expression is not
+defined, save that it is deterministic. That is, for a particular choice of
```
TSonono, line 3901: I think "save" here is confusing, but that might just be me

TSonono (commented):
Good to include this in the docs!

doc/1.4/language.md
lwaern-intel: Perhaps this is slightly grammatically incorrect and I should be saying "save for that" instead of simply "save that" (I know I've definitely seen both used, but that doesn't mean both are right). Regardless, yeah, "save" can be used to mean "with the only exception of" or "aside from only". I suspect the origin of that usage comes from the idea of "it would be this thing, if it weren't saved by this one exception".  It's one of the less common uses of the word.

mandolaerik: or "except it is deterministic"

lib/1.4/dml-builtins.dml
```diff
@@ -365,7 +365,12 @@ device is created, *before* any attributes have been
 initialized. Typically used to initialize a default value, or to set up data
 structures.
 
-The method `init` is automatically called on all objects that implement the `init` template.
+The method `init` is automatically called on all objects that implement the
+`init` template. The order in which `init` of objects are called is not
+defined, except that `init` of a particular object is guaranteed to be called
+before `init` of any of its parent objects. In particular, `init` of the device
```
mandolaerik, line 371: consider s/device/`device`/ (but both are ok)

lwaern-intel: I protest against the latter, as this is not referring to the `device` object kind, template, or template type in this context, but rather, referring to the single canonical DML object representing the device.

doc/1.4/language.md
lwaern-intel: ah whatever. Went for "except for that it is deterministic."

mandolaerik (approved):

PR #272, opened 2024-04-18
Test and repair `const` variants of all types

mandolaerik:
- **Factor out methods**
- **Test and repair `const` variants of all types**

### DIFF
--- a/RELEASENOTES.docu
+++ b/RELEASENOTES.docu
@@ -1766,4 +1766,9 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
   <build-id _6="next" _7="next"><add-note> DMLC will now report errors
       for some uses of function types that are illegal in C. These
       would previously yield broken C code.</add-note></build-id>
+  <build-id _6="next" _7="next"><add-note>Created a compatibility feature
+      <tt>function_in_extern_struct</tt>. The syntax <tt>void
+      m(conf_object_t)</tt> for function pointers is only permitted
+      when this feature is enabled. The feature will be disabled in
+      Simics API versions > 7.</add-note></build-id>
 </rn>
--- a/RELEASENOTES.docu
+++ b/RELEASENOTES.docu
@@ -1763,4 +1763,7 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
   </build-id>
   <build-id _6="next" _7="next"><add-note> Added support for the
       <tt>log warning</tt> statement.</add-note></build-id>
+  <build-id _6="next" _7="next"><add-note> DMLC will now report errors
+      for some uses of function types that are illegal in C. These
+      would previously yield broken C code.</add-note></build-id>
 </rn>
--- a/py/dml/c_backend.py
+++ b/py/dml/c_backend.py
@@ -778,17 +778,22 @@ def generate_implement_method(device, ifacestruct, meth, indices):
         require_fully_typed(None, meth)
 
         # Calculate the expected method signature
-        ifacemethtype = ifacestruct.get_member_qualified(meth.name)
-        if not ifacemethtype:
+        member_type = ifacestruct.get_member_qualified(meth.name)
+        if not member_type:
             raise EMEMBER(meth.site, meth.parent.name, meth.name)
-        ifacemethtype = safe_realtype(ifacemethtype)
-        if isinstance(ifacemethtype, TPtr):
-            ifacemethtype = ifacemethtype.base
-        if not isinstance(ifacemethtype, TFunction):
-            raise EBTYPE(meth.site, ifacemethtype,
-                         TPtr(TFunction([], TVoid())))
-        iface_input_types = ifacemethtype.input_types[1:]
-        iface_num_outputs = 0 if ifacemethtype.output_type.void else 1
+        member_type = safe_realtype(member_type)
+        if not isinstance(member_type, TPtr):
+            raise EIMPLMEMBER(
+                meth.site,
+                f'{meth.parent.name}_interface_t.{meth.name}',
+                ifacestruct.declaration_site)
+        func_type = member_type.base
+        if not isinstance(func_type, TFunction):
+            raise EIMPLMEMBER(meth.site,
+                              f'{meth.parent.name}_interface_t.{meth.name}',
+                              ifacestruct.declaration_site)
+        iface_input_types = func_type.input_types[1:]
+        iface_num_outputs = 0 if func_type.output_type.void else 1
 
         # Check the signature
         if len(meth.inp) != len(iface_input_types):
@@ -797,7 +802,7 @@ def generate_implement_method(device, ifacestruct, meth, indices):
         if len(meth.outp) != iface_num_outputs:
             raise EMETH(meth.site, None,
                         'different number of output parameters')
-        if ifacemethtype.varargs:
+        if func_type.varargs:
             # currently impossible to implement a varargs interface
             # method in DML
             raise EMETH(meth.site, None, 'interface method is variadic')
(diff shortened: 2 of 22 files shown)


lwaern-intel (changes requested):

py/dml/types.py
```diff
@@ -1158,14 +1181,15 @@ def clone(self):
...
-        params = [ t.declaration("arg%d" % n)
-                   for (n, t) in enumerate(self.input_types) ]
+        arglist = ', '.join(t.declaration("arg%d" % n)
+                            for (n, t) in enumerate(self.input_types))
         if self.varargs:
-            varargs = ", ..."
+            arglist += ", ..."
+        if self.const:
+            funtype = self.output_type.declaration(f'({arglist})')
+            return f'const typeof({funtype}) {var}'
```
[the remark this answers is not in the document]
mandolaerik: I don't (possibly because someone thought about it and realized there is no risk).
However, you do get a hard error for
```
        typedef const typeof(void (int)) *t1;
        typedef       typeof(void (int)) *t2;
        void (*f)(t1) = NULL;
        void (*g)(t2) = f;
```
... because pointer incompatibility means invalid C.

py/dml/types.py
```diff
@@ -893,6 +912,10 @@ def cmp(self, other):
     def key(self):
         return f'{self.const_str}sequence({self.traitname})'
 
+    def c_type(self):
+        const = 'const ' if self.const else ''
+        return f'{const}_each_in_t'
```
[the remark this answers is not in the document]
mandolaerik: Fixed.

py/dml/ctree_test.py
```diff
@@ -1373,5 +1373,66 @@ def assign_trunc(self):
...
+            types.TTraitList('struct_t'),
+            types.TExternStruct({}, 'struct_t', 'struct_t'),
+            types.TStruct({'x': types.TBool()}, 'struct_label'),
+            types.TLayout(
+                'big-endian', {
+                    'x': (site, types.TEndianInt(24, True, 'big-endian'))},
+                'struct_label'),
+            types.TFunction([], types.TVoid()),
+            types.THook([]),
+        ]
```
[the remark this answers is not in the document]
mandolaerik: Good catch, fixed.

syssimics:
Verification [#13269417](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13269417): fail

syssimics:
Verification [#13270893](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13270893): fail

syssimics:
Verification [#13271177](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13271177): pass

syssimics:
Verification [#13273115](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13273115): fail

syssimics:
Verification [#13274113](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13274113): pass

py/dml/types.py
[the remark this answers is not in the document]
mandolaerik: Reworked now.

RELEASENOTES.docu
```diff
@@ -1761,4 +1761,12 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
       <tt>typeof</tt> within a method signature would lead to incorrect
       compile-time errors when the method is overridden.</add-note>
   </build-id>
+  <build-id _6="next" _7="next"><add-note> DMLC will now report errors
+      for some uses of function types that are illegal in C. These
+      would previously yield broken C code.</add-note></build-id>
```
lwaern-intel, line 1766: Why not mention the compatibility flag?

py/dml/c_backend.py
```diff
@@ -778,17 +778,22 @@ def generate_implement_method(device, ifacestruct, meth, indices):
...
         # Calculate the expected method signature
-        ifacemethtype = ifacestruct.get_member_qualified(meth.name)
-        if not ifacemethtype:
+        member_type = ifacestruct.get_member_qualified(meth.name)
+        if not member_type:
             raise EMEMBER(meth.site, meth.parent.name, meth.name)
-        ifacemethtype = safe_realtype(ifacemethtype)
-        if isinstance(ifacemethtype, TPtr):
-            ifacemethtype = ifacemethtype.base
-        if not isinstance(ifacemethtype, TFunction):
```
lwaern-intel, line 787: wait so the code was actually _intentionally written to support it?_ why the hell...?

py/dml/c_backend.py
```diff
@@ -778,17 +778,22 @@ def generate_implement_method(device, ifacestruct, meth, indices):
...
-            raise EBTYPE(meth.site, ifacemethtype,
-                         TPtr(TFunction([], TVoid())))
-        iface_input_types = ifacemethtype.input_types[1:]
-        iface_num_outputs = 0 if ifacemethtype.output_type.void else 1
+        member_type = safe_realtype(member_type)
+        if not isinstance(member_type, TPtr):
+            raise EIMPLMEMBER(
+                meth.site,
+                f'{meth.parent.name}_interface_t.{meth.name}',
+                ifacestruct.declaration_site)
```
lwaern-intel, line 789: Why did you decide to have the compatibility feature preserve the use of the thing in the context of interface objects (you haven't touched `mkInterfaceMethodRef`), but not here in the context of implements?

py/dml/types.py
```diff
@@ -1146,13 +1169,21 @@ class TFunction(DMLType):
     def __init__(self, input_types, output_type,
                  varargs = False, const = False):
         assert isinstance(output_type, DMLType)
+        assert not self.const
```
lwaern-intel, line 1172: Why do this when you have
```
    @property
    def const(self): return False
```
straight-up clear as day? You might as well `assert True`. Especially as this does not protect against anything even _if_ that definition of `TFunction.const` would be removed as you're checking `self.const` before `DMLType.__init__` is called to initialize it. The check would be more meaningful if it was replaced with `assert not const`, or if it was moved to after `__init__` (but it would still be redundant regardless. Even the former check is covered by the `const.setter` override.)

lwaern-intel (changes requested):

py/dml/types.py
mandolaerik: Right, leftover from before I realized the `@property` solution. Thanks for spotting.

py/dml/c_backend.py
mandolaerik: Well spotted, I didn't think we had any code for this on the interface side, but indeed we do; `mkInterfaceMethodRef` indeed does
```
    if isinstance(ftype, TPtr):
        ftype = ftype.base
```
I will update that one too.

py/dml/c_backend.py
mandolaerik: A guess:
1. some user sees "wow, I can express interface struct types as `void f(conf_object_t*);` so they look like normal C function declarations", and this worked out-of-the-box for `interface` objects (GCC doen't care, and not much typechecking was done by DMLC)
2. some user notices that such interfaces don't work with `implement` and reports a bug on DMLC
3. the above is a simple fix for that bug

RELEASENOTES.docu
mandolaerik: The compatibility feature is mentioned in the releasenote just below this comment.

lwaern-intel: Uhhh
I have no excuse.

py/dml/c_backend.py
lwaern-intel: What's the point of the compatibility feature if it doesn't support the most prominent existing use case of functions-in-structus?

syssimics:
Verification [#13281765](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13281765): fail

syssimics:
Verification [#13281974](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13281974): pass

py/dml/c_backend.py
lwaern-intel: Ahh, I see.

test/1.4/errors/T_ESERIALIZE.dml
```diff
@@ -39,6 +39,12 @@ typedef int vect int_vect;
 /// ERROR ESERIALIZE
 saved int_vect x;
 
+// ESERIALIZE causes the compiler to omit the faulty declarations in further
+// analyses; if this didn't happen, function types would cause some sneaky
+// crashes.
+/// ERROR ESERIALIZE
+saved void function(void);
```
lwaern-intel, line 46: Hmm. What about `session void function(void);`?

lwaern-intel (approved):
Did you want to make the change to linemarks re. `local { int i; } s;` part of this PR or separate?

PR #273, opened 2024-04-19
Break out default implementations of `read_register`/`write_register`

lwaern-intel:
... So they are always reachable. Internal as they are stand-ins for the lack of template-qualified method implementation calls.

### DIFF
--- a/lib/1.4/dml-builtins.dml
+++ b/lib/1.4/dml-builtins.dml
@@ -2929,7 +2929,8 @@ template register is (_conf_attribute, get, set, shown_desc,
         return this.val & enabled_bits;
     }
 
-    shared method read_register(uint64 enabled_bytes, void *aux)-> (uint64) default {
+    // TODO: stand-in for SIMICS-22264. Inline and remove when longer needed
+    shared method _base_read_register(uint64 enabled_bytes, void *aux)-> (uint64) {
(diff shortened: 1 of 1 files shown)


gwikland (approved):

syssimics:
Verification [#13272231](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13272231): pass

lib/1.4/dml-builtins.dml
```diff
@@ -2929,7 +2929,7 @@ template register is (_conf_attribute, get, set, shown_desc,
         return this.val & enabled_bits;
     }
 
-    shared method read_register(uint64 enabled_bytes, void *aux)-> (uint64) default {
+    shared method _base_read_register(uint64 enabled_bytes, void *aux)-> (uint64) {
```
mandolaerik, line 2932: Would not hurt with an explanatory comment, e.g. it's internal/subject to change, stand-in for a hopefully-upcoming feature, see SIMICS-NNNN, etc

mandolaerik: SIMICS-22264, that is

lwaern-intel: Done.

mandolaerik (approved):

PR #275, opened 2024-04-24
Validate that we don't forget linemarks

mandolaerik:


mandolaerik:
Draft for discussion. The handling of `header`/`footer` is wasteful, probably deserves some fast path. I'm pondering if `linemark` would better be folded into `out`; i.e., when generating code that originates from DML code, then you call `out('..', loc=('file', line))`, otherwise you call a plain `out('...')` which resets the line directive.

syssimics:
Verification [#13289608](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13289608): fail

PR #277, opened 2024-04-29
Use compat features to deprecate bad version statements

mandolaerik:
- **Use compat features to deprecate bad version statements**
- **Improve wording**
- **Use package_path from Simics base**

### DIFF
--- a/py/dml/compat.py
+++ b/py/dml/compat.py
@@ -102,6 +102,18 @@ class function_in_extern_struct(CompatFeature):
     short = 'Disallow non-pointer function members in extern structs'
     last_api_version = api_7
 
+
+@feature
+class optional_version_statement(CompatFeature):
+    '''When this compatibility feature is enabled, the version
+    specification statement (`dml 1.4;`) statement at the start of
+    each file is optional (but the compiler warns if it is
+    omitted). Also, `dml 1.3;` is permitted as a deprecated alias for
+    `dml 1.4;`, with a warning.'''
+    short = "Make the DML version statement optional"
+    last_api_version = api_7
+
+
 @feature
 class io_memory(CompatFeature):
     '''The `transaction` interface was introduced in 6, and will
--- a/py/dml/compat.py
+++ b/py/dml/compat.py
@@ -99,7 +99,7 @@ class function_in_extern_struct(CompatFeature):
     } my_interface_t;
     ```
     '''
-    short = 'Disallow non-pointer function members in extern structs'
+    short = 'Allow non-pointer function members in extern structs'
     last_api_version = api_7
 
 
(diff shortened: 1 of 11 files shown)


syssimics:
Verification [#13306627](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13306627): pass

mandolaerik:
test failure is unrelated

lwaern-intel (changes requested):
This currently has parts from #279 and it shouldn't! Filter them out.

py/dml/compat.py
```diff
@@ -99,9 +99,21 @@ class function_in_extern_struct(CompatFeature):
...
     ```
     '''
-    short = 'Disallow non-pointer function members in extern structs'
+    short = 'Allow non-pointer function members in extern structs'
     last_api_version = api_7
 
+
+@feature
+class optional_version_statement(CompatFeature):
+    '''When this compatibility feature is enabled, the `dml 1.4;`
```
lwaern-intel, line 108: > The version specification statement (e.g. `dml 1.4;`) at the start of each file is optional

py/dml/toplevel.py
```diff
@@ -74,33 +74,41 @@ def determine_version(filestr, filename):
...
-        lineno = 1
-        column = 1
+        if compat.optional_version_statement in dml.globals.enabled_compat:
+            report(WNOVER(SimpleSite(f"{filename}:1")))
+            version = (1, 2)
+            lineno = 1
+            column = 1
+        else:
+            raise ESYNTAX(
+                SimpleSite(f"{filename}:1"), None,
```
lwaern-intel, line 92: Consider breaking out the identical usages of `SimpleSite` to a variable defined before the `if`.

lwaern-intel (changes requested):
Approved modulo a minor quibble.

lwaern-intel (approved):
no really it is approved. I want you to address the quibble, but if you disagree with it, that's fine.

syssimics:
Verification [#13346188](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13346188): pass

py/dml/toplevel.py
mandolaerik: No need to, duplication in dying code is harmless

py/dml/compat.py
mandolaerik: I would agree that this clarification would be needed if this were in a non-dying part of the spec, but this is a dying path so easy-to-read is more important than formal correctness. It's easy to understand that it refers to `dml 1.2;` in a 1.2 context.

syssimics:
Verification [#13346831](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13346831): pass

PR #278, opened 2024-05-02
Template-Qualified Method Implementation Calls -- SIMICS-22264

lwaern-intel:
Completely untested save that I checked that one (1) existing 1.4 test still passed and nothing else

Currently limited: you can only make a TQMIC if the template specified actually does provide an implementation of the method. In contrast, you could consider that, in principle, if the specified template doesn't provide an implementation then the TQMIC should resolve to the the nearest ancestor of the specified template that provides an implementation. I couldn't think of a good way to do that however with regular methods (shared methods are easy thanks to the `Trait` architecture.)

### DIFF
--- a/RELEASENOTES-1.4.docu
+++ b/RELEASENOTES-1.4.docu
@@ -537,4 +537,11 @@
       template now allows for its implementations of <tt>unmapped_read</tt>,
       <tt>unmapped_write</tt>, and <tt>unmapped_get</tt> to be
       overridden. </add-note></build-id>
+  <build-id _6="next" _7="next"><add-note> It's now allowed to instantiate
+      multiple unrelated templates that offer different <tt>shared</tt>
+      implementations of the same method, as long as that method is declared
+      in a common ancestor template, and the conflict is resolved through
+      an implementation that overrides all the unrelated implementations
+      <bug number="SIMICS-20431"/>. This mirrors the pre-existing semantics for
+      non-<tt>shared</tt> methods.</add-note></build-id>
 </rn>
--- a/RELEASENOTES-1.4.docu
+++ b/RELEASENOTES-1.4.docu
@@ -544,4 +544,18 @@
       an implementation that overrides all the unrelated implementations
       <bug number="SIMICS-20431"/>. This mirrors the pre-existing semantics for
       non-<tt>shared</tt> methods.</add-note></build-id>
+  <build-id _6="next" _7="next"><add-note> Added the `templates` member for
+      all objects and values of template types, which allows for making calls
+      to any particular implementation of a method as provided by a specified
+      template <bug number="SIMICS-22264"/>. Example syntax:
+      <pre>
+      // Call the `write_register()` implementation as provided by the
+      // `register` template on the `r` register of the bank `b`.
+      b.r.templates.register.write_register(...);
+      </pre>
+      Calls of this form are called <em>Template-Qualified Method Implementation
+      Calls</em>, and are primarily meant to make it easier to resolve conflicts
+      in multiple inheritance: it allows any method that overrides multiple
+      implementations from unrelated templates to individually reference and
+      call the implementations it overrides.</add-note></build-id>
 </rn>
--- a/doc/1.4/language.md
+++ b/doc/1.4/language.md
@@ -1079,6 +1079,10 @@ A template type has the following members:
   `post` and `next` are members of
   the `simple_time_event` template type.
 
+* The `templates` member, which permits [template-qualified method
+  implementation calls](#template-qualified-method-implementation-calls) to
+  the `shared` method implementations of the template type's ancestor templates.
+
 Template members are dereferenced using the `.` operator,
 much like struct members.
 
@@ -1680,6 +1684,15 @@ method from within the body of the overriding method using the identifier
 ```
 x = default(...);
 ```
+
+In addition to `default`, there exists the [`templates` member of
+objects](#template-qualified-method-implementation-calls) which allows for
+calling the particular implementation of a method as provided by a specified
+template. This is particularly useful when `default` _can't_ be used due to the
+method overriding implementations provided by multiple hierarchically unrelated
+templates, such that `default` can't be unambiguously resolved (see [Resolution
+of overrides](#resolution-of-overrides).) Unlike `default`, `templates` can also
+be used even outside the body of the overriding method.
 
 DML supports _compound initializer syntax_ for the arguments of called methods,
 meaning arguments of struct-like types can be constructed using
@@ -3235,6 +3248,176 @@ local (int i, uint8 j) = m(e1);
 // caller.
 return m(e1)
 ```
+
+### Template-Qualified Method Implementation Calls
+Every object, as well as [every template type](#templates-as-types), has a
+`templates` member to allow for calling _particular_ implementations of that
+object's methods, as opposed to only the final overriding implementations
+that are reachable directly. Specifically, `templates` allows for invoking any
+particular implementation as provided by a specified template instantiated by
+the object. Such invocations are called _template-qualified method
+implementation calls_, and are made as follows:
+```
+template t {
+    method m() default {
+        log info: "implementation from 't'"
+    }
+}
+
+template u is t {
+    method m() default {
+        log info: "implementation from 'u'"
+    }
+}
+
+group g is u {
+    method m() {
+        log info: "final implementation"
+    }
+}
+
+method call_ms() {
+    // Logs "final implementation"
+    g.m();
+    // Logs "implementation from 'u'"
+    g.templates.u.m();
+    // Logs "implementation from 't'"
+    g.templates.t.m();
+}
+```
+
+Template-qualified method implementation calls are primarily meant as a way
+for an overriding method to reference overridden implementations, *even when*
+the implementations are provided by hierarchically unrelated templates such that
+`default` can't be used (see [Resolution of
+overrides](#resolution-of-overrides).) In particular, this typically allows for
+ergonomically resolving conflicts introduced when multiple orthogonal templates
+are instantiated, as long as all conflicting implementations are overridable,
+and one of the following is true:
+* The implementations can be combined together by calling each one of them, as
+  long as that can be done without risking e.g. side-effects being duplicated.
+* The implementations can be combined by choosing one particular template's
+  implementation to invoke (typically the one most complex), and then adding
+  code around that implementation call in order to replicate the behaviour of
+  the implementations of the other templates. Ideally, the other templates would
+  provide methods that may be leveraged so that their behaviour may be
+  replicated without the need for excessive boilerplate.
+
+The following is an example of the first case:
+```
+template alter_write is write {
+    method write(uint64 written) {
+        default(alter_write(written));
+    }
+
+    method alter_write(uint64 curr, uint64 written) -> (uint64);
+}
+
+template gated_write is alter_write {
+    method write_allowed() -> (bool) default {
+        return true;
+    }
+
+    method alter_write(uint64 curr, uint64 written) -> (uint64) default {
+        return write_allowed() ? written : curr;
+    }
+}
+
+template write_1_clears is alter_write {
+    method alter_write(uint64 curr, uint64 written) -> (uint64) default {
+        return curr & ~written;
+    }
+}
+
+template gated_write_1_clears is (gated_write, write_1_clears) {
+    method alter_write(uint64 curr, uint64 written) default {
+        local uint64 new = this.templates.write_1_clears.alter_write(
+            curr, written);
+        return this.templates.gated_write.alter_write(curr, new);
+    }
+}
+
+// Resolve the conflict introduced whenever the two orthogonal templates are
+// instantiated by also instantiating gated_write_1_clears when that happens
+in each (gated_write, write_1_clears) { is gated_write_1_clears; }
+```
+
+The following is an example of the second case:
+```
+template very_complex_register is register {
+    method write_register(uint64 written, uint64 enabled_bytes,
+                          void *aux) default {
+        ... // An extremely complicated implementation
+    }
+}
+
+template gated_register is register {
+    method write_allowed() -> (bool) default {
+        return true;
+    }
+
+    method on_write_attempted_when_not_allowed() default {
+        log spec_viol: "%s was written to when not allowed", qname;
+    }
+
+    method write_register(uint64 written, uint64 enabled_bytes,
+                          void *aux) default {
+        if (write_allowed()) {
+            default(written, enabled_bytes, aux);
+        } else {
+            on_write_attempted_when_not_allowed();
+        }
+    }
+}
+
+template very_complex_gated_register is (very_complex_register,
+                                         gated_register) {
+    // No sensible way to combine the two implementations by calling both.
+    // Even if there were, calling both implementations would cause each field
+    // of the register to be written to multiple times, potentially duplicating
+    // side-effects, which is undesirable.
+    // Instead, very_complex_register is chosen as the base implementation
+    // called, and the behaviour of gated_register is replicated around that
+    // call.
+    method write_register(uint64 written, uint64 enabled_bytes,
+                          void *aux) default {
+        if (write_allowed()) {
+            this.templates.very_complex_register.write_register(
+                written, enabled_bytes, aux);
+        } else {
+            on_write_attempted_when_not_allowed();
+        }
+    }
+}
+
+in each (gated_register, very_complex_register) {
+    is very_complex_gated_register;
+}
+```
+
+A template-qualified method implementation call is resolved by using
+the method implementation provided to the object by the named template.
+If no such implementation is provided (whether it be because the template does
+not specify one, or specifies one which is not provided to the object due to its
+definition being eliminated by an [`#if`](#conditional-objects)), then the
+ancestor templates of the named template are recursively searched for the
+highest-rank (most specific) implementation provided by them. If the ancestor
+templates provide multiple hierarchically unrelated implementations, then the
+choice is ambiguous and the call will be rejected by the compiler. In this case,
+the modeller must refine the template-qualified method implementation call to
+name the ancestor template whose implementation they would like to use.
+
+A template-qualified method implementation call done via [a value of template
+type](#templates-as-types) functions differently compared to compile-time
+object references. In particular, `this.templates` within the bodies of `shared`
+methods functions differently. The specified template must be an ancestor
+template of the value's template type, the <tt>object</tt> template, or the
+template type itself; furthermore, the specified template **must provide or
+inherit a `shared` implementation of the named method**. It is not sufficient
+that the method is simply _declared_ `shared` such that it is part of the
+template type: the implementation itself must also be `shared`. For more
+information, see the documentation of the [`ENSHAREDTQMIC` error
+message.](messages.html#ENSHAREDTQMIC)
 
 ### After Statements
 
--- a/py/dml/codegen.py
+++ b/py/dml/codegen.py
@@ -1245,37 +1245,10 @@ def expr_objectref(tree, location, scope):
                     dmlparse.start_site(tree.site), '', prefix))
     return e
 
-def try_resolve_len(site, lh):
-    if isinstance(lh, NonValue):
-        if isinstance(lh, AbstractList):
-            return mkIntegerConstant(site,
-                                     len(tuple(lh.iter_flat())), False)
-        elif isinstance(lh, NonValueArrayRef):
-            return mkIntegerConstant(site,
-                                     lh.local_dimsizes[len(lh.local_indices)],
-                                     False)
-    return None
-
 @expression_dispatcher
 def expr_member(tree, location, scope):
     [lh, op, rh] = tree.args
     lh = codegen_expression_maybe_nonvalue(lh, location, scope)
-    if op == '.':
-        if not tree.site.dml_version() == (1, 2) and rh == 'len':
-            member = try_resolve_len(tree.site, lh)
-            if member:
-                return member
-        elif isinstance(lh, TemplatesRef):
-            return mkTemplatesSubRef(tree.site, lh, rh)
-        elif isinstance(lh, TemplatesSubRef):
-            return mkTemplateQualifiedMethodRef(tree.site, lh, rh)
-        elif isinstance(lh, TraitTemplatesRef):
-            return mkTraitTemplatesSubRef(tree.site, lh, rh)
-        elif isinstance(lh, TraitTemplatesSubRef):
-            return mkTraitTemplateQualifiedMethodRef(tree.site, lh, rh)
-
-    if isinstance(lh, NonValue) and not isinstance(lh, NodeRef):
-        raise lh.exc()
 
     return mkSubRef(tree.site, lh, rh, op)
 
--- a/py/dml/codegen.py
+++ b/py/dml/codegen.py
@@ -1260,10 +1260,19 @@ def try_resolve_len(site, lh):
 def expr_member(tree, location, scope):
     [lh, op, rh] = tree.args
     lh = codegen_expression_maybe_nonvalue(lh, location, scope)
-    if not tree.site.dml_version() == (1, 2) and op == '.' and rh == 'len':
-        member = try_resolve_len(tree.site, lh)
-        if member:
-            return member
+    if op == '.':
+        if not tree.site.dml_version() == (1, 2) and rh == 'len':
+            member = try_resolve_len(tree.site, lh)
+            if member:
+                return member
+        elif isinstance(lh, TemplatesRef):
+            return mkTemplatesSubRef(tree.site, lh, rh)
+        elif isinstance(lh, TemplatesSubRef):
+            return mkTemplateQualifiedMethodRef(tree.site, lh, rh)
+        elif isinstance(lh, TraitTemplatesRef):
+            return mkTraitTemplatesSubRef(tree.site, lh, rh)
+        elif isinstance(lh, TraitTemplatesSubRef):
+            return mkTraitTemplateQualifiedMethodRef(tree.site, lh, rh)
 
     if isinstance(lh, NonValue) and not isinstance(lh, NodeRef):
         raise lh.exc()
--- a/py/dml/ctree.py
+++ b/py/dml/ctree.py
@@ -3800,9 +3800,11 @@ class TraitMethodRef(NonValue):
         return self.call_expr(args, rettype)
 
 class TraitMethodDirect(TraitMethodRef):
-    '''The DML expression 'default', evaluated inside a trait method. In
-    other words, a direct reference to a specific trait method
-    instance, as opposed to an indirect vtable lookup.'''
+    '''A direct reference to a specific trait method instance,
+    as opposed to an indirect vtable lookup. This is used for references
+    to unoverridable shared implementations, the DML expression 'default'
+    when evaluated inside a shared method, and template-qualified method
+    implementation calls to shared implementations'''
     @auto_init
     def __init__(self, site, traitref, methodref): pass
 
@@ -4330,7 +4332,7 @@ def mkTemplatesSubRef(site, templates_ref, template_name):
     if tmpl is None:
         raise ENTMPL(site, template_name)
     if tmpl not in templates_ref.node.templates:
-        raise TQMIC(site, templates_ref.node.ident, template_name)
+        raise ETQMIC(site, templates_ref.node.ident, template_name)
     return TemplatesSubRef(site, templates_ref, template_name)
 
 def mkTemplateQualifiedMethodRef(site, templates_subref, method_name):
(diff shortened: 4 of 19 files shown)


RELEASENOTES-1.4.docu
```diff
@@ -537,4 +537,29 @@
...
+      </pre>
+      Calls of this form are called <em>Template-Qualified Method Implementation
+      Calls</em>. This feature is primarily offered to make it more ergonomic to
+      write overriding methods that get defined in order to resolve conflicts
+      introduced from the usage of multiple unrelated templates that provide
+      differing implementations of the same methods. `templates` provides a
+      means for such an override to reference the implementations it overrides,
+      whereas it previously couldn't (as <tt>default()</tt> would be ambiguous.)
+      For more information, see the DML 1.4 Reference
+      Manual.</add-note></build-id>
```
lwaern-intel, line 564: Verbose. I could leave it at the first paragraph, but that would mean we don't say anything about intended usage here.

doc/1.4/language.md
```diff
@@ -3815,6 +3828,271 @@ independent method callback(int i, void *aux) {
...
+A template-qualified method implementation call done via [a value of template
+type](#templates-as-types) &mdash; most significantly, done via the (implicit)
+`this` within the body of a `shared` method &mdash; function differently in
+comparison to with compile-time object references. The specified template
+must be an ancestor template of the value's template type, the <tt>object</tt>
+template, or the template type itself; and the specified template **must
+provide or inherit a `shared` implementation of the named method**. It is not
+sufficient that the method is simply _declared_ `shared` such that it is part
+of the template type: the implementation itself must also be `shared`.
+For example, the following is permitted:
```
lwaern-intel, line 3998: This part of the TQMIC documentation goes on very long for a rather rare circumstance for anyone to need deal with. This is really just to highlight how unusual these restrictions are (that being that the method is part of the template type is _not_ enough), and how one would go about dealing with that in case you _do_ want to a non-shared implemented TQMIC from a value of template type. If I should cut this down, what should I cut away.

py/dml/structure.py
```diff
@@ -1196,8 +1204,9 @@ def default_sym(self, indices):
         ancestry_path = self.trait_node.traits.ancestry_paths[
             default_def_trait][0]
         if default_def_trait is not self.trait_method.vtable_trait:
-            ancestry_path += default_def_trait.ancestry_paths[
-                self.trait_method.vtable_trait][0]
+            # mustn't use +=!
+            ancestry_path = (ancestry_path + default_def_trait.ancestry_paths[
+                self.trait_method.vtable_trait][0])
```
lwaern-intel, line 1209: I ran into a bug I _thought_ was caused by this, but turned out it wasn't. Nevertheless, this is **dangerous**, so I kept this diff.

py/dml/structure.py
```diff
@@ -1398,9 +1411,9 @@ def process_method_implementations(obj, name, implementations,
                 typecheck_method_override(impl.method_ast,
                                           overridden.method_ast,
                                           location)
-            if (impl.method_ast.site.dml_version() == (1, 2) and throws
+            if (impl.site.dml_version() == (1, 2) and throws
                 and not overridden.throws
-                and overridden.method_ast.site.dml_version() == (1, 4)):
+                and overridden.site.dml_version() == (1, 4)):
```
lwaern-intel, line 1416: The refactoring for this region to use `impl.site` instead of `impl.method_ast.site` was due to a crash from the `EAMBINH` 1.2 test, where `overridden` happened to be `TraitMethodHandle`, which doesn't have `method_ast`. I traced back `impl.site` vs `impl.method_ast.site`, and found that `impl.site` always corresponded to the site of the method body AST, so I just refactored to use `impl.site` directly instead.

py/dml/structure.py
```diff
@@ -1426,12 +1439,21 @@ def process_method_implementations(obj, name, implementations,
                     raise EMETH(impl.site, overridden.site,
                                 "different nothrow annotations")
 
+        template = (impl.obj_spec.parent_template
+                    if isinstance(impl.obj_spec, InstantiatedTemplateSpec)
+                    else None)
```
lwaern-intel, line 1444: This one thing is the entire reason I've refactored `MethodHandle`s  to carry around `obj_spec`:s rather than `rank`s.

py/dml/structure.py
```diff
@@ -1457,7 +1479,7 @@ def process_method_implementations(obj, name, implementations,
 # Second stage of object creation, protected by error context
 # TODO: we should probably split up mkobj into several methods with
 # clearer responsibilities
-def mkobj2(obj, obj_specs, params, each_stmts):
+def mkobj2(obj, obj_specs, params, each_stmts, used_templates):
```
lwaern-intel, line 1482: Oh, I don't use `used_templates` anymore as part of `mkobj2`, only `mkobj`. The parameter for it can be tossed away.

py/dml/structure.py
```diff
@@ -2977,7 +3025,10 @@ def mkmethod(site, rbrace_site, location, parent_obj, name, inp_ast,
                       'Methods with (partially) const output/return '
                       + 'values are not yet supported.')
     if default_level:
-        name += "___default%d" % (default_level,)
+        if template:
+            name = f'templates__{template.name}__{name}'
+        else:
+            name += f'___default{default_level}'
```
lwaern-intel, line 3031: Figured this would make generated C easier to parse, especially for GCC or Coverity errors.

py/dml/template.py
```diff
@@ -123,15 +128,28 @@ def defined_symbols(self):
                     assert sub.kind == 'error'
             for (_, name, _, specs) in composite:
                 symbols[name] = ('subobj', specs.site)
+        self._defined_symbols = symbols
```
lwaern-intel, line 131: Memoization is cause of the use in `mkTraitTemplateQualifiedMethodRef`. I suppose `mkTraitTemplateQualifiedMethodRef` could leverage the new `get_potential_method_impl_details` instead?

py/dml/traits.py
```diff
@@ -568,8 +607,7 @@ def __init__(self, ancestors, ancestor_vtables):
         # vtable the method or parameter appears
         self.ancestor_vtables = ancestor_vtables
 
-    @property
-    def ancestors(self): return self.ancestry_paths.keys()
+        self.ancestors = set(self.ancestry_paths.keys())
```
lwaern-intel, line 610: This change is purely to avoid `merge_method_impl_maps()` from being pathologically `O(n^4)` in behavior. That ordo was too high for me to be comfortable.

doc/1.4/language.md
```diff
@@ -3815,6 +3828,271 @@ independent method callback(int i, void *aux) {
...
+  replicated without the need for excessive boilerplate.
+
+The following is an example of the first case:
+```
+template simple_write_alter is write {
+    method write(uint64 written) {
+        default(alter_write(written));
+    }
+
+    method alter_write(uint64 curr, uint64 written) -> (uint64);
```
mandolaerik, line 3892: inconsistency: alter_write vs simple_alter_write. I suggest using alter_write consistently, `simple` is redundant

py/dml/structure.py
lwaern-intel: And is also why `InstantiatedTemplateSpec` exists.

doc/1.4/language.md
lwaern-intel: :thumbsup:

doc/1.4/language.md
```diff
@@ -3815,6 +3828,271 @@ independent method callback(int i, void *aux) {
...
+}
+```
+
+A template-qualified method implementation call is resolved by using
+the method implementation provided to the object by the named template.
+If no such implementation is provided (whether it be because the template does
+not specify one, or specifies one which is not provided to the object due to its
+definition being eliminated by an [`#if`](#conditional-objects)), then the
+ancestor templates of the named template are recursively searched for the
+highest-ranking (most specific) implementation provided by them. If the ancestor
```
mandolaerik, line 3983: I think s/ranking/rank/ is how we refer to it elsewhere. s/ranking/ranked/ is also fine.

doc/1.4/language.md
```diff
@@ -3815,6 +3828,271 @@ independent method callback(int i, void *aux) {
...
+templates provide multiple hierarchically unrelated implementations, then the
+choice is ambiguous and the call will be rejected by the compiler. In this case,
+the modeller must refine the template-qualified method implementation call to
+name the ancestor template whose implementation they would like to use.
+
+A template-qualified method implementation call done via [a value of template
+type](#templates-as-types) &mdash; most significantly, done via the (implicit)
+`this` within the body of a `shared` method &mdash; function differently in
```
mandolaerik, line 3991: `s/function/functions/`, also can the sentence be split in two?

syssimics:
Verification [#13366627](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13366627): pass

doc/1.4/language.md
```diff
@@ -3815,6 +3828,271 @@ independent method callback(int i, void *aux) {
...
+comparison to with compile-time object references. The specified template
+must be an ancestor template of the value's template type, the <tt>object</tt>
+template, or the template type itself; and the specified template **must
```
mandolaerik, line 3994: either `s/; and/, and/` or `s/; and /; furthermore, /`

doc/1.4/language.md
lwaern-intel: As in:
```
A template-qualified method implementation call done via [a value of template
type](#templates-as-types) function differently in comparison to with compile time
object references. In particular, `(this).templates` within the bodies of `shared` methods
function differently.
```
? Maybe... Do you prefer the above?

doc/1.4/language.md
lwaern-intel: yeah i struggled with this. I used `;` instead of `,` because I wanted a clear separation between the two conditions, though maybe that's not needed.

doc/1.4/language.md
lwaern-intel: Yeah, it's highest-rank.

py/dml/ctree.py
```diff
@@ -4297,6 +4307,191 @@ def read(self):
...
+            rank_to_candidate[tmpl.spec.rank] = (tmpl, candidate)
+            break
+
+        if implements_method:
+            candidate = node.template_method_impls.get((tmpl, method_name))
+            if candidate is not None:
+                rank_to_candidate[tmpl.spec.rank] = (tmpl, candidate)
+                continue
+
+        pending_templates.update(filter(lambda c: c not in studied_templates,
```
mandolaerik, line 4391: `update(c for c in next_remplates if c not in studied_templates)` ?

lwaern-intel: I suppose it is ever-so-slightly less verbose... goddamit `filter` is so hard to find an excuse to use

py/dml/ctree.py
```diff
@@ -4297,6 +4307,191 @@ def read(self):
...
+                impl_template, shared_method)
+
+
+    studied_templates = set()
+    pending_templates = Set((dml.globals.templates[template_name],)
+                            # don't bother going through regular methods
+                            # if if some shared method impl. is
+                            # unoverridable (can only be one if so)
+                            * (shared_method is None
+                               or shared_method.overridable))
```
mandolaerik, line 4363: I think explicit control flow would improve readability drastically:
```
if not shared or overridable:
   pending = [one_element]
   studied = set()
   while pending:
      ...      
```

Also, possibly more straightforward to keep pending_templates a list, and eliminate duplicates on pop instead (I prefer native Python types when I can). But I don't insist.

lwaern-intel: God no, that incurs another indendation level

lwaern-intel: > Also, possibly more straightforward to keep pending_templates a list, and eliminate duplicates on pop instead (I prefer native Python types when I can). But I don't insist.
I could be OK with that.

py/dml/ctree.py
```diff
@@ -4297,6 +4307,191 @@ def read(self):
...
+    # ... in which case the candidate implementation of Z should be excluded
+    from .traits import calc_minimal_ancestry
+    candidates = [rank_to_candidate[r]
+                  for r in calc_minimal_ancestry(Set(rank_to_candidate))[None]]
+
+    if not candidates:
+        raise EMEMBER(site, templates_subref, method_name)
+
+    if len(candidates) > 1:
+        raise EAMBTQMIC(site, template_name, method_name, candidates)
```
lwaern-intel, line 4409: Fun fact: I originally had these `raise`s as `report`s, but that caused these errors to be reported multiple times because of `try_codegen_invocation`, and I figured that was not worth it.

py/dml/ctree.py
```diff
@@ -4297,6 +4307,191 @@ def read(self):
...
+
+    (tmpl, method) = candidates[0]
```
lwaern-intel, line 4411: I suspect `rank_to_candidates` and `candidates` having their values be tuples instead of just the implementation is an old artifact?

lwaern-intel: Oh, no, the template part is used for the error messages, nevermind

py/dml/ctree.py
```diff
@@ -4297,6 +4307,191 @@ def read(self):
...
+
+    if isinstance(method, objects.Method):
+        return mkNodeRef(site, method, indices)
+    else:
+        impl_trait = method.trait
+        ancestry_path = node.traits.ancestry_paths[impl_trait][0]
+        if impl_trait is not method.vtable_trait:
+            # mustn't use +=!
+            ancestry_path = ancestry_path + impl_trait.ancestry_paths[
+                method.vtable_trait][0]
```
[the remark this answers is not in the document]
lwaern-intel: Ahh, yes, they are! I got confused -- the inner items of ancestry_paths are tuples, but the outer are lists:
```
        self.ancestry_paths = {
            p: ([(p,)] if p in self.direct_parents
                else [(dp,) + path
                      for dp in self.direct_parents
                      if p in dp.ancestry_paths
                      for path in dp.ancestry_paths[p]])
            for p in ancestors}
```

In that case I'll just revert the changes.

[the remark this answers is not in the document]
lwaern-intel: ... or so I say, but the usage of `+=` on tuples does smack of being evil to me... errrgh...... I mean, if you do it, you almost do need a comment explaining why it's safe. Ahh, the wonders of Python being a dynamically typed language.

py/dml/template.py
```diff
@@ -144,6 +162,49 @@ def traits(self):
             return Set().union(
                 *[sup.traits() for (_, sup) in self.spec.templates])
 
+    def get_potential_method_impl_details(self, method_name):
```
mandolaerik, line 165: Need docstring/comments, in particular the contract for the return value.

If it has no viable use outside tqmic, then possibly better expressed as a top-level function.

lwaern-intel: Oh right, I started writing a docstring, but it ended up needing so many revisions that I said "y'know what, I'll save the docstring for when I have actually figured this out" and then I forgot about it

lwaern-intel: > If it has no viable use outside tqmic, then possibly better expressed as a top-level function.
It is not _hyper_-specific to TQMIC, but it could be that it is too specific. What this method essentially does (through repeated calls of it) is provide you with a tree of the current and ancestor templates, filtered by which may provide the highest-rank implementation of the named method, ("may" because the definition could be conditionally eliminated). It's possible that this functionality can be generalized to param definitions as well.
That is interesting if you need to get an idea of the possible implementations that a specific template provides, which can possibly be interesting for static analysis? Or if you need information about all method implementations provided by templates to to an object, and not just the final implementations the object has, which is what TQMIC uses it for.

py/dml/codegen.py
```diff
@@ -1260,10 +1260,19 @@ def try_resolve_len(site, lh):
...
-    if not tree.site.dml_version() == (1, 2) and op == '.' and rh == 'len':
-        member = try_resolve_len(tree.site, lh)
-        if member:
-            return member
+    if op == '.':
+        if not tree.site.dml_version() == (1, 2) and rh == 'len':
+            member = try_resolve_len(tree.site, lh)
+            if member:
+                return member
+        elif isinstance(lh, TemplatesRef):
```
mandolaerik, line 1268: Perhaps `s/elif/if`, to cover this evil corner:
```
template len {
  method len() {}
}
group g is len;
...
g.templates.len.len();
```

mandolaerik: Also, this looks like a good opportunity to try the new `match` statement?

py/dml/codegen.py
```diff
@@ -1260,10 +1260,19 @@ def try_resolve_len(site, lh):
...
+            return mkTemplatesSubRef(tree.site, lh, rh)
+        elif isinstance(lh, TemplatesSubRef):
+            return mkTemplateQualifiedMethodRef(tree.site, lh, rh)
+        elif isinstance(lh, TraitTemplatesRef):
+            return mkTraitTemplatesSubRef(tree.site, lh, rh)
+        elif isinstance(lh, TraitTemplatesSubRef):
+            return mkTraitTemplateQualifiedMethodRef(tree.site, lh, rh)
```
mandolaerik, line 1275: I think other similar things are resolved in mkSubRef.

Looking at mkSubRef raises another corner:
```
template x {
  method x() {}
}
template y { is x; }
group g is y;
...
local y y = cast(g, y);
local y *yp = &y;
// should we choke on `->`?
local x x = yp->template.x;
```

lwaern-intel: > I think other similar things are resolved in mkSubRef
Actually no. `mkSubRef` resolves that which is type-driven, while `expr_member` resolves that which is  expression-class-driven. That said there is no good reason for that discrepency, so everything that's in `expr_member` may as well be moved to `mkSubRef`.
```
local x x = yp->templates.x;
```
We may as well permit it. Easy with the move into `mkSubRef`.

doc/1.4/language.md
mandolaerik: One idea is to move details into the docstring of the corresponding error message(s) in messages.py, if that's applicable. Those are in an appendix, where lengthy details are less of a concern. People are unlikely to explore that appendix by themselves, so the main text should probably contain forward-links if they provide interesting info.

lwaern-intel: Not too bad of an idea, other than it makes reading the source of `messages.py` a strange experience when one docstring is suddenly miles longer than the others.

py/dml/structure.py
lwaern-intel: not actually a bug. Will be replaced with a comment explaining the use of `+=` is safe.

py/dml/traits.py
mandolaerik: really? I'm pretty sure `dict.keys()` is a lazy O(1) view of the key set (class `dict_keys`), offering O(1) `in` and O(n) iteration. Unlike `set`, it also preserves iteration order.

py/dml/codegen.py
lwaern-intel: I wondered "how" but then I found out that `isinstance` chains can be replaced by so-called "class patterns". Oh yes, that is a good idea.

py/dml/traits.py
lwaern-intel: Oh hey, you're right! I tried to research about this and drew the conclusion that `dict_keys` was just an iterator, but I've experimentally it does seem to be `O(1)`. Clearly my research failed.

doc/1.4/language.md
mandolaerik: Then I suggest `furthermore`.

py/dml/ctree.py
mandolaerik: > God no, that incurs another indendation level

I object! high-level control flow is more important than indent level. If indent level is too deep, then it's time to factor out a function. "repeat until this list, that starts at size 1 or 0 depending on XYZ, is empty" is not a natural way of thinking.

py/dml/template.py
mandolaerik: > It's possible that this functionality can be generalized to param definitions as well.

I bet we will get a feature request for TQPIC pretty soon, but I have a feeling that we should handle that with a separate modified copy of the code path for methods (methods and params are too dissimilar for that kind of abstraction).

py/dml/ctree.py
```diff
@@ -4297,6 +4307,191 @@ def read(self):
...
+            rank_to_candidate.clear()
+            if next_templates:
+                [tmpl] = next_templates
+            elif not implements_method:
+                break
+
+            candidate = node.template_method_impls.get((tmpl, method_name))
+            assert candidate is not None
+            rank_to_candidate[tmpl.spec.rank] = (tmpl, candidate)
+            break
```
mandolaerik, line 4383: I didn't grasp why we need to care about `final` vs `default` here. Non-default should *only* mean that overrides are banned, it should not affect how symbols are resolved. It does carry some (horrible) semantics in 1.2, and sometimes it's used to guess a hint in an error message, but this is not what happens here.

py/dml/ctree.py
mandolaerik: Indeed, `raise` often gives "remove current object and continue" as fallback, which is good enough for our purposes (to combine many error tests in one file).

py/dml/template.py
mandolaerik: why don't you use `@functools.cache` to memoize?

py/dml/codegen.py
lwaern-intel: Actually `local x x = yp->templates.x;` is already permitted; that is actually resolved via `mkSubRef`. Thus the only question is whether the non-type-driven components should be part of `mkSubRef`, and I'm thinking "why not?"

py/dml/ctree.py
lwaern-intel: uurggh. Fine. I wanted to avoid indentation due to the long name of `get_potential_method_impl_details`, which would incur horrible-to-read-and-resolve linebreaks, but it turns out I can just _barely_ avoid the worst cases.

py/dml/template.py
lwaern-intel: cause i didn't know it existed

py/dml/codegen.py
lwaern-intel: nevermind fisketur hates `match` :(

py/dml/ctree.py
lwaern-intel: The docstring will hopefully clear things up as to what `final` actually means.

py/dml/codegen.py
lwaern-intel: I've made the move to `mkSubRef`. Resolve this if you're satisfied.

py/dml/codegen.py
lwaern-intel: Evil corner covered. Won't be using `match`.

py/dml/ctree.py
lwaern-intel: Check the docstring, see if that clears things up for you now.

py/dml/ctree.py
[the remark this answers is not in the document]
lwaern-intel: Added comments explaining tuples are used.

py/dml/ctree.py
lwaern-intel: Irrelevant now with other changes.

doc/1.4/language.md
lwaern-intel: done.

doc/1.4/language.md
lwaern-intel: Done.

py/dml/traits.py
lwaern-intel: Reverted.

py/dml/template.py
lwaern-intel: `@functools.cache` now used. `mkTraitTemplateQualifiedMethodRef` now uses `get_potential_method_impl_details()`, so the memoization for `defined_symbols()` has been completely removed.

doc/1.4/language.md
lwaern-intel: Not done yet, as it's not clearly a better solution.

syssimics:
Verification [#13379755](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13379755): pass

doc/1.4/language.md
```diff
@@ -3815,6 +3828,271 @@ independent method callback(int i, void *aux) {
...
+}
+
+in each (gated_register, very_complex_register) {
+    is very_complex_gated_register;
+}
+```
+
+A template-qualified method implementation call is resolved by using
+the method implementation provided to the object by the named template.
+If no such implementation is provided (whether it be because the template does
```
mandolaerik, line 3979: Either `whether it is because` or `be it because`, I think

doc/1.4/language.md
mandolaerik: I think it's OK to make messages.py weird, I'm more concerned that it might make the refmanual look weird. I guess one way to find out is to try moving it and see if it does look outrageous.

And as a general direction, it would not disturb me if we'd move docs on error corners to messages.py.

doc/1.4/language.md
lwaern-intel: A googling says "whether it be" is correct grammar. Your options would also work (though _maybe_ not `whether it is because`? My English skills are failing me here)

py/dml/template.py
mandolaerik: `functools.cache` makes it feel natural to keep it a method (although top-level would still work of course).

py/dml/template.py
```diff
@@ -144,6 +156,88 @@ def traits(self):
...
+          an implementation of the method (which may be conditionally provided)
+        * `next_candidates`: A list of all hierarchically unrelated ancestor
+          templates that specify (and thus may provide) a highest-rank
+          implementation of the named method when excluding the current
+          template. If the current template unconditionally provides an
+          implementation, this list will be empty.
+
+        Example return values of `t.get_potential_method_impl_details(m)` and
+        their meaning:
+        * `(True, True, [])`
```
mandolaerik, line 183: Given `functools.cache`, safer to make `next_candidates` a tuple.

py/dml/template.py
```diff
@@ -144,6 +156,88 @@ def traits(self):
             return Set().union(
                 *[sup.traits() for (_, sup) in self.spec.templates])
 
+    @functools.cache
+    def get_potential_method_impl_details(self, method_name):
```
mandolaerik, line 160: I know you hate type annotations, but they are a quite nice way to document return tuples.

py/dml/template.py
lwaern-intel: Nice catch!

py/dml/template.py
lwaern-intel: errrgggghhhh....

i guess. I'm fine with type annotations as long as we have no illusions that they are anything more _but_ documentation

py/dml/template.py
lwaern-intel: This is for you to decide; I'd rather keep it a method. Moving it to top-level would be due to your insistence.

doc/1.4/language.md
lwaern-intel: Alright, gonna do it then.

py/dml/template.py
lwaern-intel: Done.

py/dml/template.py
lwaern-intel: Done.

doc/1.4/language.md
lwaern-intel: Done. You'll have to run `./documentation` and see what the result looks like. In my opinion, it _does_ stick out quite a bit in the Messages section, _but_ it is acceptable.

syssimics:
Verification [#13390846](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13390846): pass

syssimics:
Verification [#13392212](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13392212): fail

lwaern-intel:
Verification failure's unrelated

doc/1.4/language.md
```diff
@@ -3249,6 +3249,176 @@ local (int i, uint8 j) = m(e1);
...
+object references. In particular, `this.templates` within the bodies of `shared`
+methods functions differently. The specified template must be an ancestor
+template of the value's template type, the <tt>object</tt> template, or the
+template type itself; furthermore, the specified template **must provide or
+inherit a `shared` implementation of the named method**. It is not sufficient
+that the method is simply _declared_ `shared` such that it is part of the
+template type: the implementation itself must also be `shared`. For more
+information, see the documentation of the [`ENSHAREDTQMIC` error
+message.](messages.html#ENSHAREDTQMIC)
+
```
lwaern-intel, line 3421: This diff is 2 things: I moved the TQMIC subsubsection to the Method Statements subsection (after Method Calls), and I moved out the final part about resolving `ENSHAREDTQMIC` to the documentation of `ENSHAREDTQMIC`, replacing it with a forward reference to that documentation.

test/1.4/expressions/T_tqmic.dml
```diff
@@ -0,0 +1,90 @@
...
+        return 4.0;
+    }
+}
+
+group g1 is w;
+group g2 is w {
+    param enable_u3 = false;
+}
+
+method init() {
```
mandolaerik, line 63: Perhaps also test e.g. `(&g.templates.w.sm)(dev.obj)`? Or is that banned for now?

py/dml/traits.py
```diff
@@ -417,51 +421,109 @@ def qualifier_check(qualifier_name, qualifier0, qualifier1):
...
+
+def get_highest_ranks(ranks):
+    '''Given a set of ranks, return the subset of highest unrelated ranks'''
+    if len(ranks) <= 1:
+        return ranks.union()
+    # .intersection(ranks) is redundant, but may serve to avoid creating large
+    # intermediate sets
+    return ranks.difference(*(r.inferior.intersection(ranks) for r in ranks))
+
+def calc_minimal_ancestry(ranks):
```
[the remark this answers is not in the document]
mandolaerik: It's a nice change (also makes it easy to unit test, but no need to do that today)

test/1.4/expressions/T_tqmic.dml
lwaern-intel: It's not, but why? There's nothing special about a TQMIC once it's resolved. At that point it's just a `NodeRef` to a method or a `TraitMethodDirect`, and you can do anything with it you could with a regular method reference.

lwaern-intel: Everything that makes TQMIC special has to do with the resolution, and that's what the tests... test.

mandolaerik (approved):
I would suggest calling it something like "Template Qualified Method Resolution" or perhaps "Explicit Method Resolution" instead. But that naming discussion does not block a merge.

RELEASENOTES-1.4.docu
mandolaerik: After the example, you might just wrap up with something like "this is sometimes useful to resolve conflicts in multiple inheritance"

lwaern-intel: How about...
> Calls of this form are called <em>Template-Qualified Method Implementation Calls</em>, and are primarily meant to make it easier to resolve conflicts in multiple inheritance: it allows any method that overrides multiple implementations from unrelated templates to individually reference and call the implementations it overrides.

py/dml/ctree.py
```diff
@@ -4386,7 +4583,25 @@ def mkSubRef(site, expr, sub, op):
         else:
             raise EREF(site, sub, expr)
 
-    elif isinstance(expr, AddressOf) and op == '->':
+    if isinstance(expr, NonValue):
+        if op == '.':
+            if not site.dml_version() == (1, 2) and sub == 'len':
```
mandolaerik, line 4588: `site.dml_version() != (1, 2)`?

lwaern-intel: Oh. Sure. That's an inherited issue, but no reason not to address it.

py/dml/ctree.py
```diff
@@ -4386,7 +4583,25 @@ def mkSubRef(site, expr, sub, op):
         else:
             raise EREF(site, sub, expr)
 
-    elif isinstance(expr, AddressOf) and op == '->':
+    if isinstance(expr, NonValue):
+        if op == '.':
```
mandolaerik, line 4587: Untested corner:
```
some_template t = cast(obj, some_template);
some_template *tp = &t;
/// ERROR ENVAL (?)
tp->templates.parent_template.m
```

doc/1.4/language.md
mandolaerik: ok, then the lack of English skills is probably on my side.

py/dml/template.py
mandolaerik: The purpose of my previous comment was to lift my insistence. Method is better.

RELEASENOTES-1.4.docu
mandolaerik: Yes, that's better.

py/dml/ctree.py
lwaern-intel: What part of that is untested, specifically? That you can do `tp->templates` is tested in T_tqmic, and that a resolved TQMI ref gives ENVAL I don't feel needs to be tested. As stated elsewhere it's just a `NodeRef`/`TraitMethodDirect`.

syssimics:
Verification [#13396592](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13396592): pass

py/dml/ctree.py
mandolaerik: Sorry for the noise! I misunderstood the code and thought it would give an error, but didn't see a matching error test. Indeed there is a positive test for it.

PR #279, opened 2024-05-03
File-local provisionals and explicit param declarations and overrides  

mandolaerik:
- **Use compat features to deprecate bad version statements**
- **Improve wording**
- **Use package_path from Simics base**
- **Refactor grammar for 'device'**
- **Add base syntax for provisional features**
- **Add syntax for walrus params**
- **Unify ASTs for param and typedparam**
- **Report errors for walrus violations**


mandolaerik:
The commits from #277 were accidentally included (because I need some of them locally). Please review 277 before approving this one, to reduce noise.

syssimics:
Verification [#13326740](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13326740): fail

syssimics:
Verification [#13337244](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13337244): fail

syssimics:
Verification [#13340724](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13340724): fail

lwaern-intel (commented):
This PR contains a lot of different changes, and I can't really tell which relate to what. How many of these do you need for the walrus params & introduction of provisionals? Are there any parts that can be split to a separate PR without one blocking the other? I'm especially curious to the many changes to the error tests you've made -- what are they about?

py/dml/dmlparse.py
```diff
@@ -892,7 +900,8 @@ def object_statements_empty(t):
 @prod
 def object_statement(t):
     '''object_statement : object_statement_or_typedparam'''
-    if t[1].kind == 'typedparam':
+    if (t[1].kind == 'param' and t[1].args[1] is not None
+        and t[1].args[1].kind == 'paramtype'):
         report(ESYNTAX(site(t), ':',
```
lwaern-intel, line 905: `s/':'/t[1].args[1]`

syssimics:
Verification [#13346906](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13346906): fail

py/dml/messages.py
```diff
@@ -1138,6 +1138,25 @@ class EAUTOPARAM(DMLError):
     library, and they may not be overridden."""
     fmt = "bad declaration of automatic parameter '%s'"
 
+class ENWALRUS(DMLError):
+    """When the `walrus_params` provisional feature is enabled, parameter
+    declarations that do not override existing declarations must be
+    declared using the `:=` or `:default` syntax.
+    """
+    fmt = "non-overriding param should use :%s syntax (from walrus_params feature)"
```
lwaern-intel, line 1146: This error message is the **key feature** of walrus params, and the message should play to that.
```
fmt = "parameter not previously declared: '%s'. To declare and define a new parameter, use the ':%s' syntax (this requirement is due to the 'walrus_params' provisional feature being enabled)."
```

py/dml/messages.py
```diff
@@ -1138,6 +1138,25 @@ class EAUTOPARAM(DMLError):
...
+
+class EWALRUS(DMLError):
+    """When the `walrus_params` provisional feature is enabled, parameters
+    declared using `:=` or `:default` syntax may not override other parameters.
+    """
+    fmt = "use of :%s to override existing param (from walrus_params feature)"
```
lwaern-intel, line 1152: `(the :%s syntax is provided by the walrus_params provisional feature)`, possibly with `... provisional feature, which has been enabled)`

py/dml/provisional.py
```diff
@@ -0,0 +1,65 @@
...
+
+def feature(cls: type[ProvisionalFeature]):
+    assert issubclass(cls, ProvisionalFeature)
+    singleton = cls()
+    features[singleton.tag()] = singleton
+    return singleton
+
+
+@feature
+class walrus_params(ProvisionalFeature):
```
lwaern-intel, line 31: I'm not fond of the name! Walrus params to me was an internal naming scheme.
Not that I'm not sure of a better name... maybe `explicit_param_overrides`? Or, if it's not too long, `explicit_param_decls_and_overrides`.

py/dml/messages.py
```diff
@@ -1639,6 +1658,13 @@ class ELOGGROUPS(DMLError):
     fmt = ("Too many loggroup declarations. A maximum of 63 log groups (61 "
            + "excluding builtins) may be declared per device.")
 
+
+class ENOPROV(DMLError):
+    """
+    An invalid identifier was passed in the `provisional` statement.
+    """
+    fmt = "No such provisional feature %s. Valid values are: %s"
```
lwaern-intel, line 1666: I'd rather direct people to `--help-provisional`... once we've added that.

lwaern-intel (changes requested):
Partial review before signing off, haven't looked into the implementation in depth. Will do a more in-depth review later.

py/dml/provisional.py
lwaern-intel: What am I thinking, the best would be `explicit_param_decls`.

py/dml/messages.py
lwaern-intel: By the way, I just realized that the following isn't illegal:
```
template x {
    param p = 4; // := with the new feature
}

template y is x {
    param p;
}

is y;
```
Perhaps it should be? I'm not sure. There's also the interplay with late shared, we don't want to make:
```
template x {
    param p = 4; // := with the new feature
}

template y is x {
    param p : int;
}

is y;
```
illegal.

lwaern-intel: Errgh... late shared really does make this _weird_, doesn't it. Ideally we want `:=` to not be allowed to be used as an _override_, and yet...

```
template x {
    param p :default 4;
}

template y is x {
    param p : int;
    param p = 7;
}
```
Should be legal. If walruses truly are just syntactic sugar, why shouldn't
```
template y is x {
    param p : int = 7;
}
```
also be legal? And, by extension,

```
template y is x {
    param p := 7;
}
```
?

I think the answer should be that `:=` is only _mostly_ syntactic sugar, but does add an additional restriction that the multi-statement form does not: that the the param must not have been previously declared.

mandolaerik:
What remains is documentation for provisionals, unclear if that blocks merge.

syssimics:
Verification [#13399015](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13399015): fail

lwaern-intel:
> What remains is documentation for provisionals, unclear if that blocks merge.

And a passing verification!

When it comes to documentation, I'd like to see it both for provisionals as well as expl_param_decls (as per our R.C. discussions)

py/dml/messages.py
mandolaerik: I don't think we should add `--help-provisional`: provisionals are enabled only by syntax and thereby technically a part of the language, and describing the language is out of scope for help flags.

We may need to introduce flag-enabled provisionals later, but I hope not.

syssimics:
Verification [#13401317](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13401317): fail

py/dml/structure.py
```diff
@@ -608,22 +636,47 @@ def is_default(decl):
...
+                    other for (other_rank, other) in decls + defs
+                    if other_rank in rank.inferior] + [
+                    other for (other_rank, other) in decls
+                    if other_rank is rank]
+                report(EOVERRIDE(type_info.site, type_info.site,#parent.site,
+                               'default' if is_default else '='))
+            elif (not is_override and declared_as_override
+                  and not any(r is rank for (r, _) in decls)):
+                report(ENOVERRIDE(
+                    p.site, 'default' if is_default else '='))
```
lwaern-intel, line 664: This entire loop strikes me as very messy. I think this could greatly benefit from the newly broken out `calc_minimal_ancestry`:
```
rank_to_decl_or_def = {r: spec for (r, spec) in itertools.chain(decls, defs) }
ancestry = calc_minimal_ancestry(Set(rank_to_decl_or_def))

for (rank, p) in defs:
    if p.site.provisional_enabled(provisional.explicit_param_decls):
        overrides = list(ancestry[rank])
        if not declared_as_override and overrides:
                report(EOVERRIDE(type_info.site, overrides[0], 'default' if is_default else '='))
        ....
```
... Actually, maybe the entirety of `merge_parameters` could be simplified by this.

py/dml/structure.py
```diff
@@ -2711,11 +2766,12 @@ def mkexpr(self, indices):
...
 
 def mkparam(obj, autoparams, param):
-    _, site, name, (value, default, auto) = param
+    (_, site, name, type_info, is_default, value) = param
+    auto = type_info is not None and type_info.kind == 'auto'
 
     # auto-parameters are like normal "hard" assignments
     if name in autoparams:
-        if not auto and not default:
+        if not auto and not (dml.globals.dml_version == (1, 2) and is_default):
```
lwaern-intel, line 2774: If I'm interpreting this right, 1.2 abuses `default` declarations and possibly overrides them with autoparams? Wow.

test/1.4/provisional/T_bare.dml
```diff
@@ -0,0 +1,17 @@
+/*
+  © 2024 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+
+dml 1.4;
+
+/// COMPILE-ONLY
+
+provisional;
```
lwaern-intel, line 10: huh. I guess that makes sense. I just checked, Haskell permits a bare `{-# LANGUAGE #-}` too.

lwaern-intel (changes requested):
Besides this, there are two things I want us to do, or at least come to an agreement about:
- Documentation
- How we would offer device-wide provisionals, as that would affect documentation

RELEASENOTES.docu
```diff
@@ -1771,4 +1771,8 @@ extern typedef struct { } my_type_t;</pre> </add-note></build-id>
       m(conf_object_t)</tt> for function pointers is only permitted
       when this feature is enabled. The feature will be disabled in
       Simics API versions > 7.</add-note></build-id>
+  <build-id _6="next" _7="next"><add-note>Fixed a bug in
+      the <tt>dead_dml_methods</tt> library, which caused it to never
+      detect methods as dead if they occur directly after
+      a <tt>footer</tt> block.</add-note></build-id>
 </rn>
```
lwaern-intel, line 1778: We want a releasenote for the provisional mechanism, documenting its addition (because yes, it deserves it, in the same way `--compat` deserved it.) At the very least, we want it at the same time we offer documentation for provisionals, if we don't do that right away. Seperately we should have a releasenote for `explicit_param_decls`. (These both belong to 1.4's releasenotes.)

lwaern-intel (changes requested):

syssimics:
Verification [#13401431](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13401431): pass

test/1.4/provisional/T_bare.dml
mandolaerik: There is no intended use case, this is just a grammatical edge case; I added the test just to keep an unimportant corner stable.

py/dml/structure.py
mandolaerik: It's from `parameter indexvar default undefined;`.

BTW: This means that 1.2 incidentally allows user-provided definitions of autoparams, e.g.:
```
group g {
    parameter indexvar default 18;
    parameter bank default 3;
}
```
(where the indexvar decl will give `WEXPERIMENTAL: parameter with two default declarations` and have effect, whereas the bank decl will be silently swallowed and have no effect).

IIRC I once suggested to make it a proper auto but others (probably Jonatan) deemed that unimportant. And unimportant it is, but since you bothered to comment it I'll push another commit that cleans it up.

RELEASENOTES.docu
mandolaerik: A release note will come with its documentation.

py/dml/provisional.py
```diff
@@ -0,0 +1,79 @@
...
+
+    In some rare cases, you may need to declare a parameter without
+    knowing if it's an override or a new parameter. In this case, one
+    can accompany a `param NAME = value;` or `param NAME default
+    value;` declaration with a `param NAME;` declaration in the same
+    scope/rank. This marks that the parameter assignment may be either
+    an override or a new parameter, and no error will be printed.
+
+    Enabling the `explicit_param_decls` feature in a file only affects
+    parameter declarations in that file.
```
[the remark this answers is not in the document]
mandolaerik: My intent was to clarify the bounds of the checks; e.g., if a param is declared with `:=`, it's still OK to override it with `=` in a file that didn't enable the provisional.

py/dml/logging.py
```diff
@@ -290,6 +292,8 @@ def dml_version(self):
         return self._dml_version
     def bitorder(self):
         return 'le'
+    def provisional_enabled(self, feature):
+        return False
```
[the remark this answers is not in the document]
mandolaerik: Why not? It's pretty harmless to assume that compiler-inserted statements use standard semantics. If this code path ever is taken and we have an ICE there, then I'm pretty sure we'll just insert `if feature in {...}: return False` to silence it, which won't contribute to our wisdom.

But I can surely ICE it if you insist.

py/dml/logging.py
```diff
@@ -270,6 +270,8 @@ def lineno(self): pass
     @abc.abstractmethod
     def dml_version(self): pass
     @abc.abstractmethod
+    def provisional_enabled(self, feature): pass
```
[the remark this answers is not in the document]
mandolaerik: We can certainly let `provisional` keep them in a `frozenset` and return that. I don't think that would be significantly better or worse; I weakly prefer the current but can change if you insist.

syssimics:
Verification [#13402763](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13402763): pass

syssimics:
Verification [#13403313](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13403313): pass

py/dml/logging.py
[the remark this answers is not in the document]
lwaern-intel: We sometimes abuse `SimpleSite` for other purposes, don't we? That it's only used for compiler-inserted statements, I think, is at least not something we've firmly established.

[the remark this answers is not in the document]
mandolaerik: It should be established that statements that come from DML files don't use SimpleSite. And file-local provisionals are only applicable to statements from DML files. E.g., SimpleSite is used for user-defined expressions like `-D` (`SimpleSite('<command-line>')`), where file-local provisionals should not apply.

py/dml/logging.py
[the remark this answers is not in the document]
lwaern-intel: Naah. Do as you prefer.

py/dml/logging.py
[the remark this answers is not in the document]
lwaern-intel: Looking through the usages of `SimpleSite` in DMLC, I find that questionable. I'd prefer you ICE it unless it turns it gets triggered in a way that can only be fixed by suprressing it at a centralized site that renders the ICE effectively worthless.

syssimics:
Verification [#13411421](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13411421): ✅ pass

py/dml/structure.py
lwaern-intel: @mandolaerik I've seen your new code -- how come you are still clumsily determining a parent via a list comprehension? I don't get it. Why would you not use the same or a similar pattern as I exemplify in this comment?

lwaern-intel: oh, maybe i'm confusing. In the above comment code: `s/overrides/overridden`. I meant in the sense "that which p overrides", but that is not a good choice of words

mandolaerik: The strongest reason is that `list(ancestry[rank])[0]` depends on set iteration order. Also, your suggestion reduces to something like:
```
parent = dict(itertools.chain(decls.items(), defs.items()))[next(iter(minimal_ancestry[rank]))]
```
which I find harder to read.

lwaern-intel: With the code as-is you might as well not use `calc_minimal_ancestry` at all. I don't buy the "if don't like it cause if I minimize the code it becomes ugly" argument; then don't minimize the code. This does not affect generated C, so the undefined set iteration order argument is weak; but does carry the most weight.
What about reworking `Rank` to operate in terms of `Set`? Would the induced slowdown be too much? If so, if you insist on not using the values of the map `calc_minimal_ancestry` gives you, then don't use the function at all. I still feel the loop, both before and as-is (because as-is is just the exact same pattern & algorithm you employed before, just with `calc_minimal_ancestry` shoe-horned in) is _hideously_ ugly, and I would **really** want it improved, but if you can't find a way then so be it.

mandolaerik: Sorry, I don't see how your suggestion would improve the situation.
```
is_override = not all_ranks.isdisjoint(rank.inferior
```
vs
```
is_override = bool(calc_minimal_ancestry(all_ranks)[rank])
```
and
```
                minimal_ancestry = traits.calc_minimal_ancestry(all_ranks)
                [parent, *_] = (parent for (parent_rank, parent) in params
                                if parent_rank in minimal_ancestry[rank])
```
vs
```
                rank_to_decl_or_def = dict(itertools.chain(defs, decls))
                all_overrides = traits.calc_minimal_ancestry(all_ranks)
                [one_override, *] = all_overrides
                parent = rank_to_decl_or_def[one_override]
```
I don't understand why the latter would be better code.

lwaern-intel: Because
```
rank_to_decl_or_def = dict(itertools.chain(defs, decls))
all_overrides = traits.calc_minimal_ancestry(all_ranks)
```
Doesn't have to be part of the loop? You don't have to, you know, _recalculate_ `calc_minimal_ancestry` for every param def/decl. And I find avoiding the need to clumsily spell out `(parent for (parent_rank, parent) in params if parent_rank in minimal_ancestry[rank])` a benefit.

lwaern-intel:
```
all_overrides = traits.calc_minimal_ancestry(all_ranks)
[one_override, *] = all_overrides
```
isn't even legal. Are you simply misunderstanding what `calc_minimal_ancestry` returns?

lwaern-intel: In particular, your example of `is_override = bool(calc_minimal_ancestry(all_ranks)[rank])` is disingenuous because you can just have the result of `calc_minimal_ancestry(all_ranks)` be calculated before the loop and shared between each iteration of it, reducing the check to `is_override = bool(minimal_ancestry[rank])`.

RELEASENOTES-1.4.docu
```diff
@@ -558,4 +558,17 @@
       in multiple inheritance: it allows any method that overrides multiple
       implementations from unrelated templates to individually reference and
       call the implementations it overrides.</add-note></build-id>
+  <build-id _6="next" _7="next"><add-note>Added a system
+      for <em>provisional language features</em>: language features
+      that are disabled by default but can be explicitly enabled with
+      a special syntax. The first provisional feature
+      is <tt>explicit_param_decls</tt>, which can be enabled per file
```
lwaern-intel, line 565: You should split these up to separate releasenotes. One for the provisional system, one for the `explicit_param_decls`. By the way, don't provisionals have a JIRA of their own, or did we never create one?

lwaern-intel (changes requested):

py/dml/dmlparse.py
mandolaerik: Chose `None` instead

py/dml/messages.py
mandolaerik: Improved. I chose to include the provisional declaration on a separate context line.

py/dml/messages.py
mandolaerik: Regarding your original comment: I chose instead to not mention the provisional feature. If the walrus provisional is enabled in a file and you write `:=` in that file, then you are aware of walruses, so that information is redundant. It would make more sense to add a walrus hint to the `ESYNTAX` for `:=`, we can do that when the provisional goes stable.

py/dml/provisional.py
mandolaerik: changed!

py/dml/logging.py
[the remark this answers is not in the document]
mandolaerik: discussed in RC; this is similar to `bitorder` and `dml_version`, where a ICE in SimpleSite would not have gained us.

py/dml/structure.py
```diff
@@ -603,22 +631,47 @@ def is_default(decl):
...
+            is_override = (not decl_or_def_ranks.isdisjoint(rank.inferior)
+                           or rank in decl_ranks)
+            if is_walrus and is_override:
+                [parent, *_] = [
+                    other for (other_rank, other) in decls + defs
+                    if other_rank in rank.inferior] + [
+                    other for (other_rank, other) in decls
+                    if other_rank is rank]
+                report(EWALRUS(type_info.site, type_info.site,#parent.site,
+                               'default' if is_default else '='))
```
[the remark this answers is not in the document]
mandolaerik: fixed

py/dml/provisional.py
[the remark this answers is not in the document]
mandolaerik: Wrote something to clarify this.

py/dml/structure.py
mandolaerik: Things are somewhat complex, there are performance concerns. I solved it by memoizing calc_minimal_ancestry.

RELEASENOTES.docu
mandolaerik: A release note has been written.

RELEASENOTES-1.4.docu
mandolaerik: I didn't find a jira from a quick search.

I have split the releasenote.

syssimics:
Verification [#13429364](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13429364): ✅ pass

py/dml/structure.py
```diff
@@ -542,88 +557,130 @@ def add_templates(obj_specs, each_stmts):
...
+            if not declared_as_override and is_override:
+                [parent, *_] = (parent for (parent_rank, parent) in params
+                                if parent_rank in minimal_ancestry[rank])
+                report(EOVERRIDE(type_info.site, parent.site,
+                                 'default' if is_default else '='))
+            if not declared_as_override and rank in decls:
+                report(EOVERRIDE(type_info.site, decls[rank].site,
+                                 'default' if is_default else '='))
+            elif (not is_override and declared_as_override
+                  and rank not in decls):
```
lwaern-intel, line 666: I slightly prefer:
```
parents = minimal_ancestry[rank]
if not declared_as_override and parents:
                [parent, *_] = (parent for (parent_rank, parent) in params
                                if parent_rank in parents)
...
```
But I don't insist.

RELEASENOTES-1.4.docu
lwaern-intel: If you have, you haven't pushed that up!

lwaern-intel (changes requested):
Approved modulo a small thing that doesn't block merge, and the fact that you seem to have a commit unpushed, which does block merge.

py/dml/messages.py
lwaern-intel: Discussed outside R.C.: he does not, and there are some good arguments for not liking it. How we'll approach device-wide provisionals will have to be resolved later.

mandolaerik:
> If you have, you haven't pushed that up!

Oops, for unknown reason the split was moved to my editor's undo history. Fixed.

RELEASENOTES-1.4.docu
```diff
@@ -559,6 +559,20 @@
...
       call the implementations it overrides.</add-note></build-id>
   <build-id _6="6309" _7="7021"><add-note> Template <tt>function_io_memory</tt>
-      now searches for the bank in the same `subdevice`/`device` scope as the
-      `implement`. </add-note></build-id>
+      now searches for the bank in the same <tt>subdevice</tt>/<tt>device</tt> scope as the
+      <tt>implement</tt>. </add-note></build-id>
+  <build-id _6="next" _7="next"><add-note>Added a system
+      for <em>provisional language features</em>: language features
+      that are disabled by default but can be explicitly enabled with
+      a special syntax. </add-note></build-id>
```
lwaern-intel, line 567:
```
can be explicitly enabled per file via the new <tt>provisional</tt> statement,
which may only occur right after the DML version specification.
```

lwaern-intel (changes requested):

RELEASENOTES-1.4.docu
lwaern-intel: At the very least, that it's per-file is important to note.

syssimics:
Verification [#13431918](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13431918): ✅ pass

syssimics:
Verification [#13432062](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13432062): ✅ pass

RELEASENOTES-1.4.docu
mandolaerik: Fixed.

lwaern-intel (approved):
Approved! You should however squash together some of the 15 commits.

syssimics:
Verification [#13432352](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13432352): ✅ pass

mandolaerik:
I don't think squashing is worth the effort, only ~2 of the commits are squashable.

PR #280, opened 2024-05-06
Implement register_view_catalog

mandolaerik:


mandolaerik:
Depends on https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/6602

This is pretty much a POC, no tests yet

lib/1.4/dml-builtins.dml
```diff
@@ -2520,6 +2520,28 @@ template bank is (object, shown_desc) {
...
+        }
+        method register_offsets() -> (attr_value_t) {
+            local (const register *table, uint64 table_size) = _reginfo_table();
+            local attr_value_t ret = SIM_alloc_attr_list(table_size);
+            for (local int i = 0; i < table_size; i++) {
+                SIM_attr_list_set_item(
+                    &ret, i, SIM_make_attr_uint64(table[i].offset));
+            }
+            return ret;
+        }
```
lwaern-intel, line 2542: Maybe we could construct these `attr_value_t`s in an independent startup memoized, and then have the implementations of these methods be a simple `SIM_attr_copy` on the relevant one fetched?

syssimics:
Verification [#13346870](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13346870): fail

lib/1.4/dml-builtins.dml
lwaern-intel: Once we have RAII strings so we can get an independent `qname()` that _would_ be possible (it is possible even today but not without changes in dmllib.h). Do you think the approach I mentioned is a good direction to take? If so, add a TODO (and possibly a JIRA) for it. If not, then you don't need to bother.

lwaern-intel (approved):

lwaern-intel (changes requested):
Woah, no -- we've changed the name from `register_view_index` to `register_view_catalog`! The commit message and not to mention the commits should reflect that.

syssimics:
Verification [#14006405](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14006405): ❌ fail

mandolaerik:
V errors because this PR depends on https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/6602

lib/1.4/dml-builtins.dml
mandolaerik: I don't see a big value in memoizing:
* current impl is probably not much slower than SIM_attr_copy
* memoizing is likely a net loss, assuming that in most simulations most banks are never register_view:ed.
* We can assume bottlenecks to be elsewhere until proven otherwise. 

What we *could* do is to recycle the `strlen(bank.qname)`. I'll do that.

syssimics:
Verification [#14009305](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14009305): ❌ fail

lwaern-intel (approved):
Optimising the implementation is optional.

syssimics:
Verification [#14009594](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14009594): ❌ fail

mandolaerik:
[This](https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=14009742) is the proper V build

syssimics:
Verification [#14009679](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14009679): ❌ fail

syssimics:
Verification [#14009742](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14009742): ❌ fail

mandolaerik:
[V build](https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=14010234) against https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/7884

syssimics:
Verification [#14010234](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14010234): ❌ fail

PR #284, opened 2024-05-24
Limited scope for function_io_memory

e-malitsky:
Legacy pci bus maps pci_config banks as Device/Subdevice + function number instead of bank objects.
When we have 2+ PCI devices in 1 DML device (e.g. by using subdevice) such pcI_config bank can be unreachable because pci bank function number is hardcoded to 255.
This PR introduces a limited scope for function_io_memory - closest port object parent.

### DIFF
--- a/lib/1.4/utility.dml
+++ b/lib/1.4/utility.dml
@@ -1122,15 +1122,17 @@ template function_io_memory {
         error "The function_io_memory template can only be instantiated"
             + " inside 'implement io_memory'";
     }
-    param port_parent = (this.parent.objtype == "device" 
-                        || this.parent.objtype == "subdevice")
-                        #? this.parent.obj
-                        #: SIM_port_object_parent(this.parent.obj);
+    param subdev_parent = (this.parent.objtype == "device"
+                           || this.parent.objtype == "subdevice")
+                        #? this.parent
+                        // Implement is for a port or bank
+                        #: this.parent._nongroup_parent;
+
     method operation(generic_transaction_t *mem_op,
                      map_info_t map_info) -> (exception_type_t) {
         local uint64 offset = map_info.start - map_info.base;
-        foreach b in (each function_mapped_bank in (dev)) {
-            if (SIM_port_object_parent(b._bank_obj()) == port_parent) {
+        foreach b in (each function_mapped_bank in (subdev_parent)) {
+            if (SIM_port_object_parent(b._bank_obj()) == subdev_parent.obj) {
                 if (b.function == map_info.function) {
                     if (!b.use_io_memory) {
                         local map_target_t *mt = SIM_new_map_target(
--- a/lib/1.4/utility.dml
+++ b/lib/1.4/utility.dml
@@ -1136,14 +1136,12 @@ template function_io_memory {
                 && b.function == map_info.function)
             {
                 if (!b.use_io_memory) {
-                    local map_target_t *mt = SIM_new_map_target(
-                        b._bank_obj(), NULL, NULL);
+                    local map_target_t *mt = SIM_new_map_target(b._bank_obj(),
+                                                                 NULL, NULL);
                     if (!mt) {
-                        local exception_type_t _ex = SIM_clear_exception();
-                        log error: "failed to create map target "
-                            + "for %s: %s",
-                            SIM_object_name(b._bank_obj()),
-                            SIM_last_error();
+                        local exception_type_t _exc = SIM_clear_exception();
+                        log error: "failed to create map target for %s: %s",
+                            SIM_object_name(b._bank_obj()), SIM_last_error();
                         return Sim_PE_IO_Not_Taken;
                     }
                     // hack: VT_map_target_access doesn't accept a map_info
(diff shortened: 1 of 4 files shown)


test/1.4/lib/T_io_memory.dml
```diff
@@ -29,6 +42,30 @@ bank a {
...
+        implement io_memory {
+            is function_io_memory;
+        }
+    }
+    bank function_io_memory_bank {
+        param use_io_memory = false;  // not using bank_io_memory
+        implement io_memory {
+            is function_io_memory;
+        }
+    }
```
lwaern-intel, line 67: To make this test completely exhaustive, this subdevice should feature a subdevice of its own with function-mapped banks, and then the test should confirm that those are not being considered.

lwaern-intel (changes requested):
Thanks for this! This change makes complete sense to me; that `function_io_memory` is not limited to the (sub)device is likely something that @mandolaerik just overlooked when he developed subdevices. Erik, please confirm if that's the case, or if the current state is intentional.

All this PR needs for my approval (beyond Erik's confirmation) is some slight tweaking to the docs, tests, and a releasenote in `RELEASENOTES-1.4.docu`.

test/1.4/lib/T_io_memory.dml
lwaern-intel: Alternatively, instead of a sub-subdevice, add a test case confirming that the main device is not considering the banks of its subdevices. That tests the same logic.

syssimics:
Verification [#13410854](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13410854): pass

lib/1.4/utility.dml
```diff
@@ -1121,40 +1121,48 @@ template function_io_memory {
...
-            if (b.function == map_info.function) {
-                if (!b.use_io_memory) {
-                    local map_target_t *mt = SIM_new_map_target(b._bank_obj(),
-                                                                 NULL, NULL);
-                    if (!mt) {
-                        local exception_type_t _exc = SIM_clear_exception();
-                        log error: "failed to create map target for %s: %s",
-                            SIM_object_name(b._bank_obj()), SIM_last_error();
+            if (SIM_port_object_parent(b._bank_obj()) == port_parent) {
+                if (b.function == map_info.function) {
```
mandolaerik: `&&` would avoid extra indent

lib/1.4/utility.dml
```diff
@@ -1121,40 +1121,48 @@ template function_io_memory {
         error "The function_io_memory template can only be instantiated"
             + " inside 'implement io_memory'";
     }
-
+    param port_parent = (this.parent.objtype == "device"
```
mandolaerik: please prefix param name with `_` since it's not documented

lib/1.4/utility.dml
```diff
@@ -1121,27 +1122,37 @@ template function_io_memory {
...
 
     method operation(generic_transaction_t *mem_op,
                      map_info_t map_info) -> (exception_type_t) {
         local uint64 offset = map_info.start - map_info.base;
-        foreach b in (each function_mapped_bank in (dev)) {
-            if (b.function == map_info.function) {
+        foreach b in (each function_mapped_bank in (_subdev_parent)) {
+            if (SIM_port_object_parent(b._bank_obj()) == _subdev_parent.obj
+                && b.function == map_info.function)
+            {
```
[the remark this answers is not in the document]
lwaern-intel: **wow** that's fast

lwaern-intel (approved):
Approved for my part. I'm curious what kind of model you're working with that combines such a new feature as subdevices together with such a legacy feature as function-mapped banks. How did that happen?

test/1.4/lib/T_io_memory.dml
mandolaerik: Also, consider wrapping one of the banks or ports in a group.

e-malitsky:
> Approved for my part. I'm curious what kind of model you're working with that combines such a new feature as subdevices together with such a legacy feature as function-mapped banks. How did that happen?

Hopefully we will migrate to new pci lib sometime

syssimics:
Verification [#13411170](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13411170): pass

e-malitsky:
> > Legacy pci bus maps pci_config banks as Device/Subdevice + function number instead of bank objects.
> 
> `pci-bus` maps banks on function number only if some device requests to be mapped on a function number through the method `pci_bus.add_map`. This used to be the default behaviour in Simics 5, but since Simics 6 the default is to map the bank object (method add_map in src/devices/dml-lib/1.4/pci/common.dml). We have seen use cases where dml-lib's add_map was overridden with a modified copy within the 5 timeframe.
> 
> If this is the case for you, then I would suggest that you also update your copy of add_map to resemble what the lib does in 6, so that your banks get mapped on object. Function mapped banks make it harder to inspect the system.
> 
> The PR itself is still valid, it does fix a bug

I was under the impression that we can somewhat customize io/mem memory spaces of pci_bus but conf is hardcoded to 255.
Thanks, will take a proper look

mandolaerik (approved):

mandolaerik:
Indeed, that's a problem in pci-bus. Wrote a PR: https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/6767

PR #286, opened 2024-05-27
Type Equality/Compatibility Revamp

lwaern-intel:
Taken from #213, but with some refinements

Gonna run this against VP once verification passes. I'm 50/50 on whether it causes breakage. On one hand, this definitely tightens down the equality logic. On the other hand, GCC should have been rejecting the cases that slips through, unless some warnings get disabled or something.


py/dml/ctree.py
```diff
@@ -4787,30 +4790,41 @@ def mkCast(site, expr, new_type):
...
+    if isinstance(real, (TVoid, TArray, TFunction)):
         raise ECAST(site, expr, new_type)
-    if isinstance(real, (TVoid, TArray, TVector, TTraitList, TFunction)):
+    if old_type.cmp(real) == 0:
+        if (old_type.is_int
+            and not old_type.is_endian
+            and dml.globals.compat_dml12_int(expr.site)):
+            # 1.2 integer expressions often lie about their actual type,
+            # and require a "redundant" cast! Why yes, this IS horrid!
+            return Cast(site, expr, new_type)
```
lwaern-intel, line 4817: This would warrant a `FIXME` if it weren't for that 1.2 integer arithmetic is dying anyway.

py/dml/types.py
```diff
@@ -784,16 +867,26 @@ def sizeof(self):
...
     def cmp(self, other):
-        if compat.dml12_misc in dml.globals.enabled_compat:
-            if isinstance(other, (TArray, TPtr)):
-                return self.base.cmp(other.base)
-        elif isinstance(other, (TPtr, TArray)):
-            if self.base.void or other.base.void:
-                return 0
-            if self.base.cmp(other.base) == 0:
-                return 0
-        return NotImplemented
```
lwaern-intel, line 796: This is a strange one. Previously, `TArray.cmp` 's logic caused it to be more lenient _without_ the `dml12_misc` compatibility feature. I originally thought that maybe you made a mistake while making #209, but no, that was true even before then. Same story with `TPtr.cmp`.
While moving old behaviors of `cmp` to `cmp_fuzzy` I've just dropped these `dml12_misc` check of `TArray` and `TPtr`.

py/dml/types.py
```diff
@@ -168,6 +171,27 @@ def conv_const(const, t):
...
+                t.base = base
+        elif t.const:
+            t = t.clone()
+            t.const = False
+        return t
+    return sub(safe_realtype(t0))
+
+def shallow_const(t):
+    t = safe_realtype_shallow(t)
+    while not t.const and isinstance(t, (TArray, TVector)):
```
lwaern-intel, line 190: The inclusion of `TVector` here is erroneous. Probably. I guess you could argue about it but really, it doesn't matter. It **will** matter for RAII vectors, but not for legacy vectors.

lwaern-intel:
Oh, we even have an issue about this: SIMICS-9504. I wouldn't say this closes completely closes (just contributes to fixing it) , due to this still hanging onto `cmp_fuzzy` as a HACK for some of the cases that SIMICS-9504 touches upon. The primary point of this PR is to introduce a mechanism for strict type equality (which we can leverage for e.g. idempotent externs), and secondarily to convert most aspects of DMLC to leverage that mechanism. The remaining parts that still uses `cmp_fuzzy` constitute the most difficult cases, where some kind of type unification or conversion are needed (such as the ternary operator and comparison operators.)

syssimics:
Verification [#13419451](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13419451): ❌ fail

py/dml/ctree.py
```diff
@@ -4787,30 +4790,41 @@ def mkCast(site, expr, new_type):
...
-        if isinstance(old_type, TStruct) and old_type.label == real.label:
-            return expr
-        raise ECAST(site, expr, new_type)
-    if isinstance(real, TExternStruct):
-        if isinstance(old_type, TExternStruct) and old_type.id == real.id:
-            return expr
+    if isinstance(real, (TVoid, TArray, TFunction)):
         raise ECAST(site, expr, new_type)
-    if isinstance(real, (TVoid, TArray, TVector, TTraitList, TFunction)):
+    if old_type.cmp(real) == 0:
```
lwaern-intel, line 4811: I just realized this path is disgustingly not always correct due to bitfields. A 64-bit bitfields is completely type compatible with `uint64`, even in the sense as laid out by the docstring of `DMLType.cmp`, and yet DMLC will treat them differently w.r.t. `mkSubRef`, so just spitting out the same expression `mkRValue`-d is not correct. I suppose we could have an ad-hoc exception for those cases.

lwaern-intel: Alternatively, instead of doing `mkRValue`, add an arg for `Cast` that tells it not to spit out a cast in generated C. Or an alternate class to `Cast` that does the same thing. Or use `Cast` anyway.

lwaern-intel: This also really highlights the difference between type _equality_ and type _compatibility_. We've gotten away with treating them as one and the same, but in this instance it falls apart. Does that mean we should introduce a separate notion for type _equality..._?

lwaern-intel: I think we'll be forced to, if we're not willing to treat bitfields as type-compatible with the corresponding `uintX` type. Consider idempotent externs; an `extern` with a bitfields arg can use `{}` initialized syntax, but an `extern` with a `uintX` arg cannot. We cannot consider two declarations differing in that aspect equal to one another, even though their signatures are completely compatible. There's a similar issues with method declarations and overrides.

lwaern-intel: The notion of type equality would thus be type compatibility with an additional restriction: that DMLC treats the type identically in all ways that are visible to the modeler.

py/dml/types.py
lwaern-intel: Now no longer delving into `TVector`.

py/dml/ctree.py
lwaern-intel: Added an ad-hoc exception for the moment.

syssimics:
Verification [#13420028](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13420028): ❌ fail

py/dml/types.py
```diff
@@ -1311,11 +1457,22 @@ def type_union(type1, type2):
 def compatible_types(type1, type2):
     # This function intends to verify that two DML types are
     # compatible in the sense defined by the C spec, possibly with
-    # some DML-specific restrictions added. TODO: DMLType.cmp is only
-    # a rough approximation of this; we should write tests and
-    # either repair cmp or rewrite the logic from scratch.
+    # some DML-specific restrictions added.
     return type1.cmp(type2) == 0
 
+# TODO(RAII) We should look into getting rid of this and cmp_fuzzy, and replace
```
lwaern-intel, line 1463: Oops, stray `(RAII)`.

lwaern-intel:
making this draft as I'm discovering there's needed stuff in the RAII pr I forgot to break out. Will make this non-draft when verification passes in the [compatibility PR](https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/6778)

lwaern-intel:
This does break VP violently, unfortunately. A compatibility feature has been added.

py/dml/c_backend.py
```diff
@@ -807,13 +807,13 @@ def generate_implement_method(device, ifacestruct, meth, indices):
...
-            if safe_realtype(mt).cmp(safe_realtype(it)) != 0:
+            if safe_realtype_unconst(mt).cmp(safe_realtype_unconst(it)) != 0:
                 raise EARGT(meth.site, 'implement', meth.name,
                             mt, mp, it, 'method')
         if iface_num_outputs and dml.globals.dml_version != (1, 2):
             [(_, mt)] = meth.outp
-            if safe_realtype(mt).cmp(
-                    safe_realtype(func_type.output_type)) != 0:
+            if safe_realtype_unconst(mt).cmp(
+                    safe_realtype_unconst(func_type.output_type)) != 0:
```
lwaern-intel, line 816: Did not gate these behind the compatibility feature. Did not seen any new error message from them. Will confirm with a rerun, though.

syssimics:
Verification [#13432053](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13432053): ✅ pass

py/dml/codegen.py
```diff
@@ -1306,18 +1306,7 @@ def expr_cast(tree, location, scope):
...
-        return mkLit(tree.site, "%s->%s" % (
-            expr.node_expr.read(), expr.method_name), type)
-
-    if isinstance(expr, NonValue) and (
-            not isinstance(expr, NodeRef)
-            or not isinstance(safe_realtype(type), TTrait)):
-        raise expr.exc()
-    else:
-        return mkCast(tree.site, expr, type)
+    return mkCast(tree.site, expr, type)
```
mandolaerik, line 1309: There is a convention that you should check `NonValue` explicitly after `codegen_expression_maybe_nonvalue`, so it's possibly better to keep the `if` where it was, even if it's functionally redundant.

mandolaerik: Alternatively, we could break out the non-value cases of mkCast to separate functions.

py/dml/ctree.py
mandolaerik: I don't think compatibility between `uintN` and `bitfields N` is important -- one can surely come up with use cases, but I can't see how they would be common (e.g., did you ever get a uint64 pointer that you want to pass to a function that takes a bitfields pointer?). If making them incompatible removes the need for a "type equality" concept, then it's most likely worth it. Especially since we now have a compat feature to hijack.

mandolaerik: If it's easy to introduce uintN vs bitfeldsN incompatibility separately, without the full type revamp, then it would be nice to run that with jenkins, to grasp how much this part would break.

py/dml/ctree.py
```diff
@@ -1396,7 +1396,7 @@ def make(cls, site, lh, rh):
         if ((lhtype.is_arith and rhtype.is_arith)
             or (isinstance(lhtype, (TPtr, TArray))
                 and isinstance(rhtype, (TPtr, TArray))
-                and compatible_types(lhtype.base, rhtype.base))):
+                and compatible_types_fuzzy(lhtype.base, rhtype.base))):
```
mandolaerik, line 1399: This looks somewhat scary: `Compare` is a widely 1.4 thing, and still uses the ill-defined fuzzy cmp. If we want a single compat feature for lenient type checking, then we should consider bundling fuzzy cmp with it?

If we don't have the time to port all forgiveness corners in the short term, then one option is to start strict, i.e., let everyone say:
```
def compatible_types(lh, rh):
  if lenient in enabled_compat:
      return compatible_types_fuzzy(lh, rh)
  ...
```
If we later see that this is too strict, then it's safe to relax relevant corner later.

py/dml/types.py
```diff
@@ -582,7 +655,13 @@ def declaration(self, var):
         else:
             return 'uint8 ' + self.const_str + var + '[' + str(self.bytes) + ']'
 
-class TLong(IntegerType):
+class ArchDependentIntegerType(IntegerType):
+    '''Integer types whose definition and/or properties are architecture
```
mandolaerik, line 659: Singular, `Integer type whose ... is ...`, because an instance represents one such type.

py/dml/types.py
```diff
@@ -477,20 +529,35 @@ def get_member_qualified(self, member):
         return t
 
     def cmp(self, other):
+        if self.const != other.const:
+            return NotImplemented
         if not other.is_int:
             return NotImplemented
+        if ((self.is_arch_dependent or other.is_arch_dependent)
+            and type(self) is not type(other)):
+            return NotImplemented
```
mandolaerik, line 538: Cleaner to express as `ArchDependentIntegerType.cmp`?

lwaern-intel: Not quite good enough, consider `TInt.cmp(TLong)`. That said, maybe it's not unwarranted to just change the mechanism of how `IntegerType.cmp` works and make it per-subclass based, and get rid of the reliance on `is_int` and `is_endian`.

py/dml/ctree.py
lwaern-intel: > This looks somewhat scary: Compare is a widely 1.4 thing, and still uses the ill-defined fuzzy cmp.

Like I've said earlier:

> The primary point of this PR is to introduce a mechanism for strict type equality (which we can leverage for e.g. idempotent externs), and secondarily to convert most aspects of DMLC to leverage that mechanism. The remaining parts that still uses `cmp_fuzzy` constitute the most difficult cases, where some kind of type unification or conversion are needed (such as the ternary operator and comparison operators.)

I even note in the documentation of `fuzzy_cmp` and `compatible_types_fuzzy` that all uses of them should be considered HACKs. In particular, we can't use `compatible_types` here for `Compare` as that would mean you couldn't compare a void pointer with any other pointer.

I rather not introduce even more mechanisms _as part of this PR_ to cover compares and the ternary operator unless you insist upon it.

lwaern-intel: ok maybe actually the `Compare` case is moot? Maybe the following is good enough:
```
    ...
    and (lhtype.base.void or rhtype.base.void
            or compatible_types(safe_realtype_unconst(lhtype.base),
                                safe_realtype_unconst(rhtype.base)))):
```

I think I'll need to look at the C standard to be sure.

lwaern-intel: C standard says for `==` and `!=`:
> One of the following shall hold:
> * both operands have arithmetic type;
> * both operands are pointers to qualified or unqualified versions of compatible types;
> * one operand is a pointer to an object or incomplete type and the other is a pointer to a
qualified or unqualified version of `void`; or
> * one operand is a pointer and the other is a null pointer constant.
Case 3 and 4 are separated to forbid comparing equality between a function pointer and a void pointer. I suggest we don't bother with that.

C standard says for `<`, `>`, `<=`, `>=`:
> One of the following shall hold:
> * both operands have real type;
> * both operands are pointers to qualified or unqualified versions of compatible object types; or
> * both operands are pointers to qualified or unqualified versions of compatible
incomplete types

This excludes function pointers entirely, but also, to my surprise, even excludes void pointers. I don't know why, honestly. Well if C forbids it we have no reason to permit it.

Regardless, these rules are simple enough to not need `compatible_types_fuzzy`. Now lemme check the ternary case...

mandolaerik: > I rather not introduce even more mechanisms as part of this PR to cover compares and the ternary operator unless you insist upon it.

Fixing up remaining fuzzy cmp will introduce a different kind of type checking lenience. I can see some options:
* Make this compat feature unstable/provisional-ish, so we can stricten it in a later PR (requires that we add some base support for unstable compats)
* Plan to add unfuzzing as a separate compat feature later. Preparation should include renaming this feature to something more specific, to make it easy to distinguish from the unfuzzing compat feature.
* Blindly unfuzz all typechecks when lenient is disabled, and loosen the rules in a later PR (i.e. what I suggested earlier).

lwaern-intel: Bleh.

> One of the following shall hold for the second and third operands:
> * both operands have arithmetic type;
> * both operands have the same structure or union type;
> * both operands have `void` type;
> * both operands are pointers to qualified or unqualified versions of compatible types;
> * one operand is a pointer and the other is a null pointer constant; or
> * one operand is a pointer to an object or incomplete type and the other is a pointer to a
qualified or unqualified version of `void`

I'm not even sure what to make of this. In particular, I don't know if there's any difference between point 2 (when taking into account point 1) and "both operands are qualified or unqualified versions of compatible object types"

mandolaerik: I don't think ternary `?:` needs to be that hard: We don't need the full complexity of "common type" in C; it should suffice if numeric are handled as by other binary operands, and otherwise require operands to be of compatible types (possibly modulo shallow_unconst).

lwaern-intel: > it should suffice if numeric are handled as by other binary operands, and otherwise require operands to be of compatible types (possibly modulo shallow_unconst).
Nope, we need void-pointer compatibility. Though that's not a particularly hard exception to add either.

lwaern-intel: Really the hardest part would be, as you say, common type-ing. In particular, const-unification.

lwaern-intel: wait we already have code for that, it just has to be improved a slight bit. Yeah. Ok. This actually does seem plausible! We can get rid of the non-gated `compatible_type_fuzzy`s!

py/dml/types.py
mandolaerik: Another option is to embrace that it is a cross-cutting concern, and make it a global function with `match` instead.

py/dml/ctree.py
lwaern-intel: hah my ad-hoc exception wasn't good enough to clear jenkins:
```
        vid1 = (cast(&vid1, vlan_tag_t *))->vid;
```
`&vid` is `uint16 *` which is considered compatible with `vlan_tag_t *` and thus `mkRValue` is used, oops.

Anyhoo, I'll try out adding the rule that a bitfields is only considered compatible to another bitfields if they share the same exact members (maybe with some memoization...?) in addition to the other properties like signedness and endianness.

py/dml/ctree.py
mandolaerik: What I mean is that we can start with a restrictive baseline: completely forbid `foo_t *` : `const foo_t *` to start with, see what breaks, and conservatively soften the ban to polish the roughest corners.

We'll immediately notice that `char *:const char *` is needed, for `cond?variable:"literal"`, and for that it should suffice to say that if we have two pointers, then compatibility of target types (modulo const) is sufficient for acceptance, where const in either operand takes precedence. This may forbid some corners that should be accepted, but if that's rare then it's procrastinatable.

syssimics:
Verification [#13435945](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13435945): ✅ pass

py/dml/ctree.py
```diff
@@ -4787,30 +4798,46 @@ def mkCast(site, expr, new_type):
...
         raise ECAST(site, expr, new_type)
-    if isinstance(real, TExternStruct):
-        if isinstance(old_type, TExternStruct) and old_type.id == real.id:
-            return expr
-        raise ECAST(site, expr, new_type)
-    if isinstance(real, (TVoid, TArray, TVector, TTraitList, TFunction)):
+    if old_type.cmp(real) == 0:
+        if (old_type.is_int
+            and not old_type.is_endian
+            and dml.globals.compat_dml12_int(expr.site)):
```
mandolaerik, line 4822: I prefer compat check first in the conjunction when possible: allows you to quicker skip to the interesting stuff

py/dml/types.py
```diff
@@ -231,25 +256,47 @@ def sizeof(self):
...
         2. in ctree, to compare how large values different numerical
         types can hold
-        3. when judging whether a method override is allowed, as an inaccurate
-        replacement of TPtr(self).canstore(TPtr(other))[0]
+        3. when judging whether a method override is allowed
 
         See SIMICS-9504 for further discussions.
+        """
+        return (0 if type(self) is type(other) and self.const == other.const
+                else NotImplemented)
```
mandolaerik, line 285: I would suggest switching signature while we are at it, e.g. `def compatible(self, other) -> bool:`
Not strictly needed within this PR, though.

py/dml/types.py
```diff
@@ -231,25 +256,47 @@ def sizeof(self):
...
+        The method is used for three purposes:
         1. in TPtr.canstore(), to judge whether pointer target types
         are compatible.
         2. in ctree, to compare how large values different numerical
         types can hold
-        3. when judging whether a method override is allowed, as an inaccurate
-        replacement of TPtr(self).canstore(TPtr(other))[0]
+        3. when judging whether a method override is allowed
 
         See SIMICS-9504 for further discussions.
```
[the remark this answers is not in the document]
mandolaerik: Oh, and I think this PR fixes SIMICS-9504, so feel free to sprinkle jira references in commit messages and release notes.

py/dml/types.py
```diff
@@ -231,25 +256,47 @@ def sizeof(self):
...
+        not satisfy (2).
+
+        The method is used for three purposes:
         1. in TPtr.canstore(), to judge whether pointer target types
         are compatible.
         2. in ctree, to compare how large values different numerical
         types can hold
-        3. when judging whether a method override is allowed, as an inaccurate
-        replacement of TPtr(self).canstore(TPtr(other))[0]
+        3. when judging whether a method override is allowed
```
mandolaerik, line 280: I guess your patch makes these remarks pretty much redundant (they were there to help understanding how cmp differs from canstore)

test/1.4/legacy/lenient_typechecking.dml
```diff
@@ -0,0 +1,42 @@
+/*
+  © 2024 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+
+dml 1.4;
```
mandolaerik, line 6: Other legacy test have a boilerplate reminder about the SCAN-FOR-TAGS trick.

test/1.4/legacy/lenient_typechecking.dml
```diff
@@ -0,0 +1,42 @@
...
+    method m(int *p) -> (int *) { return p; }
+
+    /// ERROR EMETH
+    method n(int *p) -> (int *) { return p; }
+}
+
+header %{
+    #define MACRO(a,b,c) ((void)0)
+%}
+
```
mandolaerik, line 30: We probably want more test coverage of `cmp`, e.g. whether `uint16` is compatible with `bitfields 16`. Perhaps something for `types_test.py`?

mandolaerik: Might also be nice with a ctree_test-style test to get reasonable coverage of the invariant that DML type compatibility implies C type compatibility.

py/dml/compat.py
```diff
@@ -178,6 +178,23 @@ class suppress_WLOGMIXUP(CompatFeature):
...
+    generated C.
+    This discrepency mostly affects method overrides or uses of `extern`:d C
+    macros, because in those scenarios DMLC can become wholly responsible for
+    proper type checking.
+
+    While migrating from this feature, novel type errors due to uses of
+    `extern`:d macros can often be resolved by changing the signature of
+    the `extern` declaration to more accurately reflect the macro's effective
+    type.
+    '''
```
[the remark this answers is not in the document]
lwaern-intel: Errgghhh... there are so many different cases that I struggle to think of a **best** example. In fact, the things this affects are so varied that highlighting a common case in particular may be misleading, as it obscures the fact that there are other, _distinct_, cases that affects people's (SSM's) code just as much.
I highlight `extern`:d macros and method overrides because those are the things DMLC are entirely responsible for, but for SSM who just loooove to ignore GCC warnings there are far more and more varied beasts out there.

When it comes to a specific thing that causes most breakage, the most significant is disagreement in const-qualification. But the _ways_ that this causes breakage is varied. For one, there are the fact that `cmp` used to not take const-qualification into account at all; this is what affects method overrides. However there's also the fact that `TPtr.canstore` used to not care about discarding const qualifiers once `void` was involved. This is the big thing (but not the ONLY) thing that is affecting `extern`:d macros (due to the `extern` signature not properly const-qualifying); but once you ignore GCC warnings, the lax logic re. discarding const has more impact than just improper `extern` declarations. It allows three distinct cases, each of which occur in SSM code:
```
const blabla_t *bla = ...;
...
// discarding const qualifier and ignoring GCC warning. 
// DMLC doesn't error as the pointer base type of the parameter is `void`
MM_FREE(bla); 
```

```
// discarding const qualifier and ignoring GCC warning. 
// DMLC doesn't error as the pointer base type of the argument is (`const`) `void`
local some_iface_t *iface = SIM_get_interface(dev, "blablabla");
```

And probably the most promiment:
```
bytes_t whatever;
// discarding const qualifier and ignoring GCC warning. 
// DMLC doesn't error as the pointer base type of the parameter (and also the argument) is (`const`) `void`
memcpy(whatever.data, ..., ...);
```

Disabling `lenient_typechecking` will upgrade each of these cases to a proper error.

in short: _what example should we promote, that actually gives an accurate idea of what people can expect_?

py/dml/types.py
lwaern-intel: aa

I think I'd hate that.

So: what are we doing for now? Status quo, or do you want me to be rid of the current architecture of `cmp` and `canstore` surrounding `IntegerType` and replace it with an entirely subclass-based one?

py/dml/types.py
lwaern-intel: How so, given no usage remains of `cmp_fuzzy` not guarded by `lenient_typechecking`?

py/dml/types.py
lwaern-intel: What about naming? Now we're preferring equality, should the name be `Type.eq` and `equal_types` instead of `compatible_types`?

py/dml/ctree.py
lwaern-intel: Now gotten rid of all ungated usages of `cmp_fuzzy`.

py/dml/codegen.py
lwaern-intel: Discussed verbally

py/dml/types_test.py
```diff
@@ -0,0 +1,282 @@
...
+                    TDevice("a")):
+            typ_clone = typ.clone()
+            self.assertTrue(
+                realtype(typ_clone).eq(realtype(typ)))
+            self.assertTrue(
+                realtype(typ).eq(realtype(typ_clone)))
+            typ_clone.const = True
+            self.assertFalse(typ.const)
+            typ = typ_clone.clone()
+            self.assertTrue(typ.const)
```
lwaern-intel, line 38: Note this addition to the clone test. This catches the previously bad `TVoid` and `TBool` implementations.

py/dml/ctree.py
lwaern-intel: i forgot to do this
I'll push it up with whatever other stuff i'll need to address in your next round of feedback

py/dml/types.py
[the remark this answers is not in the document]
lwaern-intel: Forgot having a reference in the release notes/commit message

py/dml/types.py
lwaern-intel: these have been removed

test/1.4/legacy/lenient_typechecking.dml
lwaern-intel: _done_

py/dml/compat.py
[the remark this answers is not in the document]
lwaern-intel: Discussed verbally; no example given, instead clarifying the most common kind of type errors you'd observed.

py/dml/types.py
lwaern-intel: ... _though i forgot `is_arch_dependent`, oops_

py/dml/types_test.py
```diff
@@ -0,0 +1,282 @@
...
+        t = bitfields()
+
+        def neq_bitfields_testcase():
+            nonlocal t
+            self.assert_neq(bitfields(), t)
+            t = bitfields()
+
+        # Presence of members matter
+        del t.members['a']
+        neq_bitfields_testcase()
```
mandolaerik, line 116: macro-like function, probably clearer to just expand it every time. AFAICS, the cost in code size is zero

py/dml/types_test.py
```diff
@@ -0,0 +1,282 @@
...
+            a = self.mkTArray(4)
+            b = self.mkTArray(4)
+
+        # Base type matters
+        a.base.signed = True
+        array_testcase(False)
+
+        # Constness: constness of base type and array are normalized
+        a.const = True
+        array_testcase(False)
```
mandolaerik, line 180: Same here, what about:
```
def two_array_types():
  return (self.mkTArray(4), self.mkTArray(4))
...
self.assert_eq(a, b)
(a, b) = two_array_types()
```

syssimics:
Verification [#13481859](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13481859): ✅ pass

py/dml/types_test.py
lwaern-intel: The reason I did this here is because I wanted to reduce the boilerplate and noise needed for each test-case specification. The fact this pattern incurs noise _before_ the mass of specifications, or the code size would be identical were the function to be expanded are of less relevance to me compared to the fact that each test case would always need two lines of boilerplate in the thick of it instead of one. I feel that hampers readability.

py/dml/types_test.py
lwaern-intel: dun wanna

(for a more persuasive answer, see above)

py/dml/types_test.py
mandolaerik: My problem with the function is that it is not an abstraction: it does capture a recurring sequence of operations, but the best description I can come up with is "do the last thing we usually do in the end of one testcase, and then do the first thing we usually do in the beginning of the next". Every time you read a call to the function, you need to do this expansion in your head to understand what will happen. This makes the reading heavier than just reading those two lines of code straight up.

py/dml/types.py
```diff
@@ -279,7 +328,23 @@ def print_declaration(self, var, init = None, unused = False):
...
     def key(self):
+        '''A string key uniquely identifying the DML type, and is suitable for
+        use in checkpointing. Will raise DMLUnkeyableType if such a key can't
+        be produced.
+
+        If neither a, b is unkeyable, then:
+        * a.eq(b) must imply a.key() == b.key()
+        * a.key() == b.key() implies a.eq(b) with overwhelming likelihood
+          (though not necessarily guaranteed; see IntegerType.key())
+        '''
```
lwaern-intel, line 347: The changes to `key()` in this PR technically causes backwards checkpoint incompatibility... were it not for the fact that it's `key()` currently only used for `hook.send_now()` callbacks of afters. Making it completely irrelevant in practice.

syssimics:
Verification [#13483956](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13483956): ✅ pass

py/dml/types_test.py
lwaern-intel: In that case, how about:

```

        @contextmanager
        def neq_bitfields_testcase():
            t = bitfields()
            yield t
            self.assert_neq(bitfields(), t)
```

This _is_ actually an abstraction, and would actually resemble the way I'd do it in a **better** language like Haskell (I'd do a higher-order function, and then leverage lambdas that have the luxury of actually being able to stretch multiple lines.)

lwaern-intel: We've agreed on the context manager solution

py/dml/types_test.py
lwaern-intel: context manager used instead

mandolaerik (approved):

syssimics:
Verification [#13485192](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13485192): ✅ pass

syssimics:
Verification [#13485399](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13485399): ✅ pass

lwaern-intel:
[Final for-safety's-sake Jenkins run](https://cje-fm-simics-prod01.devtools.intel.com/satg-simics-main/job/sre-vp/job/_pretest/137597/)

mandolaerik (approved):

PR #288, opened 2024-05-28
Migrate away from `SIM_register_typed_attribute` -- SIMICS-22406

lwaern-intel:
Draft as we need to deal with the compatibility question of the drop of support of dictionary (`D`) types.


lwaern-intel (commented):
Doh, I forgot to make sure getter/setter trampolines work.

syssimics:
Verification [#13471519](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13471519): ✅ pass

syssimics:
Verification [#13471492](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13471492): ✅ pass

lwaern-intel:
@mandolaerik nag

test/1.4/legacy/legacy_attributes.dml
```diff
@@ -0,0 +1,43 @@
...
+attribute a is uint64_attr;
+
+bank b {
+    saved int s;
+    hook() h;
+    attribute a is uint64_attr;
+}
+
+method init() default {
+    local attr_value_t val = SIM_make_attr_uint64(0xC0FFEE);
```
mandolaerik, line 21: Note: we usually use lowercase hex. Unsure if the style mandates it, and for bogus test values I don't care.

mandolaerik (approved):

test/1.4/legacy/legacy_attributes.dml
lwaern-intel: huh? we do? if so, we are **definitely** not consistent about it like, at all.

PR #291, opened 2024-05-31
Improve TQMIC error messages

lwaern-intel:
Two major differences:
* Uses the new `EMEMBERTQMIC` instead of the generic `EMEMBER`. This has specialized wording and also takes care to point out if the template only provides/inherits an abstract declaration of named method (requested by @gwikland)
* TQMIC error messages now detect and note if they could have been caused by specifications getting `#if`-d away.


py/dml/messages.py
```diff
@@ -1707,6 +1712,31 @@ def log(self):
...
                 + f"'{ancestor.name}'")
 
+class EMEMBERTQMIC(DMLError):
+    """A template-qualified method implementation call can only be done if
+    the specified template actually does provide or inherit an implementation
+    of the named method for the object instantiating the template. That the
+    template provides or inherits an abstract declaration of the method is not
+    sufficient.
+
+    Besides more mundane causes (e.g. misspellings), this error could happen if
```
mandolaerik, line 1722: I think `Beside`, or `Apart from`/`Except for`, (cf "Besides, ..." which would give a parse error)

lwaern-intel: Going with "Apart from", though I'm 90% sure that "Besides" (and not "Beside") is correct in this context.

py/dml/messages.py
```diff
@@ -1707,6 +1712,31 @@ def log(self):
...
+    not being provided to the object instantiating the template, due to every
+    implementation being eliminated by an `#if` statement."""
+    fmt = ("invalid template-qualified method implementation call, '%s' does "
+           + "not provide nor inherit an implementation of a method "
+           + "'%s'%s")
+    def __init__(self, site, template, method, abstract, node):
+        if node:
+            extra = (f" for the object '{node.identity()}': all "
+                     + "implementations that could have been given have been "
+                     + "eliminated by #if statements")
```
mandolaerik, line 1733: Did you consider saying this in a separate print_site_message on one of the eliminated `#if`s? Might read better.

Same with `abstract`, but less important since we know for sure it's within the same template.

And with neither abstract nor node, would be useful with a print_site_message to simply point you to the template decl that doesn't contain the method.

mandolaerik: > as in... the qualified template of the TQMIC? What's the benefit of doing that?

You did apparently misunderstand something about that template, so you'll need to study it in one way or another to find out what you missed. Sometimes you'll *also* need to study a supertemplate, but at least we can help out with a starting point.

syssimics:
Verification [#13451492](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13451492): ❌ fail

lwaern-intel:
Verification failure's unrelated

lwaern-intel:
@mandolaerik awaiting your final approval

syssimics:
Verification [#13470160](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13470160): ❌ fail

mandolaerik (approved):

PR #292, opened 2024-06-03
Document requirment on class registration ordering

mandolaerik:
Discovered by Andreas Hedström during cmake migration.


lib/1.4/dml-builtins.dml
```diff
@@ -1391,6 +1391,20 @@ a private helper object.
...
 `classname`. This defines the class of the automatically created object.
 
+<div class="note">
+
+**Note:** The subobject class defined by the `classname` parameter is looked up
+using `SIM_get_class` while the module of the device class is being
+loaded.
+This may cause problems if the subobject class is defined within the
+same module as the device class:
+If the device class is defined before the subobject class, then the subobject
```
lwaern-intel, line 1401: nit: `if`, not `If`.

lwaern-intel (approved):

andreashedstrom-intel (approved):

syssimics:
Verification [#13455451](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13455451): ✅ pass

lib/1.4/dml-builtins.dml
mandolaerik: My feeling was that both are valid and I didn't know which one is better. ChatGPT chose your side so I changed

syssimics:
Verification [#13455981](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13455981): ❌ fail

PR #298, opened 2024-06-17
Check indices passed to method functions -- SIMICS-22491

lwaern-intel:
@TSonono This is intended to solve your problem. Try to check-out this commit in the `src/tools/dmlc` submodule and build DMLC using it, then see if it makes any difference when you try to build your device (I haven't managed to reproduce your issue, which is no surprise given how fickle GCC is about this)

### DIFF
--- a/include/simics/dmllib.h
+++ b/include/simics/dmllib.h
@@ -41,11 +41,53 @@ static inline uint64 DML_combine_bits(uint64 x, uint64 y, uint64 mask)
 } while (0)
 
 
-#define DML_ASSERT_FMT(filename, lineno, expr, ...) do {                      \
-    if (unlikely(!(expr))) {                                                  \
-        strbuf_t sb = sb_newf(__VA_ARGS__);                                   \
-        assert_error((lineno), (filename), "", sb_str(&sb));                  \
-} while (0)
+UNUSED static char *
+_DML_format_indices(const char *fmt, const uint32 *indices, uint32 len) {
+    strbuf_t buf = SB_INIT;
+    for (uint32 i = 0; i < len; ++i) {
+        const char *next_index = strchr(fmt, '%');
+        ASSERT(next_index && next_index[1] == 'u');
+        sb_addfmt(&buf, "%.*s%u", (int)(next_index - fmt), fmt, indices[i]);
+        fmt = next_index + 2;
+    }
+    ASSERT(!strchr(fmt, '%'));
+    sb_addstr(&buf, fmt);
+    return sb_detach(&buf);
+}
+
+UNUSED NORETURN static void
+_DML_assert_indices_fail(const char *restrict logname,
+                         const char *restrict method_name,
+                         const uint32 *restrict indices,
+                         const uint32 *restrict dimsizes,
+                         int len) {
+    char *qname_indices = _DML_format_indices(logname, indices, len);
+    char *qname_dims = _DML_format_indices(logname, dimsizes, len);
+    FATAL_ERROR("Attempt made to call %s.%s(), which involves one or more "
+                "out-of-bounds array accesses. Array dimensions are as "
+                "follows: %s", qname_indices, method_name, qname_dims);
+}
+
+UNUSED static inline void
+_DML_assert_indices(const char *restrict logname,
+                    const char *restrict method_name,
+                    const uint32 *restrict indices,
+                    const uint32 *restrict dimsizes,
+                    int len) {
+    for (int i = 0; i < len; ++i) {
+        if (unlikely(indices[i] >= dimsizes[i])) {
+            _DML_assert_indices_fail(logname, method_name, indices, dimsizes,
+                                     len);
+        }
+    }
+}
+
+// Passed 'indices' and 'dimsizes' are expected to be compound literals.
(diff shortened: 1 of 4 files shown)


TSonono:
It did indeed resolve the issue I encountered, nice work!

syssimics:
Verification [#13511456](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13511456): ✅ pass

py/dml/ctree.py
```diff
@@ -595,6 +596,27 @@ def mkInlinedMethod(site, method, pre, body, post):
...
+            self.method.dimsizes)
+        fmt = (f"Attempt made to call {logname_indices}(), which involves "
+               + "one or more out-of-bounds array accesses. Array dimensions "
+               + f"are as follows: {logname_dimsizes}")
+        check = ' && '.join(f'({idx.read()}) < {size}'
+                            for (idx, size) in zip(self.indices,
+                                                   self.method.dimsizes))
+        self.linemark()
+        idx_args = ', '.join(f'({idx.read()})' for idx in self.indices)
+        out(f'ASSERT_FMT({check}, "{fmt}", {idx_args});\n')
```
mandolaerik, line 616: `FATAL_ERROR_IF` is preferred over `ASSERT_FMT`. Also, I'm a bit worried about increased code expansion; the message is duplicated many times. Perhaps the check should be abstracted into a separate C function in dmllib.h.

mandolaerik: Perhaps a message based on (object-identity, dml-filename, dml-line, variable-name) is sufficient?

include/simics/dmllib.h
```diff
@@ -38,7 +38,56 @@ static inline uint64 DML_combine_bits(uint64 x, uint64 y, uint64 mask)
 #define DML_ASSERT(filename, lineno, expr) do {         \
     if (unlikely(!(expr)))                              \
         assert_error((lineno), (filename), "", NULL);   \
-} while (0)
+    } while (0)
```
lwaern-intel, line 41: this stray diff should be removed

py/dml/ctree.py
lwaern-intel: Replaced this with an mechanism that should tick all boxes.

syssimics:
Verification [#13515947](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13515947): ✅ pass

include/simics/dmllib.h
```diff
@@ -38,7 +38,56 @@ static inline uint64 DML_combine_bits(uint64 x, uint64 y, uint64 mask)
...
+        if (unlikely(indices[i] >= dimsizes[i])) {
+            _DML_assert_indices_fail(logname, method_name, indices, dimsizes,
+                                     len);
+        }
+    }
+}
+
+// Passed 'indices' and 'dimsizes' are expected to be compound literals.
+// This indirection is to delimit their life span.
+#define DML_ASSERT_INDICES(logname, method_name, indices, dimsizes, len) do { \
```
mandolaerik, line 87: why a macro indirection?

py/dml/ctree.py
```diff
@@ -595,6 +596,22 @@ def mkInlinedMethod(site, method, pre, body, post):
...
+
+    def toc_stmt(self):
+        self.linemark()
+        length = len(self.indices)
+        indices = ', '.join(f'({idx.read()})' for idx in self.indices)
+        dimsizes = ', '.join(map(str, self.method.dimsizes))
+        out('DML_ASSERT_INDICES('
+            + f'_id_infos[{self.method.parent.uniq - 1}].logname, '
+            + f'"{self.method.name}", ((const uint32 []) {{ {indices} }}), '
+            + f'((const uint32 []) {{ {dimsizes} }}), {length});\n')
```
mandolaerik, line 611: I don't remember details, but IIRC, compound literals *sometimes* induce GCC slowness (I think this is a main source of vtable init slowness). IIRC it's ok as long as the total size is small, but I don't remember. Deserves profiling, since this affects all methods.

include/simics/dmllib.h
lwaern-intel: it

lwaern-intel: it says right there

py/dml/ctree.py
lwaern-intel: benching reveals the use of compound literals make no difference vs. array variables.

include/simics/dmllib.h
lwaern-intel: No longer relevant regardless.

py/dml/ctree.py
lwaern-intel: Though the loop _is_ relevant and slow. Tweaking eliminates the overhead.

syssimics:
Verification [#13517920](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13517920): ✅ pass

syssimics:
Verification [#13520767](https://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13520767): ✅ pass

mandolaerik (approved):

PR #304, opened 2024-07-08
Generalize WNEGCONSTCOMP to warn for all constants outside a type's range

mandolaerik:


syssimics:
Verification [#13597323](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13597323): ❌ fail

py/dml/messages.py
```diff
@@ -2110,16 +2110,18 @@ def log(self):
...
+    can be rewritten as `cast(-1, uint64)`.
+    """
     def __init__(self, site, expr, ty):
-        DMLWarning.__init__(self, site)
+        super().__init__(site)
         self.expr = expr
         self.ty = ty
-    fmt = ("Comparing negative constant to unsigned integer has a constant "
-           + "result")
+    fmt = 'Comparison with constant outside the range of data type'
```
lwaern-intel, line 2124: I'm slightly worried this will be hard to understand... maybe
> comparing a value of type '%s' against a constant outside the range of that type

possibly s/against/with. This might also need some tweaking in the case of e.g. `cast(a, uint32) > 0xffffffff`, where  `0xffffffff` is in _range_ of uint32, and yet this check is still always constant false. (Your code does handle this case).

lwaern-intel (changes requested):

py/dml/ctree.py
```diff
@@ -1602,10 +1608,18 @@ def make(cls, site, lh, rh):
...
+                    not (lhtype.min() <= rh.value <= lhtype.max())
+                    or (rhtype.is_float and rh.value % 1)):
+                report(WTYPELIMITS(site, rh, lhtype))
             lh_maybe_negative = lhtype.signed
             lh = as_int(lh)
             lhtype = realtype(lh.ctype())
         if rhtype.is_int:
+            if lh.constant and lhtype.is_arith and (
+                    not (rhtype.min() <= lh.value <= rhtype.max())
+                    or (lhtype.is_float and lh.value % 1)):
```
lwaern-intel, line 1621: Huh. `x % 1` does actually seem to Just Work, even for nan and inf. I'm surprised.
My earlier way to go about this was `not math.isfinite(x) or x != int(x)`. Your idiom has the benefit of `x` only occurring once, at the cost of being more obtuse.

lwaern-intel: I think I'm fine with the `% 1` route.

syssimics:
Verification [#13597509](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13597509): ✅ pass

py/dml/ctree.py
mandolaerik: Good point, inf/nan deserve test cases.

mandolaerik: Indeed broken, I get no warning from the following:
```
    local int64 b = -1;
    assert b < inf;
    assert inf > b;
    assert b > -inf;
    assert b != -inf;
    assert !(b < nan);
    assert !(b > nan);
    assert b != nan;
```

py/dml/messages.py
mandolaerik: I changed to gcc's phrasing, "comparison is always false/true due to limited range of data type", at the cost of a flag programming hack to know whether we are in the `!=` inverted case

syssimics:
Verification [#13597668](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13597668): ✅ pass

syssimics:
Verification [#13597716](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13597716): ✅ pass

lwaern-intel:
@mandolaerik Have you run this past Jenkins?

mandolaerik:
no, do they use --werror?

lwaern-intel:
Some teams do for some devices for some platforms.

lwaern-intel:
More to the point, I want to make sure this won't cause overt logspam. We might want a compatibility feature like `WLOGMIXUP`.

PR #306, opened 2024-07-26
Typecheck shared methods after trait processing -- SIMICS-22166

lwaern-intel:
This allows for the use of any template type in the signature of any shared method. The logic has also been extended to check that each named type in the signature is valid.

This new releasenotes format is confusing me. I can only hope this works.

The validation of shared method signatures, even if unused, could potentially break VP as it happens even if the trait is considered dead. I'll run a pretest just in case.

### DIFF
--- a/py/dml/traits.py
+++ b/py/dml/traits.py
@@ -66,8 +66,6 @@ def process_trait(site, name, subasts, ancestors, template_symbols):
                               + ' '.join(['independent']*independent
                                          + ['startup']*startup
                                          + ['memoized']*memoized))
-                # TODO: a trait method cannot use a subtrait in the
-                # argument list
                 inp = eval_method_inp(inp_asts, None, global_scope)
                 outp = eval_method_outp(outp_asts, None, global_scope)
                 check_namecoll(mname, ast.site)
@@ -349,18 +347,13 @@ def mktrait(site, tname, ancestors, methods, params, sessions, hooks,
                     report(EDMETH(msite, orig_trait.method_impls[name].site,
                                   name))
                     bad_methods.add(name)
-                else:
-                    if name not in ancestor_vtables:
-                        raise ICE(msite,
-                                  'ancestor is overridable but not in vtable')
-                    try:
-                        typecheck_method_override(
-                            (msite, inp, outp, throws, independent, startup,
-                             memoized),
-                            ancestor_vtables[name].vtable_methods[name])
-                    except DMLError as e:
-                        report(e)
-                        bad_methods.add(name)
+                elif name not in ancestor_vtables:
+                    raise ICE(msite,
+                              'ancestor is overridable but not in vtable')
+
+                # Type-checking of overrides is done later, after typedefs
+                # have been populated with all template types.
+                # See Trait.typecheck_methods()
 
     for name in bad_methods:
         del methods[name]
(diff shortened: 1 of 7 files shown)


lwaern-intel:
Jenkins run: https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=13662584

syssimics:
Verification [#13662464](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13662464): ✅ pass

py/dml/traits.py
```diff
@@ -759,6 +752,40 @@ def __lt__(self, other):
...
+                    typecheck_method_override(
+                        (sm.site, sm.inp, sm.outp, sm.throws, sm.independent,
+                         sm.startup, sm.memoized),
+                        self.ancestor_vtables[sm.name].vtable_methods[sm.name])
+                except DMLError as e:
+                    report(e)
+
+        # We don't attempt to purge the bad methods. Subtemplates could
+        # be referencing them, and there is no good way to purge those
+        # references. This doesn't seem to be too big of a deal; DMLC stops
```
mandolaerik, line 785: when we have reported an error, the ambition is to cover up to the point that we dont ICE. it is fine e.g. if the fallback causes the c backend to generate broken / inconsistent code, or if it sometimes reports weird secondary errors.

py/dml/types.py
```diff
@@ -88,6 +89,32 @@ def __str__(self):
...
+        for (mn, mt) in t.members.items():
+            check_named_types(mt)
+    elif isinstance(t, (TPtr, TVector, TArray)):
+        check_named_types(t.base)
+    elif isinstance(t, TFunction):
+        for pt in t.input_types:
+            check_named_types(pt)
+        check_named_types(t.output_type)
+    elif isinstance(t, TTraitList):
+        if t.traitname not in dml.globals.traits:
```
mandolaerik, line 108: perhaps an early assert that globals.traits is initialized?

py/dml/types.py
```diff
@@ -88,6 +89,32 @@ def __str__(self):
 # anonymous structs used by global typedefs, label -> type
 global_anonymous_structs = {}
 
+def check_named_types(t):
+    '''Checks that a type does not reference a non-existing type'''
+    if isinstance(t, TNamed):
```
mandolaerik, line 94: match/case?

lwaern-intel: I rather not piss off fisketur.
Tangentially, the very existence of `check_named_types` is questionable considering `_ = safe_realtype(t)`, from what I can tell, fits exactly the same niche. I employed it for verifying shared method signatures because it's what we employ to verify _regular_ method signatures. I'm not sure if there's anything that makes `check_named_types` preferable. What's your opinion?

py/dml/types.py
lwaern-intel: Considering `dml.globals.traits` starts off as `None` It would be noticeable either way (`t not in None` is not legal.) An assert does solidify the precondition, though. I'm ambivalent, I'll add an assert if you insist.

py/dml/traits.py
lwaern-intel: Indeed. This comment is because, before this PR, the bad shared methods _used_ to be pruned (and still are if they fail on something other than the signature.) But now that the typechecking happens so late that's no longer feasible (or, at least, not easy.) From my (admittedly shallow) investigation it doesn't _seem_ to be provoking ICE:s; however, the fact that the overriding method will linger about with an incorrect signature means methods can invoke it, which can lead to hilarity. That's the downside.

py/dml/types.py
mandolaerik: then an explicit assert is not warranted. i think my only concern is that too early calls to this function are only captured if trait types are involved, making that hypothetical bug harder to catch. not a  disaster.

perhaps we should ideally have well-defined compile phases, and say that this function is only available after a particular phase.

lwaern-intel: I very much like that idea. In particular, I've internally always lamented the fact that it's very unclear when `realtype` becomes safe to use. Having a phase after which you are guaranteed the ability to validate and normalize any type reference, and having `realtype` ICE if used before then, is a very tempting idea to me.

This'll mean rigorously splitting up the `structure` phase into subphases. What's nice I don't even think it'd take much for us to be able to establish these subphases: sure, there are some particular questions that complicate matters (e.g. there are some corners in resolving types besides `TNamed`, e.g. `Layout.resolve()` and `Hook.validate()`) but I think these are solvable. I'm thinking something like...
- Collecting all global declarations (in particular, templates and types)
- Resolve and verify global types
- Resolve and verify all other global declarations

... after which you deal with the device tree (which could also potentially be split into subphases)

mandolaerik: unfortunately, conditional imports make matters more complicated. so i guess it is not a topic for this PR.

mandolaerik (approved):

syssimics:
Verification [#13791577](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13791577): ✅ pass

PR #307, opened 2024-08-06
Added length method to bank template

TSonono:
Does not add startup time but utilizes the startup memoized method `_sorted_regs()` which means this method should not scale on bank length.

Feel free to close this PR if it doesn't make sense to include this in dml-builtins.

### DIFF
--- a/lib/1.4/dml-builtins.dml
+++ b/lib/1.4/dml-builtins.dml
@@ -1954,6 +1954,15 @@ template bank is (object, shown_desc) {
         return (src, num);
     }
 
+    shared independent method length() -> (uint64) {
+        local (const register *regs, int num_mapped_regs) = _sorted_regs();
+        if (num_mapped_regs > 0) {
+            local const register last_reg = regs[num_mapped_regs - 1];
+            return last_reg.offset + last_reg._size();
+        }
+        return 0;
+    }
+
     // Register dispatcher. Given the offset and size of an access, return a
     // list of registers that the access covers, and a bitmask showing 0xff for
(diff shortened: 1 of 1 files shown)


mandolaerik:
I'm not so concerned about startup time, so `memoized` would be fine (the static cost is an integer per bank, 8 to 16 bytes in total).

If we are concerned about this cost, then we can make it an optional utility template, like:
```
template bank_length is bank { shared independent startup memoized method length() -> (uint64) { ...}
```
so whoever wants it universally can say:
```
in each bank { is bank_length; }
```

mandolaerik:
And as usual, naming is a hard problem. "Length" is somewhat ambiguous (other interpretations: number of regs, sum of reg sizes, difference between lowest and highest offs of a register byte). One option could be to let a method return the highest-offset `register` object; here, static typing would help eliminating ambiguity in naming.

mandolaerik:
Also, we need to consider implications for custom implementations of read or write on bank level (such as "simple regs"): Would the existence of a `length` method mean that `read`/`write` overrides are expected to override the `length` method, to return the largest address with custom behaviour? This is a semantic subtlety that a `register` return type may help to address: it enforces an unambiguous technical definition; we *can* only concrete statically declared `register` objects, so the only sensible return value is the one with the highest `offset`.

mandolaerik:
Also, once we agree on a design, we also need to extend the PR with tests and docs.

syssimics:
Verification [#13711935](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13711935): ✅ pass

PR #311, opened 2024-08-19
Added initial TM grammar file taken from vscode extension

TSonono:
This would be used in an effort of bringing syntax highlighting support for DML in Github. The content of the file was taken from the modeling vscode extension, which is currently not a public repository.


MagnusIntel (approved):
I don't know if we ever reviewed this in detail, but I guess it is a good enough starter.

syntaxes/Dml.tmlanguage
```diff
@@ -0,0 +1,1175 @@
...
+		</dict>
+		<dict>
+			<key>match</key>
+			<string>\b(break|case|continue|default|do|else|for|goto|if|_Pragma|return|switch|while|foreach|select|where)\b</string>
+			<key>name</key>
+			<string>keyword.control.c</string>
+		</dict>
+		<dict>
+			<key>match</key>
+			<string>\b(asm|__asm__|auto|bool|_Bool|char|_Complex|double|enum|float|_Imaginary|int|long|short|signed|struct|typedef|union|unsigned|void)\b</string>
```
lwaern-intel, line 48: These are leftover C aspects that rather obviously don't belong in DML `asm`, `__asm__`, `auto`, `_Bool`, `_Complex`, `enum`, `_Imaginary`, and `union`. However, they might still be helpful for `header %{ %}` blocks. Possibly. I don't think this blocks a merge.

lwaern-intel (approved):
From a quick glance, I suspect there's room for improvement, but we can deal with that later.

TSonono:
> From a quick glance, I suspect there's room for improvement, but we can deal with that later.

I agree. I just took this file from the vscode extension and I'm not even sure who wrote it. All I know is that is adds a "good enough" syntax highlighter in vscode.

This file can always change down the line. If linguist would approve DML as a language, we can still make changes to this file and for every release of linguist, our changes here would be reflected in the production version of linguist.

JonatanWaern:
The file is originally lifted from some SSM repo somewhere, I think. Before we merge this in we have some outstanding grammar changes pending on the vscode side that should be merged in IMO. And probably take more than a second look at it.

I'd rather not duplicate this file between the vscode repo and here. So this might be a good case for a git submodule, or something (not overly familiar with all the fun ways to link git repos)? (although, frankly, the DML repo is probably a better authoritive source for it, so the submodule would be in the vscode extension repo) I also suspect that if when we add this grammar as a remote to the linguist upstream it wont be the entirety of the DMLC repo anyway, right?

MagnusIntel:
This PR is more about pipe cleaning the flow between a DML TM Language file hosted in the public repository (e.g. on github.com/intel) and linguist. We should certainly review and update the file itself and if we already have outstanding PR, lets merge those.

Regarding sharing files between public repos, I don't know what is best practice nor do I know what linguist requires. But creating another public repo takes quite some work and if we at all can put the TM language in the DML repo, I'd rather do that. As you say, it does fit well in the DML repo in any case. But when publish the extension, we need to figure out how VS-code can get it from this repo.

TSonono:
The linguist repo would be sudmodule:ing this repo, the vscode extension repo could simply do the same.

TSonono:
While I intend on merging this now. The PR to the linguist repo won't be made until I have the permission to do so, so changes can be made to this file in the meanwhile. Further more, like I said before we can keep updating the file in perpetuity, new releases of linguist would take those changes in to account.

PR #313, opened 2024-08-20
Add the access_count explicitly

fredriklarssonintelcom:
Soon in Simics, the access_count is not added for objects that do not need it. So the test class used here does not get the access_count attribute. So the test fails for the expected attributes. By adding it explicitly here, the test will pass.

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -1700,6 +1700,12 @@ class CompareIllegalAttrs(BaseTestCase):
                       encoding='utf-8').stdout.splitlines()
               if x.startswith('ATTR ')}
 
+        # access count is added for objects implementing io_memory or
+        # transaction interfaces. In DML the name is illegal to avoid a clash in
+        # bank objects. TODO: we should adjust the test to also extract the set
+        # of attributes from an empty bank, not only the device object.
+        sl.add("access_count")
+
         # Extract list of illegal attributes from dmlc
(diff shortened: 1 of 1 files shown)


syssimics:
Verification [#13772840](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13772840): ✅ pass

fredriklarssonintelcom:
Please approve. The test can be improved later.

test/tests.py
```diff
@@ -1700,6 +1700,10 @@ def test(self):
                       encoding='utf-8').stdout.splitlines()
               if x.startswith('ATTR ')}
 
+        # access count is added for objects implementing io_memory or transaction
+        # interfaces, but not for any other object, i.e., it is illegal in dml
+        sl.add("access_count")
```
mandolaerik, line 1705: I think this is the best path in the short term (for merge technical reasons), but should be improved. I suggest a comment like:
```
# access count is added for objects implementing io_memory or transaction
# interfaces. In DML the name is illegal to avoid a clash in bank objects.
# TODO: we should adjust the test to also extract the set of attributes from
# an empty bank, not only the device object.
```

mandolaerik (approved):

syssimics:
Verification [#13818651](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13818651): ❌ fail

PR #314, opened 2024-08-23
Make `explicit_param_decls` stable

lwaern-intel:
I want this in in time for SQPFU.

### DIFF
--- a/RELEASENOTES-1.4.md
+++ b/RELEASENOTES-1.4.md
@@ -320,3 +320,4 @@
 - `release 6 6321`
 - `release 7 7033`
 - `release 6 6324`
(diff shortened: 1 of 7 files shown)


syssimics:
Verification [#13774908](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13774908): ✅ pass

mandolaerik (approved):
Two remarks, no need to fix:
* "grammatically" in a commit msg
* we could consider an ESYNTAX test for `param x :;`

lwaern-intel:
> * we could consider an ESYNTAX test for `param x :;`

It'd be a bit of an odd-duck as it's forbidden _grammatically_, as in it'd be a parse error. We don't typically have tests for those.

syssimics:
Verification [#13782576](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13782576): ✅ pass

PR #315, opened 2024-08-26
Add picoseconds (`ps`) as a unit of time -- SIMICS-16019

lwaern-intel:
i dislike how obviously-but-not-to-what-degree separate the picosecond clock is

i really want to know about its caveats more

consider this a partially draft PR until we find out more

### DIFF
--- a/doc/1.4/language.md
+++ b/doc/1.4/language.md
@@ -3466,7 +3466,8 @@ after <em>scalar</em> <em>unit</em>: <em>method</em>(<em>e1</em>, ... <em>eN</em
 In this form, the specified point in the future is given through a time delay
 (in simulated time, measured in the specified time unit) relative to the time
 when the after delay statement is executed. The currently supported time units
-are `s` for seconds and `cycles` for cycles.
+are `s` for seconds (with type `double`), `ps` for picoseconds
+(with type `uint64`), and `cycles` for cycles (with type `uint64`).
 
 Every argument to the called method is evaluated at the time the `after`
 statement is executed, and stored so that they may be used when the method call
--- a/py/dml/c_backend.py
+++ b/py/dml/c_backend.py
@@ -1231,14 +1231,28 @@ def generate_immediate_after_callbacks(device):
 
 def generate_simple_events_control_methods(device):
     start_function_definition(
-        'void _cancel_simple_events(conf_object_t *_obj, _identity_t domain)')
+        'void _cancel_simple_delay_events(conf_object_t *_obj, '
+        'conf_object_t *_clock, const _identity_t *_domain)')
+    out('{\n', postindent=1)
+    out('int (*_pred)(lang_void *, lang_void *) UNUSED = '
+        + '_domain ? _simple_event_predicate : NULL;\n')
+    for key in dml.globals.after_delay_infos:
+        out('SIM_event_cancel_time('
+            + f'_clock, {crep.get_evclass(key)}, _obj, _pred, '
+            + '(lang_void *)_domain);\n')
+    out('}\n', preindent=-1)
+
+    start_function_definition(
+        'void _cancel_simple_events(conf_object_t *_obj, _identity_t _domain)')
     out('{\n', postindent = 1)
     out(crep.structtype(device) + ' *_dev UNUSED = ('
         + crep.structtype(device) + '*)_obj;\n')
-    for key in dml.globals.after_delay_infos:
-        out('SIM_event_cancel_time('
-            + f'SIM_object_clock(_obj), {crep.get_evclass(key)}, _obj, '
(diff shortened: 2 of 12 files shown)


syssimics:
Verification [#13783245](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13783245): ✅ pass

py/dml/c_backend.py
```diff
@@ -1231,14 +1231,30 @@ def generate_immediate_after_callbacks(device):
 
 def generate_simple_events_control_methods(device):
     start_function_definition(
-        'void _cancel_simple_events(conf_object_t *_obj, _identity_t domain)')
-    out('{\n', postindent = 1)
+        'void _cancel_simple_delay_events(conf_object_t *_obj, '
+        'conf_object_t *_clock, const _identity_t *_domain)')
+    out('{\n', postindent=1)
     out(crep.structtype(device) + ' *_dev UNUSED = ('
         + crep.structtype(device) + '*)_obj;\n')
```
mandolaerik, line 1238: this looks unused? Looks like you can move the function into `dmllib.h` if you remove it?

lwaern-intel: Ah, I assumed `crep.get_evclass(key)` needed `_dev` but upon closer thought it doesn't.

You can't have it in `dmllib.h` in either case. No access to `dml.globals.after_delay_infos`.

py/dml/codegen.py
```diff
@@ -2720,9 +2720,15 @@ def stmt_after(stmt, location, scope):
     delay = codegen_expression(delay, location, scope)
     old_delay_type = delay.ctype()
     if unit == 's':
+        clock = 'SIM_object_clock(&_dev->obj)'
```
mandolaerik, line 2723: We should arguably switch to ps clock also for seconds. That however deserves a different PR, and is possibly breaking enough to warrant a compat feature.

py/dml/c_backend.py
```diff
@@ -1231,14 +1231,30 @@ def generate_immediate_after_callbacks(device):
 
 def generate_simple_events_control_methods(device):
     start_function_definition(
-        'void _cancel_simple_events(conf_object_t *_obj, _identity_t domain)')
-    out('{\n', postindent = 1)
+        'void _cancel_simple_delay_events(conf_object_t *_obj, '
```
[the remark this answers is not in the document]
lwaern-intel: The `simple_event` naming scheme is a historical artifact from the first implementations of `after`. I agree, revamping the naming convention is warranted, but not as part of this PR.

doc/1.4/language.md
```diff
@@ -3466,7 +3466,8 @@ after <em>scalar</em> <em>unit</em>: <em>method</em>(<em>e1</em>, ... <em>eN</em
 In this form, the specified point in the future is given through a time delay
 (in simulated time, measured in the specified time unit) relative to the time
 when the after delay statement is executed. The currently supported time units
-are `s` for seconds and `cycles` for cycles.
+are `s` for seconds (fractional), `ps` for picoseconds (integer),
```
mandolaerik, line 3469: Although the term 'fractional' is a pretty good word here, we don't use it elsewhere in the document. Perhaps "floating-point" is better?

Wording is somewhat tricky in this domain because the true technical term is the hopelessly ambiguous word 'double'.

py/dml/c_backend.py
mandolaerik: good point

doc/1.4/language.md
lwaern-intel: "Fractional numbers" is a pretty standard way to refer to, well, fractional numbers. I think using "fractional" is fine.

lwaern-intel: Upon further consideration, I would do
> are `s` for seconds (`double`), `ps` for picoseconds (`uint64`), `cycles` for cycles (`uint64`)

Agreed?

lwaern-intel: maybe
> (with type `double`)
?

mandolaerik: > "Fractional numbers" is a pretty standard way to refer to, well, fractional numbers. I think using "fractional" is fine.

I agree that fractional is a sane word here in principle, but it is inconsistent with the rest of our docs. E.g. we say things like "Floating-point numbers: There is only one floating-point type, called `double`. It corresponds to the C type `double`.". I would be OK with a global `s/floating-point/fractional` in our docs, but mixing the two adds unnecessary confusion.

I'm also OK with "type `double`", or just floating-point, if you prefer that.

syssimics:
Verification [#13791106](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13791106): ✅ pass

mandolaerik (approved):

PR #316, opened 2024-08-30
Support `continue` within `foreach` (on sequences)

lwaern-intel:
My compilation issue fixing on simple regs revealed we still don't support `continue` within `foreach`. This is dumb, so here's a quick PR to allow it.

I couldn't find a JIRA issue about this.


py/dml/codegen.py
```diff
@@ -104,18 +104,26 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         assert LoopContext.current is self
         LoopContext.current = self.prev
+    def break_(self, site):
+        raise ICE(site, f'{__class__.__name__}.break_ not implemented')
+    def cont_(self, site):
+        raise ECONTU(site)
```
lwaern-intel, line 110: Motivation for having `break_` ICE but not `cont_`: `ECONTU` exists. `EBREAKU` doesn't.

py/dml/codegen.py
```diff
@@ -104,18 +104,26 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         assert LoopContext.current is self
         LoopContext.current = self.prev
+    def break_(self, site):
+        raise ICE(site, f'{__class__.__name__}.break_ not implemented')
+    def cont_(self, site):
```
mandolaerik, line 109: `continue_`?
also, if this didnt cause a ECONTU test failure then the test should be fixed

py/dml/codegen.py
mandolaerik: would `@abstractmethod` be an option?

mandolaerik (changes requested):

py/dml/codegen.py
lwaern-intel: Oops! `cont_` is a bad old artifact.
And no, I just forgot to run `1.2/errors/ECONTU` and only ran the 1.4 tests; so it's not a test gap.

py/dml/codegen.py
lwaern-intel: It would.

mandolaerik (approved):

syssimics:
Verification [#13801526](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13801526): ✅ pass

PR #317, opened 2024-08-30
Relax rule for naming groups 'port'/'bank'

lwaern-intel:
~This exists purely for TC to be aware of it.~ Now a proper PR.

### DIFF
--- a/lib/1.2/dml-builtins.dml
+++ b/lib/1.2/dml-builtins.dml
@@ -185,12 +185,9 @@ template group {
     is object;
     parameter objtype = "group";
     parameter log_group = undefined;
-    if (($this._nongroup_parent.objtype == "port"
-         || $this._nongroup_parent.objtype == "device")
-        && $this.name == "port") {
-         error "Cannot name a group 'port' in a location where"
-             + " port objects are allowed";
-    }
+    parameter _simics_namespace_clash =
+        $this.name == "port" || ($parent.objtype == "group"
+                                 ? $parent._simics_namespace_clash : false);
 }
 
 template device {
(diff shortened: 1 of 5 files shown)


syssimics:
Verification [#13874923](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13874923): ✅ pass

lib/1.2/dml-builtins.dml
```diff
@@ -386,6 +383,13 @@ template bank {
     is _hard_reset;
     is _soft_reset;
 
+    // Redundant due to SIMICS-19009; banks must be top-level in 1.2 devices
+    //
+    // if ($parent.objtype == "group" ? $parent._simics_namespace_clash
+    //     : false) {
+    //     error "Cannot declare a bank underneath a group named 'port'";
+    // }
```
mandolaerik, line 391: I see why you write this, but commented-out code is a pretty strong anti-pattern. Can it be expressed concisely enough on some other form, e.g. English?

lib/1.4/dml-builtins.dml
```diff
@@ -659,19 +659,17 @@ as parent.
...
     param objtype = "group";
     param _is_simics_object = false;
     param log_group = undefined;
-    #if ((_nongroup_parent.objtype == "subdevice"
-          || _nongroup_parent.objtype == "device")
-         && (this.name == "port" || this.name == "bank")) {
-         error "Cannot name a group 'port' or 'bank' in a location where"
-             + " port or bank objects are allowed";
-    }
+    param _simics_namespace_clash =
```
mandolaerik, line 669: It's initially struck me as concerning that we add a new param to our most lightweight object. But I doubt it can be a problem, should be the same kind of expensive as the `#if` we remove.

mandolaerik (approved):
Consider fixing the commented-out code if you agree, but ok to merge if you don't agree.

lib/1.2/dml-builtins.dml
lwaern-intel: Yeah, I can.

PR #318, opened 2024-09-02
Don't emit linemarks for implicit returns -- SIMICS-22685

lwaern-intel:
Also some refactoring to make use of a documented `as_simple_site` function for the `SimpleSite(site.loc())` pattern

@TSonono try this PR out and see if you're satisfied with the behavior, or if there's some other strangeness that remains and sticks out to you.

### DIFF
--- a/RELEASENOTES-1.4.md
+++ b/RELEASENOTES-1.4.md
@@ -332,3 +332,7 @@
 - `release 6 6329`
 - `note 6` `continue` can now be used within `foreach` loops (use within
   `#foreach` loops remains unsupported)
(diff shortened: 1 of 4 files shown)


syssimics:
Verification [#13810578](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13810578): ✅ pass

TSonono:
Will try it out today, thanks!

lwaern-intel:
Turns out C implicit linemarks _are_ being considered by coverage tools (and likely debuggers as well.) This PR is thus invalid.

PR #319, opened 2024-09-04
`destroy` template -- SIMICS-16161

lwaern-intel:
This is to get the ball rolling and force you to accept or reject it ;)

### DIFF
--- a/RELEASENOTES-1.4.md
+++ b/RELEASENOTES-1.4.md
@@ -334,3 +334,7 @@
   `#foreach` loops remains unsupported)
 - `release 7 7037`
 - `release 6 6330`
+- `note 6` Added the `destroy` template, the device deletion equivalent of the
+  `init` and `post_init` templates. When the device is being deleted,
+  `destroy()` will be called of every object that instantiates the `destroy`
+  template. The device object always instantiates this template.
--- a/lib/1.4/dml-builtins.dml
+++ b/lib/1.4/dml-builtins.dml
@@ -408,6 +408,56 @@ template post_init {
 }
 
 /**
+### destroy
+Provides an abstract method `destroy`, which is called when the
+device is being deleted. Typically used to `delete` any device state that has
+been dynamically allocated using `new`.
+
+The method `destroy` is automatically called on all objects that implement the
+`destroy` template. The order in which `destroy` of objects are called is not
+defined, except that `destroy` of a particular object is guaranteed to be
+called before `destroy` of any of its parent objects.
+In particular, `destroy` of the device object will be called only after all
+other implementations of `destroy`.
+
+Usage note: while the device is being deleted, it is not allowed to communicate
+with any other Simics object (whether that is by accessing `connect`ed devices
+or by leveraging the Simics API.) This includes even the device's own clock;
+don't attempt to post or cancel time/cycle-based events in `destroy()`
+(including posting events using `after` with a time/cycle delay). The
+cancellation of such events is handled automatically, before the `destroy()`
+calls are invoked. The use of `.cancel_after()` inside `destroy()` is tolerated
+(even if almost always redundant.)
+
+<div class="note">
+
+**Note:** the `destroy` template can't be instantiated for
+`event` objects. This is because `event` objects, through the
+`custom_time_event` or `custom_cycle_event` template, may already require
+defining a method named `destroy` whose purpose is different from the one
+required by the `destroy` template. To work around this limitation, you may
+declare a `group` within the `event` object that instantiates the `destroy`
+template instead.
+
+</div>
+*/
+template destroy {
+    param _each_destroy : sequence(destroy);
+    param _each_destroy = each destroy in (this);
+    shared method _rec_destroy() {
+        foreach obj in (this._each_destroy)
+            obj._rec_destroy();
+        this.destroy();
+    }
+    // Perform local destruction (deallocation etc.)
+    shared method destroy();
+
+    #if (this.objtype == "event") {
+        error "event objects may not instantiate the 'destroy' template";
+    }
+}
+
+/**
 ### object
 
 Base template, implemented by all objects. Inherits the templates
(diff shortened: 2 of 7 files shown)


syssimics:
Verification [#13817248](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13817248): ❌ fail

lwaern-intel:
https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=13817259

lwaern-intel:
Another verification run with compatibility patch: https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=13826183

RELEASENOTES-1.4.md
```diff
@@ -334,3 +334,7 @@
   `#foreach` loops remains unsupported)
 - `release 7 7037`
 - `release 6 6330`
+- `note 6` Added the `destroy` template, the device deletion equivalent of the
+  `init` and `post_init` templates. When the device is being deleted,
+  `destroy()` will be called of every object that instantiates the `destroy`
+  template (the device object always instantiates it.)
```
mandolaerik, line 340: I'd strip the bracket to improve text flow, e.g. `destroy template. The device object always instantiates this template.`

lib/1.4/dml-builtins.dml
```diff
@@ -407,6 +407,52 @@ template post_init {
...
+called before `destroy` of any of its parent objects.
+In particular, `destroy` of the device object will be called only after all
+other implementations of `destroy`.
+
+The `destroy` template can't be instantiated for `event` objects. This is
+because `event` objects, through the `custom_time_event` or
+`custom_cycle_event` template, may already require defining a method named
+`destroy` whose purpose is different from the one required by the `destroy`
+template. To work around this limitation, you may declare a `group` within the
+`event` object that instantiates the `destroy` template instead.
```
mandolaerik, line 429: This section maybe deserves to be deemphasised through `<div class="note"> **Note:**`

lib/1.4/dml-builtins.dml
```diff
@@ -407,6 +407,52 @@ template post_init {
...
+
+Usage note: while the device is being deleted, it is not allowed to communicate
+with any other Simics object (except the device's `bank`s, `port`s and
+`subdevice`s). This includes even its own clock; don't attempt to post or
```
mandolaerik, line 433: The exception is questionable -- I'd say it's dubious to SIM_get_interface on your own port objects during shutdown (you will get away with it but not sure if we offer any guarantees). And it is *unsafe* to interact with state in private port objs that `destroy` destroys (because of hierarchical destroy order). For portobj-related state owned by the device struct,  like `saved` variables, it is so obviously safe to interact with it during `destroy` that we don't need to mention it.

mandolaerik: We could instead mention why, e.g. `with any other simics object. This is because that any other objects may be incapable to interaction due to being simultaneously deleted,`

Actually I don't fully remember the details -- are there device interactions that can be safe during `destroy` if you own both devs and know what you are doing, e.g. calling an interface that only relies on data coallocated with the device struct? or does simics reserve the right to deallocate other objects before destroy has been invoked?

lib/1.4/dml-builtins.dml
```diff
@@ -545,6 +592,8 @@ template device {
     // _init and _post_init are called automatically on object instantiation
     is init;
     is post_init;
+    // _delete is called automatically on object deletion
```
mandolaerik, line 595: `s/_delete/_destroy/`?

syssimics:
Verification [#13827354](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13827354): ✅ pass

mandolaerik (commented):
The change looks good, just some minor remarks on text

lib/1.4/dml-builtins.dml
lwaern-intel: yup

lib/1.4/dml-builtins.dml
lwaern-intel: > The exception is questionable -- I'd say it's dubious to SIM_get_interface on your own port objects during shutdown (you will get away with it but not sure if we offer any guarantees)
> For portobj-related state owned by the device struct, like `saved` variables, it is so obviously safe to interact with it during destroy that we don't need to mention it.

The latter's exactly why I bring it up, You say it's "obviously safe", but I'm not so sure. I just want to preempt the question of "wait, my banks are other objects. does this mean I can't do anything in my banks in `destroy()`"?

> We could instead mention why

That might be a better way to imply why the interaction's unsafe... unless, again, people may wonder "wait, my _banks_ get deleted at the same time, right? Does that mean I can't do anything with them in `destroy()`"
> are there device interactions that can be safe during destroy if you own both devs and know what you are doing?

Not sure but I can't imagine the Simics API enshrines any as acceptable.

lib/1.4/dml-builtins.dml
lwaern-intel: Yup, and if so, it should be moved to the last part of the docs.

RELEASENOTES-1.4.md
lwaern-intel: Makes sense.

lib/1.4/dml-builtins.dml
mandolaerik: > n understanding of the model. They'll know what destroy() of each bank/port/subdevices does and doesn't do, and what is safe to interact with within destroy() of, say, the device object. Or am I misunderstanding you?
I meant that interaction with state that is destroyed is unsafe. The following has a 50% probability of crashing, kind of.
```
group x {
  session int *x;
  method init() { x = new int[3]; }
  method destroy() { delete x; }
}
group y {
  method destroy() {
    log info: "%d", x.x[0];
  }
}

Comment (lwaern-intel) [reply lib/1.4/dml-builtins.dml:433]: Oh yes, certainly. My point is that, say, the following is safe:

```
bank b {
  session int *x;
  ...
}

method init() { x = new int[3]; }
method destroy() { delete x; }
```

Comment (lwaern-intel) [reply lib/1.4/dml-builtins.dml:433]: I think the best way to communicate my point is to clarify what is meant by communicate in this context:
```
Usage note: while the device is being deleted, it is not allowed to communicate
with any other Simics object (whether that is by accessing `connect`ed devices
or by leveraging the Simics API.) This includes even the device's own clock...
```

This makes it rather clear I'm not talking about, say, accessing `.val` of registers.

lwaern-intel: I used this phrasing. Take a look and approve or disapprove.

syssimics:
Verification [#13838283](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13838283): ✅ pass

mandolaerik (approved):

PR #326, opened 2024-10-01
Resolve symlinks before comparing

mandolaerik:


syssimics:
Verification [#13925183](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=13925183): ✅ pass

PR #329, opened 2024-10-21
Avoid dependency on liblzma

jhbaarnh:
Not all Pythons are built with xz compression.

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -915,8 +915,8 @@ class DumpInputFilesTestCase(CTestCase):
         super().test()
         dir = Path(self.scratchdir) / 'dumped'
         dir.mkdir()
-        with tarfile.open(Path(self.scratchdir) / f'T_{self.shortname}.tar.xz',
-                          'r:xz') as tf:
+        with tarfile.open(Path(self.scratchdir) / f'T_{self.shortname}.tar.bz2',
+                          'r:bz2') as tf:
             tf.extractall(dir)
         assert (dir.joinpath(*self.prefix)
                 / os.path.basename(self.filename)).is_file()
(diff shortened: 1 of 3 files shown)


test/tests.py
```diff
@@ -915,8 +915,8 @@ def test(self):
         super().test()
         dir = Path(self.scratchdir) / 'dumped'
         dir.mkdir()
-        with tarfile.open(Path(self.scratchdir) / f'T_{self.shortname}.tar.xz',
-                          'r:xz') as tf:
+        with tarfile.open(Path(self.scratchdir) / f'T_{self.shortname}.tar.gz',
+                          'r:gz') as tf:
```
mandolaerik, line 919: The corresponding `tarfile.open` that created the file is in `py/dml/dmlc.py`.

The xz usage is for an odd use case and it's not important that it works for all pythons, so from a usability perspective it's OK if this test breaks. However, if switching to a different compression scheme makes something easier then I'm OK with that. I'd prefer bz2 over gz, though.

jhbaarnh: I'm setting up our CI to run all our tests with some host Pythons, and then run into this. An alternative is to build all the Pythons used to link with liblzma, or to allow XFAILing based on which Python was used.

syssimics:
Verification [#14004354](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14004354): ❌ fail

test/tests.py
mandolaerik: The choice of format is unimportant; switching from xz to bz2/gz is fine if that makes anything easier in CI.

jhbaarnh: It appears that the Pythons in our lab that I was planning to use all have bz2 support, so I could use that instead if you prefer that.

mandolaerik:
Docs also need updating (`grep xz` reveals `README.md`)

mandolaerik (approved):

PR #336, opened 2024-11-07
Fix typo

mandolaerik:
- **Fix typo**

### DIFF
--- a/test/1.4/errors/T_EMETH.dml
+++ b/test/1.4/errors/T_EMETH.dml
@@ -18,6 +18,11 @@ template t {
     shared method downcast_with_fallback(u fallback) -> (u) default {
         return fallback;
     }
+    /// ERROR EMETH
+    independent method indep() default {
+    }
+    /// ERROR EMETH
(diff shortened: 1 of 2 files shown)


syssimics:
Verification [#14068908](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14068908): ✅ pass

test/1.4/errors/T_EMETH.dml
```diff
@@ -45,6 +50,28 @@ template v3 is u {
...
+}
+
+group g5 is t {
+    /// ERROR EMETH
+    independent startup method indep() {
+    }
+
+    /// ERROR EMETH
+    independent startup memoized method indep_ret() -> (int) {
+        return 0;
```
mandolaerik, line 71: Some code paths are not testable: trying to override `startup` with non-`startup` errs earlier because `startup` cannot be `default`.

lwaern-intel (approved):
Wonder how we haven't caught this earlier.

PR #338, opened 2024-11-19
Make pragmas official

lwaern-intel:
Long past time we do this. I'll need a test for the `COVERITY` pragma; I'll probably do something similar to how `linemarks` or `dead_dml_methods` work.

### DIFF
--- a/py/dml/output.py
+++ b/py/dml/output.py
@@ -187,12 +187,26 @@ def coverity_markers(markers, site=None):
     if dml.globals.coverity and site_with_loc:
         custom_markers = []
         filename = site.filename()
-        lineno = site.lineno
+        tgt_lineno = site.lineno
 
-        while (filename, lineno) in dml.globals.coverity_pragmas:
-            (lineno,
-             inline_markers) = dml.globals.coverity_pragmas[(filename, lineno)]
+        while (filename, tgt_lineno) in dml.globals.coverity_pragmas:
+            (start_lineno,
+             inline_markers) = dml.globals.coverity_pragmas[(filename,
+                                                             tgt_lineno)]
             custom_markers.extend(reversed(inline_markers))
+
+            # A minor HACK to handle the case of:
+            #   /*% COVERITY foo %*/ /*% COVERITY
+            #   bar %*/
+            #   some_statement
+            #
+            # Otherwise 'foo' won't be captured
+            if (start_lineno + 1 < tgt_lineno
+                and ((filename, start_lineno + 1)
+                     in dml.globals.coverity_pragmas)):
+                tgt_lineno = start_lineno + 1
+            else:
+                tgt_lineno = start_lineno
 
         custom_markers.reverse()
         markers = custom_markers + markers
--- a/test/1.4/errors/T_WPRAGMA.dml
+++ b/test/1.4/errors/T_WPRAGMA.dml
@@ -1,0 +1,34 @@
+/*
+  © 2024 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+dml 1.4;
+
+device test;
+
+// no warning
+/*% COVERITY irrelevant %*/
+
+/// WARNING WPRAGMA
+/*% MYVERYCOOLPRAGMA %*/
+
+/// WARNING WPRAGMA
+/*% 😀 %*/
+
+/// WARNING WPRAGMA
+/*%
+  MYVERYCOOLPRAGMA%*/
+
+// no warning (parses as a pragma)
+/*%
+  COVERITY
+  irrelevant%*/
+
+// no warning (doesn't parse as a pragma)
+/*% %*/
+
+// no warning (doesn't parse as a pragma)
+/*% MYVERYCOOLPRAGMA % */
+
+// no warning (doesn't parse as a pragma)
+//% MYVERYCOOLPRAGMA
--- a/test/1.4/pragmas/COVERITY.dml
+++ b/test/1.4/pragmas/COVERITY.dml
@@ -1,0 +1,68 @@
+/*
+  © 2024 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+
+dml 1.4;
+
+device test;
+
+/// COMPILE-ONLY
+/// DMLC-FLAG --coverity
+
+// This test is limited by the fact it relies on test annotations and line
+// directives from generated C. Notably, it has no means of properly testing
+// `--coverity` and the COVERITY pragma in the case of `--noline`.
+//
+// Annotation semantics:
+// * `SUPPRESSIONS-ABOVE [suppressions...]`
+//    One or more C lines are generated from the above DML line,
+//    and each of them receives exactly the suppressions specified
+// * `SUPPRESSIONS-ABOVE-LAX [suppressions...]`
+//    One or more C lines are generated from the above DML line,
+//    and *at least one* of them receives exactly the suppressions specified
+
+session struct { const int x; } s = {4};
+// SUPPRESSIONS-ABOVE [store_writes_const_field : FALSE]
+
+template t {
+    method m() {
+        /*% COVERITY foo %*/
+        if (s.x == 4) { after 1 s: m(); }
+        // SUPPRESSIONS-ABOVE [foo]
+    }
+}
+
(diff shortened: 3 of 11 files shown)


syssimics:
Verification [#14114156](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14114156): ✅ pass

lwaern-intel:
@mandolaerik Test added, please review.

syssimics:
Verification [#14144066](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14144066): ✅ pass

test/tests.py
```diff
@@ -1550,14 +1576,32 @@ def dml_lines_iter(self, c_file, c_lines_iter):
...
+            stripped = line.lstrip()
+            if stripped.startswith('#ifndef ') and stripped.endswith("_H"):
+                assert cpp_if_depth == base_if_depth == 0
+                base_if_depth = 1
+            if stripped.startswith('#if'):
+                cpp_if_depth += 1
+            elif stripped.startswith('#endif'):
+                cpp_if_depth -= 1
+                base_if_depth = min(base_if_depth, cpp_if_depth)
+                assert cpp_if_depth >= 0, 'nonsense: unclosed endif'
```
lwaern-intel, line 1589: Rather, `#endif closing nothing`

test/1.4/pragmas/COVERITY.dml
```diff
@@ -0,0 +1,58 @@
...
+dml 1.4;
+
+device test;
+
+/// COMPILE-ONLY
+/// DMLC-FLAG --coverity
+
+// This test is limited by the fact it relies on test annotations and line
+// directives from generated C. Notably, it has no means of properly testing
+// `--coverity` and the COVERITY pragma in the case of `--noline`.
```
lwaern-intel, line 15: Should add explanations of `SUPPRESSIONS-ABOVE(-LAX)` here:
```
// Annotation semantics:
// * `SUPPRESSIONS-ABOVE [suppressions...]`
//    One or more C lines are generated from the above DML line,
//    and each of them receives exactly the suppressions specified
// * `SUPPRESSIONS-ABOVE-LAX [suppressions...]`
//    One or more C lines are generated from the above DML line,
//    and *at least one* of them receives exactly the suppressions specified
```

test/1.4/pragmas/COVERITY.dml
```diff
@@ -0,0 +1,58 @@
...
+method init() {
+    assert true; // a sanity test that this gets no suppressions
+    // SUPPRESSIONS-ABOVE
+
+    /*% COVERITY irrelevant %*/
+
+    /*% COVERITY foo %*/
+    /*% COVERITY bar %*/ /*% COVERITY baz FALSE %*/
+    assert true;
+    // SUPPRESSIONS-ABOVE [foo] [bar] [baz:FALSE]
```
mandolaerik, line 41: Not sure if this should be `-LAX` or not (we happen to know that `assert` gives a single statement so they are equivalent). Which makes me ask:
```
/*% COVERITY foo %*/
assert true; assert true;
```
... will SUPPRESSIONS-ABOVE hold there? I don't think any other test answers that question.

lwaern-intel: Yes, `SUPPRESSIONS-ABOVE` will hold here, and that **does** get tested via line 23 (many statements/expressions on one line, some of which cause many C statements to be generated.)

lwaern-intel: It should **not** be `-LAX`. One of the guarantees of the `COVERITY` pragma is that, if specified, the annotation for it gets generated for _every_ C line generated from the pragmad DML line. The reason I use `-LAX` in other cases is because those involve _built-in_ analysis annotations that we add for certain DML statements, which _do not_ have to be generated for every C line corresponding to the line with the DML statement.

mandolaerik: good! then all is fine

test/1.4/errors/T_WPRAGMA.dml
```diff
@@ -0,0 +1,13 @@
...
+*/
+dml 1.4;
+
+device test;
+
+// no warning
+/*% COVERITY irrelevant %*/
+
+/// WARNING WPRAGMA
+/*% MYVERYCOOLPRAGMA %*/
```
mandolaerik, line 13: Some other corners:
```
/*% 😀 strange character in pragma tag %*/
/*%     COVERITY weird spacing%*/
/*% COVERITY forgot closing % */
//% COVERITY this isn't a pragma, right?
```
I don't have a strong opinion on which ones should be considered valid vs invalid vs discouraged, but they deserve tests since they now are language corners.

lwaern-intel: Done.

syssimics:
Verification [#14223817](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14223817): ✅ pass

test/1.4/errors/T_WPRAGMA.dml
```diff
@@ -0,0 +1,34 @@
...
+/*% 😀 %*/
+
+/// WARNING WPRAGMA
+/*%
+  MYVERYCOOLPRAGMA%*/
+
+// no warning (parses as a pragma)
+/*%
+  COVERITY
+  irrelevant%*/
```
mandolaerik, line 25: Interesting! It's somewhat unclear what line this is attributed to -- perhaps extend the foo/bar/baz case in pragmas/COVERITY with some newlines to illustrate this? e.g., the following should attach `foo` to `x`, right?
```
/*% COVERITY foo %*/
/*%
 COVERITY
  bar
   %/
x();
```

lwaern-intel: Yeah, but experimentation revealed a bug to how I'm handling this.

```
/*% COVERITY foo %*/
/*%
 COVERITY
  bar
   %/ /*% COVERITY baz %*/
 x();
 ```
 will work as expected, but 
 ```
/*% COVERITY foo %*/
/*% COVERITY bar %*/ /*%
 COVERITY
  baz
   %/
 x();
 ```
 will not; in the latter case it'll skip `bar`. I've added a hack to fix that. Honestly, I'm not sure about how _extreme_ cases we should support. We'll likely revisit this; if we were to add a warning pragma like we discussed, we'd likely to change and generalize the line-association mechanism.

**Comment by syssimics** [comment]
Verification [#14224681](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14224681): ✅ pass

**Comment by syssimics** [comment]
Verification [#14230944](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14230944): ✅ pass

File: py/dml/output.py [hunk 6]
```diff
@@ -187,13 +187,27 @@ def coverity_markers(markers, site=None):
...
+        while (filename, tgt_lineno) in dml.globals.coverity_pragmas:
+            (start_lineno,
+             inline_markers) = dml.globals.coverity_pragmas[(filename,
+                                                             tgt_lineno)]
             custom_markers.extend(reversed(inline_markers))
 
+            # A minor HACK to handle the case of:
+            #   /*% COVERITY foo %*/ /*% COVERITY
+            #   bar %*/
+            #   some_statement
```
mandolaerik, line 201: what if you add more weirdness, like
```
/*% COVERITY foo %*/ /*%

COVERITY
bar %*/ /*

COVERITY

baz
%*/
statement
```
? do you have all corners covered?

mandolaerik: Not to mention:
```
/*% COVERITY foo %*/ /*%
foo %*/
statement
```

lwaern-intel: The only corner you mention not covered is
```
/*% COVERITY foo %*/ /*%
foo %*/
statement
```
which results in `statement` receiving no annotation, as `foo` is not a `COVERITY` pragma. This might change once we change and generalize the line-association mechanism to work with pragmas in general. However, that'd only be true for **pragmas**; not comments, so in the case of: 
```
/*% COVERITY foo %*/ /*
foo */
statement
```
then `statement` would not receive an annotation, not today and not ever. At least, not in my vision of how pragmas would work.

mandolaerik (approved):

PR #340, opened 2024-12-02
Compatibility patch for innersource/applications.simulators.simics.simics-base#1660

lwaern-intel:
This is a cross-repo dependency

### DIFF
--- a/lib/1.4/dml-builtins.dml
+++ b/lib/1.4/dml-builtins.dml
@@ -13,7 +13,6 @@ header %{
     #include <simics/dmllib.h>
 %}
 
-import "simics/C.dml";
 import "simics/devs/io-memory.dml";
 import "simics/model-iface/bank-instrumentation.dml";
 import "simics/model-iface/int-register.dml";
@@ -36,17 +35,22 @@ constant unmapped_offset = 0xffff_ffff_ffff_ffff;
 
 param NULL auto;
 
-// Handy macros
+// Handy macros and types
+extern int strcmp(const char *__s1, const char *__s2);
+extern void *memset(void *__s, int __c, size_t __n);
+extern typedef ulong uintptr_t;
+
 extern char *MM_STRDUP(const char *s);
 extern void MM_FREE(void *m);
-
-extern char *_DML_describe_uint64_event(const char *base, uint64 data);
 
 extern size_t strlen(const char *s);
 extern char *strcpy(char *s, const char *s);
 extern void *memcpy(void *dest, const void *src, size_t n);
 
 extern const atom_t ATOM_LIST_END;
+
+
+extern char *_DML_describe_uint64_event(const char *base, uint64 data);
 
 extern typedef struct {
 } _callback_entry_t;
--- a/lib/1.2/simics-api.dml
+++ b/lib/1.2/simics-api.dml
@@ -117,5 +117,14 @@ import "simics-processor.dml";
 
 // C library stuff
 
-import "simics/C.dml";
 import "simics/util/bitcount.dml";
+
+extern int feof(FILE *__stream);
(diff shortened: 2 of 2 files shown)


lib/1.4/dml-builtins.dml
```diff
@@ -36,18 +35,23 @@ constant unmapped_offset = 0xffff_ffff_ffff_ffff;
 
 param NULL auto;
 
-// Handy macros
+// Handy macros and types
+extern int strcmp(const char *__s1, const char *__s2);
+extern void *memset(void *__s, int __c, size_t __n);
+extern typedef ulong uintptr_t;
```
mandolaerik, line 41: Redundant: ulong is provided by types.py (and incorrect IIRC: uintptr_t is `unsigned long long` on one of the platforms)

lwaern-intel: Wait.

_Oh crap._ We need to build-in `uintptr_t`.

Redundant is a **horribly** incorrect word for it, by the by!

lwaern-intel: Looking at dml-api, `unsigned long long` gets translated to `uint64`. So we don't need to have it built-in, we can just use `uint64`.

lwaern-intel: ... Or `uint64_t`? Hm. `uint64_t` is likely the better choice, because IIRC the same ulong vs. unsigned long long dichotomy applies to it as it does to `uintptr_t`. That is, they are identical on both hosts.

lwaern-intel: Picked `uint64_t`

lib/1.4/dml-builtins.dml
```diff
@@ -36,18 +35,23 @@ constant unmapped_offset = 0xffff_ffff_ffff_ffff;
 
 param NULL auto;
 
-// Handy macros
+// Handy macros and types
+extern int strcmp(const char *__s1, const char *__s2);
+extern void *memset(void *__s, int __c, size_t __n);
+extern typedef uint64_t uintptr_t;
```
mandolaerik, line 41: deserves a test,
local uintptr_t *y;
local uint64_t *x = y;

also you chose different defs for 1.2 and 1.4

mandolaerik (changes requested):

lib/1.4/dml-builtins.dml
lwaern-intel: > deserves a test

I honestly don't get that. _Why?_ It's like if you have a definition `param p = 5;` and then saying "you should have a test that p == 5`. It's pointless. It serves no purpose. If this definition were ever to be changed, it would be due to an intentional choice.

> also you chose different defs for 1.2 and 1.4

_That's_ true, and an oversight on my part. I'll address it.

mandolaerik: The suggested test is needed because it validates that C agrees with your assumption (uintptr_t and uint64_t are equivalent on all current platforms). It also serves as a reminder to fix the definition on any future platforms where it doesn't hold.

mandolaerik:
I guess you'll also want to add a release note for the change in the linked base PR (the DMLC rn might not be the perfect match, but probably the best match you'll find)

lib/1.4/dml-builtins.dml
lwaern-intel: > The suggested test is needed because it validates that C agrees with your assumption
Ah, right; the test is meaningful because it's an **extern** typedef. True.

lwaern-intel: Done

mandolaerik (approved):
OK assuming you have verified that the uintptr test passes also on windows.

mandolaerik:
Windows tests seem to pass

PR #342, opened 2025-01-15
Add compat feature `broken_unused_types`

mandolaerik:


mandolaerik:
builds: [V](https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=14324965) [DML](https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=14324967) [DML](https://teamcity01-ir.devtools.intel.com/viewQueued.html?itemId=14324966)

syssimics:
Verification [#14324965](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14324965): ✅ pass

RELEASENOTES-1.4.md
```diff
@@ -347,3 +347,10 @@
   behaviour (fixes SIMICS-22874).
 - `release 6 6356`
 - `release 7 7060`
+- `note 6` Addressed a bug where DMLC did not report errors when an `extern typedef` referenced a non-existing type. E.g., the following would not give an error if `never_used_t` was never used:
+  ```
+  extern typedef struct {
+      undefined_type_t member;
+  } never_used_t;
+  ```
+  The bugfix is opt-in, because an immediate bugfix would risk breaking existing builds; the error will only be reported when the flag `--no-compat=broken_unused_types` is passed to DMLC. This flag will be automatically enabled in Simics 8.
```
lwaern-intel, line 356: Have you done a VP pretest run on this? I find it slightly curious that you'd immediately turn to a compat feature.

test/tests.py
```diff
@@ -2052,6 +2052,7 @@ def test(self):
             print("dml %s;" % (self.dml_version,), file=f)
             print("device test;", file=f)
             print("/// COMPILE-ONLY", file=f)
+            print("/// DMLC-FLAG --no-compat=broken_unused_types", file=f)
```
lwaern-intel, line 2055: You need? Huh.... I thought with dml-api on 7 the ordering of types/symbols in the txt file happened to favor us (and its only with your experimentation that things got upset.)

RELEASENOTES-1.4.md
mandolaerik: No, I just did it as a compat feature defensively, to avoid having the woes of a VP pretest. The cost of preserving the compat feature is small in this case.

It would be tempting to just skip the compat feature; it feels pretty safe at first: if you encounter a broken unused type, then why not remove the type? However, consider the case when A.dml defines a type T1, and B.dml uses the type in a definition D without importing A.dml. This works in the case when X.dml imports both A.dml and B.dml before using D, but it currently also works if Y.dml imports B.dml without using D, and this PR would break that. The fix is that B should import A, but it's no longer a trivial fix (the error message only says "undefined type", and the path to A needs to be expressed on a form that is reachable from the -I flags used by both X and Y).

lwaern-intel:
Jenkins pretest using #344: https://cje-fm-owrp-prod06.devtools.intel.com/satg-simics-main/job/sre-vp/job/_pretest/191314/console . This does showcase backwards incompatibility! Thus, the compatibility feature is justified.

lwaern-intel (approved):

PR #347, opened 2025-01-23
Run Simics from the project

jhbaarnh:
This is required to make it find Python correctly.


syssimics:
Verification [#14341600](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14341600): ✅ pass

mandolaerik:
shouldn't a base package normally be set up so that its simics.bat works? If so, and your problem is that pkgs/ in the base repo isn't set up that way, then it's probably better to fix the repo.

jhbaarnh:
Not needed anymore.

PR #349, opened 2025-01-27
Softcode import

mandolaerik:


mandolaerik:
The same can be done in DLS, but the change there should wait some time to stay compatible with base packages without the change (imports are idempotent so the change to the lib is backward compatible with old DLSes).

syssimics:
Verification [#14348287](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14348287): ❌ fail

syssimics:
Verification [#14348326](http://teamcity01-ir.devtools.intel.com/viewLog.html?buildId=14348326): ✅ pass

JonatanWaern:
> The same can be done in DLS, but the change there should wait some time to stay compatible with base packages without the change (imports are idempotent so the change to the lib is backward compatible with old DLSes).

Noted.

lwaern-intel:
Hm. What made it special? Why wasn't it soft-coded to begin with?

mandolaerik:
Used to be this not so long ago:
```
     implicit_imports = [
         ast.import_(SimpleSite("%s:0" % (inputfilename,)), "dml-builtins.dml"),
         ast.import_(SimpleSite("%s:0" % (inputfilename,)),
                     simics_api_file(version, dml.globals.api_version))]
```
and further archeology points to this change 2013:
```
 def simics_api_file(dml_version, api_version):
-    if api_version.startswith('3') or api_version == '4.0':
-        return "simics/obsolete/limited-dml-simics-api.dml"
-    else:
-        return "simics/device-api.dml"
+    assert (dml_version == (1,2)
+            and not (api_version.startswith('3') or api_version == '4.0'))
+    return "simics/device-api.dml"
```

lwaern-intel (approved):
That's pretty funny

PR #352, opened 2025-02-26
doc: Correct read_bits/write_bits -> read_field/write_field

aeliasso:
The documentation was only partially corrected regarding these in 2021 in commit b6f8d3097d7ff779c7261a0d389a7831698964ae in Simics-Base (back when DMLC was part of that repository).

Also fix a documentation typo in dml-builtins.dml.

### DIFF
--- a/doc/1.4/language.md
+++ b/doc/1.4/language.md
@@ -573,9 +573,9 @@ categories of templates:
   the `read` or `write` template. The template leaves
     a simple method `read` (or `write`) abstract;
     custom behaviour is provided by overriding the method. There is
-    also a pair of templates `read_bits`
-    and `write_bits`, which similarly provide abstract
-    methods `read_bits` and `write_bits`. These
+    also a pair of templates `read_field`
+    and `write_field`, which similarly provide abstract
+    methods `read_field` and `write_field`. These
     functions have some extra parameters, making them less convenient
     to use, but they also offer some extra information about the
     access.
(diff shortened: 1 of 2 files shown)


RELEASENOTES-1.4.md
```diff
@@ -356,3 +356,4 @@
   The bugfix is opt-in, because an immediate bugfix would risk breaking existing builds; the error will only be reported when the flag `--no-compat=broken_unused_types` is passed to DMLC. This flag will be automatically enabled in Simics 8.
 - `release 7 7063`
 - `release 6 6362`
+- `note 6` Corrected read_bits/write_bits -> read_field/write_field in the reference manual.
```
lwaern-intel, line 359: A releasenote is redundant, but I personally don't find it redundant _enough_ to protest against it. I will however protest against the formatting. You should tick the references to the identifiers (both out of principle and because the `_` will incur italics otherwise).

Also, the ASCII arrow is slightly questionable (I've no idea how pretty it'll render.) You could go with a unicode arrow, or reformulate to not use the arrow:
> Corrected old references to `read_bits`/`write_bits` in the reference manual to instead reference their modern replacements `read_field`/`write_field`.


... or you could choose to drop the releasenote entirely.

lwaern-intel (changes requested):
These must've slipped by when we did our documentation revamp week a year or two back. Thanks!

lwaern-intel (approved):

aeliasso:
Merge it at your convenience as I lack permission to.

PR #354, opened 2025-02-28
The object structure generator prototype

AleksandrDanilovIntel:
Generates the internal object structure (objects, configuration attributes, input and output interfaces) in JSON format for later use in CLI commands and Blueprints.

attaching the output format
[test-cmn-700-dml.json](https://github.com/user-attachments/files/19030831/test-cmn-700-dml.json)

### DIFF
--- a/py/dml/dmlc.py
+++ b/py/dml/dmlc.py
@@ -721,6 +721,8 @@ def main(argv):
                                        dev, dml_version, outputbase + '.g')
                 logtime("g")
 
+        dml.info_backend.generate_json(dev, outputbase + '.json')
+
         if not logging.failure:
             # report WREF for broken unused parameters. But just ignore it
             # if there's already a hard error somewhere, because if the
--- a/py/dml/info_backend.py
+++ b/py/dml/info_backend.py
@@ -182,3 +182,103 @@ def generate(device, filename):
     classname = param_str_fixup(device, 'classname', '-')
     with XMLWriter(filename) as outfile:
         dev_info(outfile, device, classname)
+
+import json
+from dml.types import *
+from . import crep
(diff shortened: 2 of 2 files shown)


jhbaarnh (commented):
You could tag this with the JIRAs SIMICS-22377 and/or SIMICS-22497

I think it is a good idea, not sure which data we need to store though.

AleksandrDanilovIntel:
> You could tag this with the JIRAs SIMICS-22377 and/or SIMICS-22497
> 
> I think it is a good idea, not sure which data we need to store though.

We can create a well known list of parameters and give a user list as an option.

peternystrom-intel:
I see how this information can be useful but it is quite a gap from generating this file to having it accessible to a user or component/blueprint designer from cli. 
If we should go down this route the C++ Api will probably need something similar.

AleksandrDanilovIntel:
> I see how this information can be useful but it is quite a gap from generating this file to having it accessible to a user or component/blueprint designer from cli. If we should go down this route the C++ Api will probably need something similar.

The next step is to generate module_load.py files and add extra info commands each registered class.
Which will be automatically piked up by framework. And user will be able to see the object structure in DML terms.  
Henrik suggested adding cli_name to each object in the JSON file for easier matching. It also solves the problem of conversion of the DML types into the framework primitives.
It seems to me that the C++ types are different and it would be better to display it as is. And the cli_name approach works well with C++.

AleksandrDanilovIntel:
I moved all parameter names into a config file. Now it's much more flexible.
Adding one more example 
[test-subdev-dml.json](https://github.com/user-attachments/files/19679213/test-subdev-dml.json)

PR #359, opened 2025-04-23
Fix devnull path in 10d817ea

jhbaarnh:
It turns out the previous commit didn't work even though I thought I tested it (tested manually on the build machine this time).

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -962,9 +962,12 @@ class ErrorTest(CTestCase):
 all_tests.append(ErrorTest(["missing"], "xyz",
                            errors=[("xyz", 0, "ENOFILE")],
                            includepath=()))
-devnull_path = str(Path(os.devnull).resolve())
-if is_windows() and sys.version_info.minor > 10:
-    devnull_path += '\\'
+# On Windows NUL is not exactly a file in the same way.
+# Path resolve into an UNC path, i.e. with extra characters.
+if is_windows():
+    devnull_path = "NUL"
+else:
+    devnull_path = str(Path(os.devnull).resolve())
 all_tests.append(ErrorTest(
     ["empty"], os.devnull,
     errors=[(devnull_path, 1, "EDEVICE")],


syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/14619390445): ✅ success

test/tests.py
```diff
@@ -964,7 +964,7 @@ def test_flags(self):
                            includepath=()))
 devnull_path = str(Path(os.devnull).resolve())
 if is_windows() and sys.version_info.minor > 10:
-    devnull_path += '\\'
+    devnull_path = 'NUL'
```
mandolaerik, line 967: An improvement, but can still be simplified, maybe something like this?
```
# resolve() on windows appends \ in new python versions
devnull = os.devnull if is_windows() else str(Path(os.devnull).resolve())
```

jhbaarnh: No, just `os.devnull` will not be capitalized so it doesn't work either. I updated the comment to what I think is the problem, but the test is not easy to follow with all the layers.

A successful run with the current patch at least:
https://github.com/intel-innersource/applications.simulators.simics.infra.ci-tools/actions/runs/14641140339

test/tests.py
```diff
@@ -962,9 +962,11 @@ def test_flags(self):
 all_tests.append(ErrorTest(["missing"], "xyz",
                            errors=[("xyz", 0, "ENOFILE")],
                            includepath=()))
-devnull_path = str(Path(os.devnull).resolve())
+# Newer Pythons resolve into an UNC path, i.e. with extra characters.
 if is_windows() and sys.version_info.minor > 10:
```
mandolaerik, line 966: The `> 10` check is redundant?  `"NUL"` should be correct also with old pythons?

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/14643397485): ✅ success

mandolaerik (approved):
s/resolve/resolves/ maybe

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/14650033316): ✅ success

test/tests.py
jhbaarnh: @mandolaerik I thought so too, but apparently the check is needed. The verification build on this PR didn't run on Windows so this was not tested properly and now the base CI is broken...

PR #361, opened 2025-04-30
Provide {uint64,simple}_{time,cycle}_event in dml12-compatibility.dml

mandolaerik:


test/1.2/misc/dml14.dml
```diff
@@ -430,6 +430,50 @@ method memop_test() {
     assert SIM_get_mem_op_value_le(&memop) == 13;
 }
 
+event simple_time_ev is (simple_time_event) {
+    method event() {
+        if (this.posted()) {
+            local double next = this.next();
+            local double (*n)(conf_object_t*) = &this.next;
```
mandolaerik, line 437: This is a crude way to contrive a compile-time check that I didn't mix up double vs uint64 between timebases `seconds` and `cycles`. I wanted to do the same for `post`, but couldn't because it throws (and has to throw, because it calls `destroy` and is supposed to propagate exceptions from there; we could work around that with a bit of extra duplication, but I felt it's not worth it).

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/14757189786): ❌ failure

test/1.2/misc/dml14.dml
mandolaerik: Removed now, broke on DML 1.4 where `next` is a shared method

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/14766013419): ✅ success

aeliasso (commented):
Looks good when reading it through! I will test it on i210 + intel-e1000 and see if it converts the events they way I want.

lwaern-intel (approved):
I decided not to pursue my idea of a different heuristic. As established, yours is fine.

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/14880355197): ✅ success

PR #367, opened 2025-05-07
Extend port-dml documentation

mandolaerik:
- **Add release notes for recent port-dml-module improvements**
- **Extend port-dml documentation**


doc/1.4/port-dml.md
```diff
@@ -75,3 +75,99 @@ tags](changes-auto.html) and apply the change manually.
...
+incrementally, porting some devices at a time. It can then happen that some of
+your newly ported 1.4 files share common code with devices that are still in DML 1.2.
+
+It is not allowed to import a DML 1.2 file from a DML 1.4 device, but a DML 1.2 device may import a DML 1.4 file with some caveats. Thus, any code common between DML 1.2 and 1.4 must be ported to 1.4 before any device can be converted.
+There are two possible strategies for this: Either convert the common file in place,
+or duplicate it into separate 1.2 and 1.4 versions.
+
+### Keep a separate DML 1.4 copy
+
+After letting conversion tools convert `foo.dml` to DML 1.4, you can rename the
```
aeliasso, line 90: Shouldn't we promote the method of using 1.2/1.4 subdirectories instead, which avoids the need for a trampoline?

RELEASENOTES.md
```diff
@@ -232,3 +232,7 @@
   non-constant expressions (fixes SIMICS-13113).
 - `release 6 6363`
 - `release 7 7064`
+- `note 6` The `port-dml-module` script has been extended to support converting
```
aeliasso, line 235: Add this description of benefits of converting multiple modules also to port-dml.md.

doc/1.4/port-dml.md
mandolaerik: The 1.2/1.4 approach has some messy details; in particular, `import "foo/bar.dml"` will look in `1.2/foo/bar.dml` rather than `foo/1.2/bar.dml` that you might expect. This gets unmanageable in SSM's setup, where they add the repo root to the path and often express their imports as the equivalent of `import "src/devices/foo/bar.dml"`.

So the trampoline approach is more universal, and 1.2/1.4 is only marginally more lightweight, so I think it's easier to follow if we just document one method.

RELEASENOTES.md
mandolaerik: Fixed.

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/14902668128): ✅ success

aeliasso (approved):

mandolaerik:
Depends on https://github.com/intel-innersource/applications.simulators.simics.simics-base/pull/9628 , which provides the changes described by release notes.

PR #370, opened 2025-05-20
dmllib.h: Suppress Coverity false positives w.r.t. `va_start`

lwaern-intel:
*Yes*, Coverity is apparently **too stupid** to understand variadic arguments.


lwaern-intel:
Thinking about it -- this _might_ just spur Coverity to complain about _subsequent_ usages of `va`. Perhaps it's better to suppress the `var_decl` event on the variable declaration of `va` (the event that tells Coverity "this variable is uninitialized")

mandolaerik (approved):
Looks good if you have tested that it does silence stuff.

lwaern-intel (commented):
Yeah, I've tested it. Turns out my earlier concern was unwarranted -- these suppressions work.

PR #371, opened 2025-05-26
Free the map target when the object is destroyed

TSonono:
I tried to make changes to `src/tools/dmlc/test/1.4/lib/T_map_target_connect.py` to test this, but I couldn't find a good way to check if the map target had been free:d. Let me know if you think this is OK or if you have a good way of testing this.

### DIFF
--- a/RELEASENOTES-1.4.md
+++ b/RELEASENOTES-1.4.md
@@ -450,3 +450,5 @@
   The device object always instantiates this template.
 - `release 6 6464`
 - `release 7 7160`
+- `note 6` The `map_target` template from `utility.dml` now frees the internally
+ created map target upon device deletion.
--- a/lib/1.4/utility.dml
+++ b/lib/1.4/utility.dml
@@ -1272,6 +1272,13 @@ template map_target is (connect, _qname) {
         this.map_target = obj ? SIM_new_map_target(obj, NULL, NULL) : NULL;
     }
 
+    group _internal_destructor is destroy {
+        method destroy() {
+            SIM_free_map_target(parent.map_target);
+            parent.map_target = NULL;
+        }
+    }
+
     /**
        * `read(uint64 addr, uint64 size) -> (uint64) throws`
 


lib/1.4/utility.dml
```diff
@@ -1272,6 +1272,11 @@ template map_target is (connect, _qname) {
         this.map_target = obj ? SIM_new_map_target(obj, NULL, NULL) : NULL;
     }
 
+    method destroy() default {
+        SIM_free_map_target(this.map_target);
+        this.map_target = NULL;
```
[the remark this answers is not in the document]
TSonono: Made it an internal destructor

lib/1.4/utility.dml
```diff
@@ -1272,6 +1272,13 @@ template map_target is (connect, _qname) {
         this.map_target = obj ? SIM_new_map_target(obj, NULL, NULL) : NULL;
     }
 
+    group _internal_destructor is destroy {
+        method destroy() default {
```
lwaern-intel: No need to make this `default`.
At first, the thought struck me that `default` could provide an escape hatch for the scenario I hypothesized -- where some modeler needs to act on `map_target` before its destroyed -- but then I realized even without `default` there is an escape hatch anyway. You could declare _yet another_ `group` instantiating `destroy`, this time inside `_internal_destructor`. `destroy` semantics will guarantee the inner group's `destroy()` method will be called before that of `_internal_destructor`. As I find the scenario unlikely, and overriding `destroy()` of something named _internal destructor_ is only slightly prettier of a solution than needing to declare a nested group, I am content with there only being one ugly escape hatch rather than two.

lwaern-intel (changes requested):

RELEASENOTES-1.4.md
```diff
@@ -450,3 +450,5 @@
   The device object always instantiates this template.
 - `release 6 6464`
 - `release 7 7160`
+- `note 6` The `map_target` template from `utility.dml` now frees the internally
+ created map target when the object instantiating the template is deleted.
```
lwaern-intel: Nit:
> now frees the internally created map target upon device deletion.

_DML_ objects aren't really deleted as such.

lwaern-intel:
Oh no the label must have broken somewhere along the line and i dont know when. it's been forever since i actually tried to use it. Oh well, I still have verify-dml-pr.sh

lwaern-intel (approved):
Will merge once verification passes

syssimics:
[DML Verification : 7](https://github.com/intel-innersource/applications.simulators.simics.infra.ci-tools/actions/runs/29321002988): ❌ failure

syssimics:
[DML Verification : 6](https://github.com/intel-innersource/applications.simulators.simics.infra.ci-tools/actions/runs/29321000677): ❌ failure

lib/1.4/utility.dml
```diff
@@ -1272,6 +1272,13 @@ template map_target is (connect, _qname) {
         this.map_target = obj ? SIM_new_map_target(obj, NULL, NULL) : NULL;
     }
 
+    group _internal_destructor is destroy {
+        method destroy() {
+            SIM_free_map_target(this.map_target);
+            this.map_target = NULL;
```
lwaern-intel: Oh, duh, `this` needs to be `parent`. I'll fix that up.

TSonono: Right, I just moved the method my bad. I did not know how to run sanity tests in the dmlc repo. I tried `make test-dmlc`, but that didn't work. Do I need to work from within a simics base repo for that to work?

Edit: Ah, `./bin/mini-python test/scripts/testrun.py src/tools/dmlc/test/` from the repo worked

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/29321005300): ❌ failure

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/29321905040): ✅ success

lib/1.4/utility.dml
lwaern-intel: That, or test-run `t126_dmlc`, which is tiny indirection to `src/tools/dmlc/test`.

PR #373, opened 2025-06-05
Update changes-manual: unmapped fields and constant registers

aeliasso:
I haven't checked whether these would be caught by --strict-dml12. I guessed that they would not. Do you know off-hand if this is true?

### DIFF
--- a/doc/1.4/changes-manual.md
+++ b/doc/1.4/changes-manual.md
@@ -404,3 +404,7 @@ by `--strict-dml12`, and must be adjusted manually:
   The first message is logged on level 2 by `silent_unimpl`,
   and level 1 by all the other templates, while subsequent messages
   are logged on level 3 in all templates.
+
+* Registers that instantiate the `constant` template should use the `init_val`
+  parameter to declare their reset value, changed from the `value` parameter in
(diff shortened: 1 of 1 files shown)


doc/1.4/changes-manual.md
```diff
@@ -404,3 +404,9 @@ by `--strict-dml12`, and must be adjusted manually:
   The first message is logged on level 2 by `silent_unimpl`,
   and level 1 by all the other templates, while subsequent messages
   are logged on level 3 in all templates.
+
+* In DML 1.2 the template `unmapped` could be instantiated by fields. In DML 1.4
+  only registers can be `unmapped`.
```
mandolaerik, line 409: It makes no sense to instantiate the template in 1.2 (although it does not give an error). The template is defined like this in 1.2,
```
template unmapped {
    parameter offset = undefined;
}
```
and you are allowed to define a param "offset" anywhere, although it's seldom meaningful outside registers.

There's many similar details where 1.4 is more strict than 1.2 in some sense; I suggest that we either remove this clause or try to say something generic.

mandolaerik (approved):

PR #375, opened 2025-06-11
Plug memory leak

mandolaerik:


include/simics/dmllib.h
```diff
@@ -2286,6 +2286,15 @@ _DML_get_qname(_identity_t id, const _id_info_t *id_infos,
     return qname;
 }
 
+void
+_DML_free_qname_cache(qname_cache_t *cache)
+{
+        for (int i = 0; i < 4; i++) {
```
TSonono, line 2292: Can this magic 4 be a define?

lwaern-intel (commented):
There's a potential pitfall, here; model code can be called following `_destroy` due to immediate after (see `c_backend.generate_deinit`,) and that model code could call `_qname` and reallocate the buffers.
The surefire way to address that is to free the qname cache separately from `_destroy`. Say, introduce the method `_destroy_qname_cache` and then `codegen_inline_byname` it in `generate_deinit` after the final `_DML_execute_immediate_afters_now`. Or we could just be fine with that extremely unlikely leak.

lib/1.4/dml-builtins.dml
```diff
@@ -663,7 +663,10 @@ template device {
 
     method _init() { _rec_init(); }
     method _post_init() { _rec_post_init(); }
-    method _destroy() { _rec_destroy(); }
+    method _destroy() {
+        _rec_destroy();
+        _DML_free_qname_cache();
```
lwaern-intel, line 668: Missing its argument.

lwaern-intel (changes requested):

PR #376, opened 2025-06-18
Simplify language and avoid confusing linebreaks in provisional doc

aeliasso:
The Markdown compiler got confused by the line starting with :default and interpreted it as a definition list item. Remove such line breaks and simplify the language by splitting each bullet point into two cases, using code blocks instead of inline typewriter text.


mandolaerik (approved):

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/15731094305): ❌ failure

PR #377, opened 2025-07-01
dmllib.h: Add function to calculate `_qname` independently

lwaern-intel:
I've found myself wanting this multiple times while working on DDM. We're keeping this a purely internal change -- one way I'm limiting its exposure is by intentionally not `extern`ing the new function within dml-builtins.


mandolaerik (approved):
Makes sense. Having it the highest level of internal-ness ducks bikeshedding (we could even have extern:ed it, but thanks to idempotent externs we can do that later instead).

include/simics/dmllib.h
```diff
@@ -2256,21 +2256,21 @@ __qname(dml_qname_cache_t *cache, const char *fmt, ...)
...
     // In order to distinguish the device object from any other, its id_info
     // logname is always "dev", but we want its name when it comes to qname.
     if (strcmp(logname, "dev") == 0) {
-        return dev_name;
+        return strdup(dev_name);
     }
 
     if (info.dimensions == 0) {
-        return logname;
+        return strdup(logname);
```
mandolaerik, line 2273: This one is a significant cost increase for a common shortcut: the intent was to return early from `_DML_get_qname` when things are simple; now we'll instead alloc/dealloc redundantly, copy it once more, and also consume an element of the qname ring buffer. I suggest that we instead move these shortcuts to `_DML_get_qname`, where they help.

lwaern-intel: This would be a microoptimization. Are you really expecting it to matter?

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/16019401024): ❌ failure

include/simics/dmllib.h
mandolaerik: * That code path *is* already a micro-opt; you are preserving that micro-opt but eliminating most of the optimization from it.
* It's not only a micro-opt, it is also a defensive hack to eases the burden on the qname ring buffer hack: If you write
  ```
  sb_newf("%s %s %s %s %s", a.qname, b.qname, c.qname, d.qname, e.qname);
  ```
  then it will currently break only if *all* five objects are inside arrays. With your change it will unconditionally break.

PR #379, opened 2025-07-03
Provide template for an efficient private RAM

mandolaerik:
I did this mostly as an experiment, needs more testing and discussions before it can be added for real.

The idea is to allow reasonably easy access to a private RAM with good performance.

### DIFF
--- a/lib/1.4/utility.dml
+++ b/lib/1.4/utility.dml
@@ -9,7 +9,9 @@ import "simics/devs/signal.dml";
 import "simics/devs/ram.dml";
 import "simics/devs/memory-space.dml";
 import "simics/devs/translator.dml";
+import "simics/model-iface/direct-memory.dml";
 import "simics/model-iface/transaction.dml";
+import "simics/simulator/conf-object.dml";
 
 template _reg_or_field {
     param is_register : bool;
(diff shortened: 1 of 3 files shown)


lib/1.4/utility.dml
```diff
@@ -1393,6 +1395,141 @@ template map_target is (connect, _qname) {
...
+
+   Provides the following methods:
+ */
+typedef struct {
+    direct_memory_handle_t handle;
+    uint8 *data;
+} _ram_page_t;
+template ram {
+    param size;
+    connect ram is (init_as_subobj, destroy) {
```
mandolaerik, line 1419: A better design is probably to create a template for a private RAM connect, without  DMI support, and then create a separate template which adds DMI magic to all private RAMs within a bank/subdev/port

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/16060328123): ❌ failure

lib/1.4/utility.dml
mandolaerik: Oh, this adds some pain: the use of get_user_data requires that you have one separate direct_memory_update ports for each RAM; if we want a shared impl, then we instead need to keep a private hashtable from handles to pages.

PR #381, opened 2025-08-21
Fix casts to template types not checking for non-values

lwaern-intel:
As encountered by @gwikland. This would previously provoke an ICE. Jonatan, you're lone reviewer unless Erik happens to see this on vacation.

### DIFF
--- a/py/dml/ctree.py
+++ b/py/dml/ctree.py
@@ -3733,6 +3733,8 @@ class TraitObjectCast(Expression):
                 + '(_traitref_t) {_object_vtables[__id.id], __id};})')
 
 def mkTraitUpcast(site, sub, parent):
+    if isinstance(sub, NonValue):
+        raise sub.exc()
     typ = safe_realtype(sub.ctype())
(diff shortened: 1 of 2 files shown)


py/dml/ctree.py
```diff
@@ -3733,6 +3733,8 @@ def read(self):
                 + '(_traitref_t) {_object_vtables[__id.id], __id};})')
 
 def mkTraitUpcast(site, sub, parent):
+    if isinstance(sub, NonValue):
+        raise sub.exc()
```
lwaern-intel, line 3737: I could also place this check in `mkCast`, but saw no reason to prefer it there as opposed to here.

JonatanWaern: This got me thinking a little about what sort of errors .exc() actually propagates up and whether they make sense here. As far as I can tell, it should be fine (though I am not entirely sure which nonvalues _can_ even appear here)

JonatanWaern (approved):
Seems fine

PR #382, opened 2025-08-27
Run Simics from project

jhbaarnh:
This makes sure that Python paths etc are set correctly.


syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/17269255167): ✅ success

PR #384, opened 2025-08-27
Describe explicit_param_decls in section about params

aeliasso:


doc/1.4/language.md
```diff
@@ -1332,6 +1332,14 @@ Also note that a parameter declaration without definition is redundant if a
 [typed parameter declaration](#typed-parameters) for that parameter already
 exists, as that already enforces that the parameter must be defined.
 
+> [!NOTE]
+> The provisional feature *explicit_param_decls* enforces that parameters must be declared before they are assigned.
+> This distinguishes between the intent to declare a new parameter, and the intent to override an existing parameter.
+> This distinction allows DML to capture misspelled parameter overrides as compile errors.
+>
+> The provisional feature also introduces a new syntax for combined declaration and definition: `:=`.
```
mandolaerik, line 1340: Some concrete code examples would probably make it easier to catch the eye. E.g., you might just explain it by example, comparing `param p = 4;` to `param p := 4;`, and what the difference conveys. People will probably come here because they spotted `:=` in code and wonder what it is about.

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/17268830856): ✅ success

doc/1.4/language.md
```diff
@@ -1332,6 +1332,14 @@ Also note that a parameter declaration without definition is redundant if a
...
+> For more information see [Provisional language features](provisional-auto.html).
```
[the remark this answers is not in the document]
lwaern-intel: honestly though we should consider autogenerating those anchors

doc/1.4/language.md
aeliasso: Added some examples that hopefully demonstrate the improved coding error detection

doc/1.4/language.md
```diff
@@ -1221,14 +1221,121 @@ template get_qname {
...
+param NAME := value;
+param :default value;
+```
+
+`explicit_param_decls` enforces that parameters are declared before they are assigned, or that the combined syntax is used.
+This distinguishes between the intent to declare a new parameter, and the intent to override an existing parameter.
+This distinction allows DML to capture misspelled parameter overrides as compile errors.
 
+DMLC signals an error if the combined declaration and definition syntax is used to override an existing parameter.
+This guards against unintentional reuse of a parameter name. An example:
```
mandolaerik, line 1295: Consider if it gets better by mentioning the motivation for the provisional earlier, e.g. right before or after you introduce it.

doc/1.4/language.md
```diff
@@ -1221,14 +1221,121 @@ template get_qname {
 }
 ```
 
-### Typed Parameters
-A *typed parameter declaration* is a parameter declaration form that may only
-appear within templates:
+## Parameters detailed
```
mandolaerik, line 1224: Poor title. I don't have a great alternative.

lwaern-intel: Here's one of the places where we get seriously struck by the fact that "parameter declarations" are an overloaded term: it may either refer to "the `param` object statement" or "a parameter declared to exist". I think it's important for us to clear this matter up -- it's a horrible status quo.
I want to find a new, unambiguous term to refer to the `param` object statement in general (which was the intended meaning behind "parameter declaration" in this section.) If we do, then we'll find the title for this section.

"The `param` object statement" or "Parameter object statements" are technically correct, but incredibly awkward. "Parameter members" is also an alternative, but would likely be too confusing.

We could also try to side-step the issue altogether by saying "Parameter declarations and definitions". I'm unsure about that option, but its the one I'm actually currently leaning most heavily towards. Whenever we do need a *singular* term to describe `param` collectively, we could turn to "the `param` object statement." An example would be error messages usage of "parameter declaration" -- but honestly, I'd rather we improve them to contextually say "declaration" or "definition" depending on the context.

lwaern-intel: > I considered merging the subsection Parameters under The Object Model 

The Object Model section is kinda meant like a secondary introduction to the most crucial elements of DML (with the first introduction being in the introduction). This is maybe somewhat questionable, but for the time being I definitely want to keep the two sections separate.

lwaern-intel: >  what about "param statements"?

I dislike that. Unqualified "statement" means method statement by precedence and by intuition. It'd invite confusion.

> The problem remains that the term "parameter declaration" still can be misunderstood as "param statement".

*True*, but the problem is minimized if we refrain from ever referring to an arbitrary `param` object statement as a "param declaration". This means refactoring our documentation. Within DMLC python source itself, a grep after `aram?.*eclar` reveals the only questionable usages to be the docstring for `explicit_param_decls` and `PCONSTANT`. The error message for `EAUTOPARAM` should also be rewritten to read "bad definition/declaration of...".
Our docs proper however use the term "parameter declaration" all over the place in a way that's improper if we are to differentiate between them and definitions, and will need a deeper look.

lwaern-intel: My personal preference is apply the clear separation between "param declaration" and "param definition" like your rewritten docs now do to all our other existing documentation. If we do that, then there is no problem at all in having the title of this section be "Parameter declarations and definitions".

That doesn't have to be part of this PR, if you find that too tedious to do or @mandolaerik disagrees with me for some reason; then you can stick with "Parameters detailed". However, I would note that with _only_ this text rewritten, there is a terrible inconsistency in it distinguishes between declarations and definitions, whereas other parts of the docs still say "parameter declaration" as a synonym to "parameter object statement". The fact that this PR introduces this inconsistency (or at least makes it way more prominent) is something of an regression -- even though I most certainly prefer the new terminology. I don't think the problem is big enough for me to block merging, but I can't help but squirm a little about it.

MagnusIntel:
@aeliasso volunteered to help improve the DML documentation and I think a week of debate is more than enough. Can we get this done now @lwaern-intel and @mandolaerik ?

lwaern-intel (approved):
It's fine from my view.

doc/1.4/language.md
lwaern-intel: Yeah I disagree with Erik on this front. There's nothing wrong with presenting the motivation after describing the feature.

PR #388, opened 2025-09-27
Clean up imports

jhbaarnh:


syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/18059500365): ✅ success

test/1.4/lib/T_bad_subobj_connect.py
```diff
@@ -1,12 +1,12 @@
...
 # SPDX-License-Identifier: MPL-2.0
 
-from simics_common import CriticalErrors
+import simics
 import stest
 conf.sim.stop_on_error = False
 try:
-    SIM_load_module(f'dml-test-bad_subobj_connect')
-except CriticalErrors as e:
+    simics.SIM_load_module(f'dml-test-bad_subobj_connect')
```
lwaern-intel, line 8: There are other tests (`1.4/methods/T_startup`, `1.4/methods/T_startup_memoized` that may call `SIM_load_module` without needing to explicitly import `simics`. Does this one usage really need to be qualified? Or was this change an accident? If the latter, then please remove the qualification.

lwaern-intel (changes requested):
Looks good except for one change I suspect is accidental. Approved if it is intentional.

mandolaerik (approved):

test/1.4/lib/T_bad_subobj_connect.py
jhbaarnh: It was only in this test that I got an error because of CriticalErrors having moved, so I took the opportunity to clean up the import in this file only. We could probably go through the other tests and fix imports there too if you want, but preferably in a separate PR. 

SIM are indeed in the global namespace anyway, but this is not desirable in this context and I want to get away from that: people use linters such as pylance in VSCode and they will flag these symbols as unknown since the linter doesn't know that the Simics will add it to the global namespace at runtime. But if you import simics, and use it as simics.SIM_load_module, it will be ok since we distribute the stub file api.pyi that contain Python declarations of all symbols in the C API.

PR #430, opened 2026-03-30
Fix up `lexpos` for symbols from empty productions -- SIMICS-23466

lwaern-intel:
@mandolaerik any suggestions on how to best test this? Y'know, besides `porting` and #431.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,7 +133,7 @@ export CC
 export SIMICS_BASE
 
 RUN_PY_UNIT_TEST := $(PYTHON) $(DMLC_DIR)/run_unit_tests.py
-%-pyunit-tested: $(DMLC_DIR)/py/dml/%.py $(OUT_PYFILES)
+%-pyunit-tested: $(DMLC_DIR)/py/dml/%.py $(OUT_PYFILES) $(OUT_GEN_PYFILES)
 	$(info Testing $<)
 	$(RUN_PY_UNIT_TEST) $(SIMICS_PROJECT)/$(HOST_TYPE) $<
 	touch $@
--- a/py/dml/dmlparse.py
+++ b/py/dml/dmlparse.py
@@ -152,6 +152,59 @@ def parse_bitorder(t, syn):
 production_rules_dml12 = {}
 production_rules_dml14 = {}
 
+def mk_get_token(lexer):
+    '''Produces a wrapper over the lexer's token retrieval function, which must
+    be used as 'tokenfunc' when calling parse; see fixup_emptyprod_lexpos'''
+    lexer.latest_token = None
+    def get_token():
+        tok = lexer.token()
+        lexer.latest_token = tok
+        return tok
+    return get_token
+
+def fixup_emptyprod_lexpos(t):
+    '''
+    Must be called in every empty production rule to guarantee a sane lexpos
+    for it, and subsequently to guarantee sane sites for any production rules
+    that may rely on it.
+
+    Ply has *exceedingly* stupid behavior in that the lexpos it assigns to a
+    symbol yielded by an empty production is just wherever the lexer
+    happens to be at the time. Typically, that means it gets a lexpos
+    *just past* whatever token *follows* the produced symbol (the look-ahead),
+    which is absolute nonsense. This has messed up our sites since forever;
+    see SIMICS-23466.
+
+    The *actual* desired lexpos for a symbol from an empty production is at the
+    start of the next token (or EOF); i.e., the look-ahead.
+    That token will typically (always?) be the most recent one that got lexed.
+    (Ply does get weird once it encounters a parse error, arranging a
+    "lookahead stack," but the error() function we provide Ply is violent
+    enough to terminate parsing completely, so we have the luxury of not
+    needing to care.)
+
+    Hence, we hijack the function the parser uses to acquire tokens
+    (conveniently, Ply actually has an option for that) to stash the most
+    recently lexed token, and then we can use that token's lexpos to fix up the
+    lexpos of every empty production, fixing the problem at the source.
+
+    Ideally instead of peppering fixup_emptyprod_lexpos(t) all over the place
+    we'd have a check in the prod decorators such as to automatically augment
+    it onto any production rule that could yield empty. Unfortunately, that
+    breaks Ply completely. Y'see, Ply does not only rely on `.__doc__` and
+    `.__name__` of functions... but also `.__code__.co_firstlineno`. Brilliant.
+    '''
+    latest_tok = t.lexer.latest_token
+    if latest_tok is not None:
+        # HACK set_lexpos not available in Ply 3.4, so internals to the rescue.
+        # No matter; since Ply is unmaintained I can *definitively say* that
+        # this code is guaranteed to be compatible with all future versions of
+        # Ply!
+        # Haha...
+        sym = t.slice[0]
+        sym.lexpos = latest_tok.lexpos
+        sym.lineno = latest_tok.lineno
+
 def prod_dml12(f):
     '''Decorator for functions that should be used as production rules
     in the DML 1.2 grammar'''
@@ -189,6 +242,7 @@ def maybe_provisional_yes(t):
 @prod
 def maybe_provisional_no(t):
     'maybe_provisional : '
+    fixup_emptyprod_lexpos(t)
 
 
 @prod
@@ -200,11 +254,13 @@ def maybe_device_yes(t):
 @prod
 def maybe_device_no(t):
     'maybe_device : '
+    fixup_emptyprod_lexpos(t)
     t[0] = None
 
 @prod
 def maybe_bitorder_no(t):
     'maybe_bitorder : '
+    fixup_emptyprod_lexpos(t)
     t.parser.file_info.bitorder = 'le'
 
 @prod
@@ -221,6 +277,7 @@ def device_statements(t):
 @prod
 def device_statements_empty(t):
     'device_statements : '
+    fixup_emptyprod_lexpos(t)
     t[0] = []
 
 @prod_dml12
@@ -282,6 +339,7 @@ def toplevel_if(t):
 @prod_dml14
 def toplevel_else_no(t):
     '''toplevel_else :'''
+    fixup_emptyprod_lexpos(t)
     t[0] = []
 
 @prod_dml14
@@ -304,6 +362,7 @@ def object_anonymous_bank(t):
 @prod
 def array_list_empty(t):
     'array_list : '
+    fixup_emptyprod_lexpos(t)
     t[0] = []
 
 @prod
@@ -325,6 +384,7 @@ def bitrangespec(t):
 @prod
 def bitrangespec_empty(t):
     'bitrangespec :'
+    fixup_emptyprod_lexpos(t)
     t[0] = []
 
 @prod_dml14
@@ -377,6 +437,7 @@ def object_field_1(t):
 @prod_dml12
 def field_array_size_no(t):
     'fieldarraysize : '
+    fixup_emptyprod_lexpos(t)
     t[0] = []
 
 @prod_dml12
@@ -496,6 +557,7 @@ def maybe_extern_yes(t):
 @prod_dml12
 def maybe_extern_no(t):
     '''maybe_extern :'''
+    fixup_emptyprod_lexpos(t)
     t[0] = False
 
 @prod
@@ -506,6 +568,7 @@ def maybe_default_yes(t):
 @prod
 def maybe_default_no(t):
     '''maybe_default :'''
+    fixup_emptyprod_lexpos(t)
     t[0] = False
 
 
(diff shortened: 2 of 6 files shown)


py/dml/dmlparse.py
```diff
@@ -540,8 +603,7 @@ def object_method_noinparams(t):
     name = t[3]
     (inp, outp, throws) = ([], t[4], True)
     if logging.show_porting:
-        i = min([4, 5, 6], key=lambda i: t.lexpos(i))
```
lwaern-intel, line 543: I _am_ a bit peeved that you knew about this but never created an issue about it. Not overly, though. My anger remains largely directed towards Ply.

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/23771640258): ✅ success

py/dml/dmlparse.py
mandolaerik: My idea was that someday we'll move on to a better parser, and then the problem will vanish for free; the accumulated cost of this shortcoming small until then. But the cost of maintaining your fix is probably smaller so that's a better choice.

py/dml/dmlparse.py
```diff
@@ -152,6 +152,59 @@ def parse_bitorder(t, syn):
...
+    Hence, we hijack the function the parser uses to acquire tokens
+    (conveniently, Ply actually has an option for that) to stash the most
+    recently lexed token, and then we can use that token's lexpos to fix up the
+    lexpos of every empty production, fixing the problem at the source.
+
+    Ideally instead of peppering fixup_emptyprod_lexpos(t) all over the place
+    we'd have a check in the prod decorators such as to automatically augment
+    it onto any production rule that could yield empty. Unfortunately, that
+    breaks Ply completely. Y'see, Ply does not only rely on `.__doc__` and
+    `.__name__` of functions... but also `.__code__.co_firstlineno`. Brilliant.
```
mandolaerik, line 195: Maybe we can create a good-enough test to ensure we have the fixup in place where it's needed? With such test in place, I would conisder the maintenance cost of peppering calls as low; it's reduced to minor noise that you don't have to give another thought until the test heuristics fails.

Whatever narrow-minded crude hack is fine, e.g. scan the source code of each production function for a fixup function call, and check that this is exactly the set of functions with `[:|] *\n` in their `__doc__`.

py/dml/dmlparse.py
```diff
@@ -152,6 +152,59 @@ def parse_bitorder(t, syn):
...
+    latest_tok = t.lexer.latest_token
+    if latest_tok is not None:
+        # HACK set_lexpos not available in Ply 3.4, so internals to the rescue.
+        # No matter; since Ply is unmaintained I can *definitively say* that
+        # this code is guaranteed to be compatible with all future versions of
+        # Ply!
+        # Haha...
+        sym = t.slice[0]
+        sym.lexpos = latest_tok.lexpos
+        sym.lineno = latest_tok.lineno
```
mandolaerik, line 206: From a performance perspective this looks pretty cheap, but it would be satisfying to see some sort of metrics that parse performance is unaffected. E.g., we could count the number of fixup calls needed to parse a fairly large device; something like
```
    global count
    if count % 10000 == 0: dbg(count)
    count += 1
```
is probably suffiicient to show convincingly that this is not anywhere near a problem.

py/dml/dmlparse.py
lwaern-intel: i feel compelled to reiterate that this bug has served as my personal sleep paralysis demon for years. It's a bit ironic that when you figured it out way-back-when you went "ehh, who cares." Guess you have your answer!

py/dml/dmlparse.py
lwaern-intel: I mbench:d parsing 10000 minimal method declarations `method m() {}`, on #431 vs `main`. That means the parsing times will not only involve the empty production fixups (of which there will be 5 per method), there's also the top-level `stray_is_check`. And the `len(t)` check in `method_qualifiers()` I guess.

Over 10 runs, parsing times on #431 had a mean of 0.8618 s, vs. main's 0.8323 s; a 3.5% regression. Mind you, that's parsing times _alone_; over the entire DMLC run that drops to a 1.5% regression and even _that_ figure is unrepresentative since all the methods are dead, meaning any work but parsing is minimal. I'm not even bothering to include the GCC times in this, this is as pathological as can be.

lwaern-intel: Oh, actually, my `mbench` also involved an `utility.dml` import, oops. Without it, the numbers are instead a mean of 0.1317s for #431, and 0.1286 for main, so a 2.4% regression; and over DMLC as a whole... a mean of 0.5881432 s vs 0.5864636 s, so a 0.2% regression.

Huh. That is... weird. Actually, I don't get it at all; `utility.dml` gets `.dmlast`:d, doesn't it? Why does the parsing regression get reflected more heavily _with_ it imported?

mandolaerik: good, thats more than enough

lwaern-intel: Disregard, I messed up my bench without the `utility.dml` import. Redoing it properly show a parsing time of 0.8617 vs. 0.822; a 4.8% regression; and over DMLC as a whole a 2.5% regression. The parsing time regression is more pronounced here as loading utility.dml's dmlast is part of the parsing stage, so that added a constant overhead to the all figures involving the import. So _these_ numbers are the ones that represent the most pathological case.

py/dml/dmlparse.py
lwaern-intel: > scan the source code of each production function for a fixup function call

I would say "**bleh,** that feels like trouble" but I just discovered `.__code__.co_names` which makes this trivial. (Of course accessing `.__code__` is really dirty, but Ply also does it so we're in good company.) So I'll do it.

Anyway, my question in the OP concerns testing the fact that we _are_ getting good lexpos:s, now. Right now it's only incidentally tested via `porting` and #431. We could add a `_test.py` which invokes the parser for, say, a `method` declaration and checks its site makes sense, but how far do we take it?

py/dml/dmlparse.py
mandolaerik: Another way to measure: I compiled some random big device I found (test-cmn-700), total number of fixups there is ~30000. Separately I clocked `for _ in range(1000000): fixup()` and got 0.12 microseconds per call. So total slowdown is a few milliseconds on a big real device.

Interesting in the sense that we could have coped with a magnitudes heavier workaround if that was needed.

py/dml/dmlparse.py
mandolaerik: That sounds like a sufficient test: If we check that it works correctly for one empty production rule, and also check that the workaround is present wherever it's needed, then it's hard to break the test other than deliberately.

lwaern-intel: Done and done.

py/dml/dmlparse.py
```diff
@@ -152,6 +152,59 @@ def parse_bitorder(t, syn):
...
 production_rules_dml14 = {}
 
+def mk_get_token(lexer):
+    '''Produces a wrapper over the lexer's token retrieval function, which must
+    be used as 'tokenfunc' when calling parse; see fixup_emptyprod_lexpos'''
+    lexer.last_token = None
+    def get_token():
+        tok = lexer.token()
+        lexer.latest_token = tok
+        return tok
```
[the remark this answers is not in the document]
lwaern-intel: buh, how embarassing.

py/dml/dmlparse.py
```diff
@@ -152,6 +152,59 @@ def parse_bitorder(t, syn):
...
+    breaks Ply completely. Y'see, Ply does not only rely on `.__doc__` and
+    `.__name__` of functions... but also `.__code__.co_firstlineno`. Brilliant.
+    '''
+    latest_tok = t.lexer.latest_token
+    if latest_tok is not None:
+        # HACK set_lexpos not available in Ply 3.4, so internals to the rescue.
+        # No matter; since Ply is unmaintained I can *definitively say* that
+        # this code is guaranteed to be compatible with all future versions of
+        # Ply!
+        # Haha...
```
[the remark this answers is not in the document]
lwaern-intel: No. You both misunderstood the meaning and also have no sense of humor.
Can you even call yourself a programmer if you *don't* make sad jokes to hide your misery?

py/dml/dmlparse_test.py
```diff
@@ -0,0 +1,45 @@
...
+            and 'fixup_emptyprod_lexpos' not in rule.__code__.co_names):
+            bad_emptyprod_rules.add(rule)
+
+if bad_emptyprod_rules:
+    sys.stderr.write(
+        "empty production rule(s) without call to 'fixup_emptyprod_lexpos':\n")
+    for rule in bad_emptyprod_rules:
+        sys.stderr.write(f'    {rule.__name__}\n')
+
+    sys.exit(1)
```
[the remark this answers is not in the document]
lwaern-intel: > That will abort unittest discovery/execution and can produce confusing failures

Nobody cares, and no. This check isn't a true unit test, it's *enforcement.* It's to guarantee building DMLC fails and the developer immediately notice if the check isn't met. The only reason it lives here instead of `dmlparse.py` is cause it's too heavy of a check to run every time dmlc is used.

py/dml/dmlparse_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+def parse(contents):
+    file_info = logging.FileInfo(
+        '<unit test>', (1, 4), content_lines=contents.splitlines(
+            keepends=True))
+    ast = toplevel.parse(contents, file_info, file_info, (1, 4))
```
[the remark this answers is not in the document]
lwaern-intel: oh, oops.

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/23809331235): ✅ success

lwaern-intel:
yet again, Copilot shows its greatest strength to be a stupid mistake checker while also showing its lack of any deeper understanding.

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/23809638578): ✅ success

py/dml/dmlparse_test.py
```diff
@@ -0,0 +1,45 @@
...
+def parse(contents):
+    file_info = logging.FileInfo(
+        '<unit test>', (1, 4), content_lines=contents.splitlines(
+            keepends=True))
+    ast = toplevel.parse(contents, file_info, file_info.name, (1, 4))
+    assert ast.kind == 'dml', ast.kind
+    return ast
+
+class test_emptyprod_based_sites(unittest.TestCase):
+    def test(self):
```
mandolaerik, line 43: this test will look arbitrary and strange to a newcomer unless you mention that it tests that the fixup function does its thing

py/dml/dmlparse_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+bad_emptyprod_rules = set()
+for rules in (dmlparse.production_rules_dml12,
+              dmlparse.production_rules_dml14):
+    for (name, rule) in rules.items():
+        if (name != 'p_error' and empty_prod_re.search(rule.__doc__)
+            and 'fixup_emptyprod_lexpos' not in rule.__code__.co_names):
+            bad_emptyprod_rules.add(rule)
+
+if bad_emptyprod_rules:
```
mandolaerik: somewhat overengineered, it is sufficient to crash hard on the first error. but that is ok i guess

py/dml/dmlparse_test.py
lwaern-intel: You don't think the name is sufficient? I'd figure searching "emptyprod" within dmlparse or even the test itself would immediately tell you what it's about.

py/dml/dmlparse_test.py
lwaern-intel: It was once even more overengineered ;)

py/dml/dmlparse_test.py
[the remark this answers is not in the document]
mandolaerik: i agree with the ai that it is an inconsistrncy, tests normally fail through test failures so exit needlessly raises my eyebrows. i also agree with Love that nobody cares, including me.

py/dml/dmlparse_test.py
mandolaerik: I wrote it instead as another commit, easier than arguing about it.

I realized that the test is somewhat brittle (if we somehow rewrite method and object without using empty production rules, then it will no longer cover anything). But whatever, let's ignore that.

Makefile
```diff
@@ -133,7 +133,7 @@ export CC
 export SIMICS_BASE
 
 RUN_PY_UNIT_TEST := $(PYTHON) $(DMLC_DIR)/run_unit_tests.py
-%-pyunit-tested: $(DMLC_DIR)/py/dml/%.py $(OUT_PYFILES)
+%-pyunit-tested: $(DMLC_DIR)/py/dml/%.py $(OUT_PYFILES) $(OUT_GEN_PYFILES)
```
lwaern-intel: Explain this part cause this I don't get.

mandolaerik: ply's parse tables are built in a caching fashion: if a parse table is pre-built then it will be used, otherwise a new one will be built (and discarded). This change ensures that we wait for the pre-built table.

mandolaerik:
some suggested fixes here: https://github.com/lwaern-intel/device-modeling-language/pull/3

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/24000040534): ✅ success

mandolaerik (approved):

PR #431, opened 2026-03-30
Add warning for stray standalone is statements

lwaern-intel:
Depends on #430; only the latter two commits shown in this PR are relevant for review.

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -271,3 +271,14 @@
   Simics Base version 7.78.
 - `release 7 7150`
 - `release 6 6460`
+- `note 6` DMLC now detects and warns about any standalone `is` that looks
+  like it was instead intended for an object declared immediately before it,
+  such as the following:
+  ```
+  // Stray semicolons cause each 'is' to be standalone, affecting the
+  // containing register rather than the fields
+  field f1 @ [15:0]; is read_only;
+
+  field f2 @ [31:16];
+      is write_only;
+  ```
--- a/py/dml/dmlparse.py
+++ b/py/dml/dmlparse.py
@@ -1049,8 +1049,8 @@ def stray_is_check(body):
     for (i, stmt) in enumerate(body):
         if stmt.kind == 'object':
             (_, obj_type, _, _, obj_body) = stmt.args
-            rough_end_site = (obj_body and obj_body[-1].site) or stmt.site
-            j = i+1
+            rough_end_site = obj_body[-1].site if obj_body else stmt.site
+            j = i + 1
             while j < len(body):
                 other_stmt = body[j]
                 other_site = other_stmt.site
--- a/py/dml/dmlparse.py
+++ b/py/dml/dmlparse.py
@@ -1055,12 +1055,9 @@ def stray_is_check(body):
                 other_stmt = body[j]
                 other_site = other_stmt.site
 
-                if other_site.lineno == rough_end_site.lineno:
-                    if other_stmt.kind == 'is':
-                        report(WSTRAYIS(other_site, obj_type))
-                        some_stray_is = True
-                elif (other_site.colno > stmt.site.colno
-                      and other_stmt.kind == 'is'):
+                if (other_stmt.kind == 'is'
+                    and (other_site.lineno == rough_end_site.lineno
+                         or other_site.colno > stmt.site.colno)):
                     report(WSTRAYIS(other_site, obj_type))
                     some_stray_is = True
                 else:
--- a/py/dml/dmlparse.py
+++ b/py/dml/dmlparse.py
@@ -230,6 +230,7 @@ def prod(f):
 @prod
 def top(t):
     'dml : maybe_provisional maybe_device maybe_bitorder device_statements'
+    stray_is_check(t[4])
     t[0] = ast.dml(site(t), t[2], t[4])
 
 
@@ -329,6 +330,7 @@ allowed_in_hashif = {
 def toplevel_if(t):
     '''toplevel_if : hashif LPAREN expression RPAREN \
                    LBRACE device_statements RBRACE toplevel_else'''
+    stray_is_check(t[6])
     bad_stmts = [stmt for stmt in t[6] + t[8]
                  if stmt.kind not in allowed_in_hashif]
     if bad_stmts:
@@ -345,6 +347,7 @@ def toplevel_else_no(t):
 @prod_dml14
 def toplevel_else_body(t):
     '''toplevel_else : hashelse LBRACE device_statements RBRACE'''
+    stray_is_check(t[3])
     t[0] = t[3]
 
 @prod_dml14
@@ -896,7 +899,8 @@ def template(t):
     'toplevel : TEMPLATE objident maybe_istemplate LBRACE template_stmts RBRACE'
     ises = [s for s in t[5] if s.kind == 'is']
     shared_methods = [s for s in t[5] if s.kind == 'sharedmethod']
-    if ises and shared_methods:
+    stray_is = stray_is_check(t[5])
+    if not stray_is and ises and shared_methods:
         report(WTEMPLATEIS(ises[0].site))
     t[0] = ast.template(site(t), t[2], t[3] + t[5])
 
@@ -1015,6 +1019,7 @@ def object_spec_none(t):
 @prod
 def object_spec(t):
     'object_spec : object_desc LBRACE object_statements RBRACE'
+    stray_is_check(t[3])
     t[0] = t[1] + t[3]
 
 @prod
@@ -1027,6 +1032,43 @@ def object_statements_empty(t):
     'object_statements : '
     fixup_emptyprod_lexpos(t)
     t[0] = []
+
+def stray_is_check(body):
+    '''Checks a block for any standalone 'is' declared following an object
+    declaration which looks like it was meant to affect that object
+    declaration, and warns for each such 'is' found.
+
+    Returns True if a warning was emitted this way.
+
+    Use this directly on a list of statements yielded by a
+    'object_statements', 'device_statements' or 'template_stmts' symbol,
+    before any manipulation of that list. Otherwise this may yield false
+    positives or negatives.
+    '''
+    some_stray_is = False
+    for (i, stmt) in enumerate(body):
+        if stmt.kind == 'object':
+            (_, obj_type, _, _, obj_body) = stmt.args
+            rough_end_site = (obj_body and obj_body[-1].site) or stmt.site
+            j = i+1
+            while j < len(body):
+                other_stmt = body[j]
+                other_site = other_stmt.site
+
+                if other_site.lineno == rough_end_site.lineno:
+                    if other_stmt.kind == 'is':
+                        report(WSTRAYIS(other_site, obj_type))
+                        some_stray_is = True
+                elif (other_site.colno > stmt.site.colno
+                      and other_stmt.kind == 'is'):
+                    report(WSTRAYIS(other_site, obj_type))
+                    some_stray_is = True
+                else:
+                    break
+
+                j += 1
+
+    return some_stray_is
 
 @prod
 def object_statement(t):
@@ -1084,6 +1126,7 @@ def object_statement_or_typedparam(t):
 @prod_dml14
 def in_each(t):
     'in_each : IN EACH istemplate_list LBRACE object_statements RBRACE'
+    stray_is_check(t[5])
     t[0] = ast.in_each(site(t), [tpl for (_, tpl) in t[3]], t[5])
 
 def validate_if_body(stmts):
@@ -1133,6 +1176,7 @@ def hashelse_nohash(t):
 def object_if(t):
     '''object_if : hashif LPAREN expression RPAREN \
                    LBRACE object_statements RBRACE object_else'''
+    stray_is_check(t[6])
     t[0] = ast.hashif(site(t), t[3], validate_if_body(t[6]), t[8])
 
 @prod
@@ -1144,6 +1188,7 @@ def object_else_no(t):
 @prod
 def object_else_body(t):
     '''object_else : hashelse LBRACE object_statements RBRACE'''
+    stray_is_check(t[3])
     t[0] = validate_if_body(t[3])
 
 @prod
--- a/py/dml/logging.py
+++ b/py/dml/logging.py
@@ -317,7 +317,7 @@ class FileInfo(object):
     with a unique FileInfo instance, which is accessible via all sites
     of the AST.'''
     __slots__ = ('name', 'version', 'bitorder', '_line_offsets',
-                 'utf8_columns', 'provisional')
+                 'provisional')
     def __init__(self, name, version, bitorder='le', content_lines=None,
                  provisional=None):
         name = str(name)
@@ -333,11 +333,6 @@ class FileInfo(object):
             content_lines = list(content_lines)
             self._line_offsets = [0] + list(
                 accumulate(len(line) for line in content_lines))
-            # Temporary py2 hack for expressing utf-8 based column numbers
-            # if line N contains unicode characters, then utf8_column[N][c]
-            # is the character index of the c:th byte
-            utf8_columns = {}
-            self.utf8_columns = utf8_columns
         self.set_name(name)
         self.provisional = provisional or {}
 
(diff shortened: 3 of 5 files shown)


py/dml/messages.py
```diff
@@ -2387,6 +2387,36 @@ class WHOOKSEND(DMLWarning):
...
+    made.
+
+    This typically happens due to a stray semicolon before the `is`, e.g.:
+    ```
+    field f @ [31:0]; is read_only;
+    ```
+    or
+    ```
+    field f @ [31:0];
+            is read_only;
```
lwaern-intel: as you could probably guess it's this second part that first led me down the `lexpos` rabbit hole. though when I realized it was also why all error messages have been screwed up since forever, it kinda got personal

syssimics:
[DML Verification : 6](https://github.com/intel-innersource/applications.simulators.simics.infra.ci-tools/actions/runs/23772210293): ❌ failure

syssimics:
[DML Verification : 7](https://github.com/intel-innersource/applications.simulators.simics.infra.ci-tools/actions/runs/23772210997): ❌ failure

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/23772211878): ❌ failure

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/23772392475): ✅ success

py/dml/dmlparse.py
```diff
@@ -152,6 +152,59 @@ def parse_bitorder(t, syn):
...
+    Hence, we hijack the function the parser uses to acquire tokens
+    (conveniently, Ply actually has an option for that) to stash the most
+    recently lexed token, and then we can use that token's lexpos to fix up the
+    lexpos of every empty production, fixing the problem at the source.
+
+    Ideally instead of peppering fixup_emptyprod_lexpos(t) all over the place
+    we'd have a check in the prod decorators such as to automatically augment
+    it onto any production rule that could yield empty. Unfortunately, that
+    breaks Ply completely. Y'see, Ply does not only rely on `.__doc__` and
+    `.__name__` of functions... but also `.__code__.co_firstlineno`. Brilliant.
```
mandolaerik, line 195: wrong PR, sorry

~Maybe we can create a good-enough test to ensure we have the fixup in place where it's needed? With such test in place, I would conisder the maintenance cost of peppering calls as low; it's reduced to minor noise that you don't have to give another thought until the test heuristics fails.~

~Whatever narrow-minded crude hack is fine, e.g. scan the source code of each production function for a fixup function call, and check that this is exactly the set of functions with `[:|] *\n` in their `__doc__`.~

py/dml/dmlparse.py
```diff
@@ -152,6 +152,59 @@ def parse_bitorder(t, syn):
...
+    latest_tok = t.lexer.latest_token
+    if latest_tok is not None:
+        # HACK set_lexpos not available in Ply 3.4, so internals to the rescue.
+        # No matter; since Ply is unmaintained I can *definitively say* that
+        # this code is guaranteed to be compatible with all future versions of
+        # Ply!
+        # Haha...
+        sym = t.slice[0]
+        sym.lexpos = latest_tok.lexpos
+        sym.lineno = latest_tok.lineno
```
mandolaerik, line 206: Sorry, wrong PR

~From a performance perspective this looks pretty cheap, but it would be satisfying to see some sort of metrics that parse performance is unaffected. E.g., we could count the number of fixup calls needed to parse a fairly large device; something like~
```
    global count
    if count % 10000 == 0: dbg(count)
    count += 1
```
~is probably suffiicient to show convincingly that this is not anywhere near a problem.~

mandolaerik:
Dealing with lexical details from the parser has a somewhat unhygienic smell about it; perhaps a different approach might be to do an isolated check on lexer level, essentially the equivalent of `re.search("field [^\n]*; *is")`. Which admittedly also has a bad smell to it, but a different one.

~I think the emptyprod fixup looks good, and can be merged separately first if you want to.~ oh you did that already

py/dml/dmlparse.py
lwaern-intel: You should be commenting on #430 !

py/dml/dmlparse.py
lwaern-intel: #430!

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/23809330657): ✅ success

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/23809627563): ✅ success

lwaern-intel:
@mandolaerik
> Dealing with lexical details from the parser has a somewhat unhygienic smell about it; perhaps a different approach might be to do an isolated check on lexer level, essentially the equivalent of `re.search("field [^\n]*; *is")`. Which admittedly also has a bad smell to it, but a different one.

The `rough_end_site` bit makes me inclined to agree with you that this solution is dirty, but I'm not convinced that doing it on the lexer level would be any better or easier. In particular, I'm dubious towards a regex because whatever heuristic we use for the warning has to be very context-sensitive. Even with the simplest pattern -- everything being on the same line -- we have to be careful it could not possible match within a comment, or a method, etc. The second pattern -- an `is` being indented underneath -- seems nightmarish to get right.

Working on the `ast` level on the other hand makes is trivial to establish context and let us understand perfectly what situations the warning is emitted. That's kind of a killer benefit and seems hard to beat.

In addition I'll note this is not _completely_ without precedence, as `port_dml` relies on the line and column number reported by the porting messages in order to perform its transformations.

py/dml/dmlparse_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+if bad_emptyprod_rules:
+    sys.stderr.write(
+        "empty production rule(s) without call to 'fixup_emptyprod_lexpos':\n")
+    for rule in bad_emptyprod_rules:
+        sys.stderr.write(f'    {rule.__name__}\n')
+
+    sys.exit(1)
+
+def parse(contents):
```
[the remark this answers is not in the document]
lwaern-intel: You don't read comments of the PR, do you? Not even the opening message? What a curious oversight. If you're gonna say wrong things, you could at least make sure it's the right PR for it.

py/dml/logging.py
```diff
@@ -356,20 +354,17 @@ def set_name(self, name):
...
-                self.utf8_columns, {p.tag(): site
-                                    for (p, site) in self.provisional.items()})
+                {p.tag(): site for (p, site) in self.provisional.items()})
     def __setstate__(self, data):
         from . import provisional
         (self.name, self.version, self.bitorder, self._line_offsets,
-         self.utf8_columns, provisionals) = data
+         provisionals) = data
         self.provisional = {provisional.features[tag]: site
                             for (tag, site) in provisionals.items()}
```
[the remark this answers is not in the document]
lwaern-intel: Nah. These are only relevant for `.dmlast`:s, which get invalidated by anything we touch in `dml-builtins.dml` or `dmlparse.py` anyway. Users don't manually `.dmlast` to our knowledge, and the `.dmlast`:s that get distributed as part of Simics Base are guaranteed to be in sync.

RELEASENOTES.md
```diff
@@ -261,3 +261,22 @@
...
+  number in rare situations.
+
+  This issue had always been present in DML, and most notably affected error
+  messages surrounding methods, as well as object declarations since
+  Simics 7.78.
+- `note 6`  DMLC now detects and warns about any standalone `is` that looks
+  like it was instead intended for an object declared immediately before it,
+  such as the following:
+  ```
+  // Stray semicolons causes each 'is' to be standalone, affecting the
```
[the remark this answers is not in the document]
lwaern-intel: dunno what happened to your formatting but yes

py/dml/dmlparse.py
```diff
@@ -829,7 +899,8 @@ def template(t):
     'toplevel : TEMPLATE objident maybe_istemplate LBRACE template_stmts RBRACE'
     ises = [s for s in t[5] if s.kind == 'is']
     shared_methods = [s for s in t[5] if s.kind == 'sharedmethod']
-    if ises and shared_methods:
+    stray_is = stray_is_check(t[5])
+    if not stray_is and ises and shared_methods:
```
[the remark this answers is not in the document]
lwaern-intel: You'd think that a state-of-the-art AI would recognize that a diff that goes out of its way to change something is indeed meant to change that thing.

A stray `is` is too likely to also incur `WTEMPLATEIS`, which would just be useless noise. At worst, if there is indeed a valid `WTEMPLATEIS` case that does suppressed by the stray `is`, then it'll simply get warned about once the stray `is` is addressed.

RELEASENOTES.md
```diff
@@ -261,3 +261,22 @@
...
 - `release 6 6457`
+- `note 6` Addressed an issue stemming from an upstream bug that would commonly
+  cause the reported column number in warning and error messages to be
+  incorrect (fixes SIMICS-23466). This issue also affected the reported line
+  number in rare situations.
+
+  This issue had always been present in DML, and most notably affected error
+  messages surrounding methods, as well as object declarations since
+  Simics 7.78.
+- `note 6`  DMLC now detects and warns about any standalone `is` that looks
```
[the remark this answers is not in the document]
lwaern-intel: another victory for the Stupid Mistake Checker!

py/dml/dmlparse.py
```diff
@@ -955,8 +1030,46 @@ def object_statements(t):
...
+                if other_site.lineno == rough_end_site.lineno:
+                    if other_stmt.kind == 'is':
+                        report(WSTRAYIS(other_site, obj_type))
+                        some_stray_is = True
+                elif (other_site.colno > stmt.site.colno
+                      and other_stmt.kind == 'is'):
+                    report(WSTRAYIS(other_site, obj_type))
+                    some_stray_is = True
+                else:
+                    break
```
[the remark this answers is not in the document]
lwaern-intel: Done intentionally, though honestly, this could go either way. The situations where the difference matters are so niche that the choice is basically arbitrary.

My thinking went: anyone intentionally putting multiple declarations on the same line is a weirdo, so it's safe to assume by default that something bad is going on as long as that remains the case, no matter what form those declarations take; it's only once we encounter a declaration on a new line that we should tread more lightly.

You _could_, however, imagine there'd be a user who thinks it'd be a great pattern to declare a bunch of fields on the same line:
```
field f1 @ [1]; field f2 @ [2]; field f3 @ [3];
```
In which case
```
field f1 @ [1]; field f2 @ [2]; is read_only; field f3 @ [3]; 
```
rather obviously only concerns `f2`, and so it'd be best to implement Copilot's suggestion.
On the other hand...
```
field f1 @ [1]; field f2 @ [2]; field f3 @ [3];
    is read_only;
```
This would read as though intended to affect *every* field, but wouldn't be detected with Copilot's suggestion. Though that's not exactly a killer argument cause the only reason why this gets detected with the existing code is because it deduce it concerns `f1` -- but not any other field. Worse still, if it was intended to affect every field, then following the suggestion would resolve the warning but only make the `is` affect `f3`:
```
field f1 @ [1]; field f2 @ [2]; field f3 @ [3]
    is read_only;
```

Again though. So niche the choice basically doesn't matter. But since I did manage to invent a _semi_-semi-plausible case where the current implementation would hurt rather than help, I suppose I'll go with what Copilot is saying.

RELEASENOTES.md
```diff
@@ -261,3 +261,22 @@
...
+  like it was instead intended for an object declared immediately before it,
+  such as the following:
+  ```
+  // Stray semicolons cause each 'is' to be standalone, affecting the
+  // containing register rather than the fields
+  field f1 @ [15:0]; is read_only;
+
+  field f2 @ [31:16];
+      is write_only;
+  ```
```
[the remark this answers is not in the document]
lwaern-intel: The indentation surrounding `extern typedef layout` almost certainly stemmed from the automatic conversion when we moved from `.docu` to `.md`.
The two-space indentation is to place it in line with the paragraph of the bullet, and renders without issue.

py/dml/messages.py
```diff
@@ -2387,6 +2387,36 @@ class WHOOKSEND(DMLWarning):
            + "information about the differences between 'send' and 'send_now'")
 
 
+class WSTRAYIS(DMLWarning):
+    """
+    A standalone `is` statement was found that looks like it was instead
+    intended to affect a preceding object declaration rather than the enclosing
+    object/template in which the `is` statement and (sub)object declaration is
```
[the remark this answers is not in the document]
lwaern-intel: thanks Stupid Mistake Checker!

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/23818329407): ✅ success

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/24000038431): ✅ success

test/1.4/errors/T_WSTRAYIS.dml
```diff
@@ -0,0 +1,56 @@
+/*
+  © 2026 Intel Corporation
+  SPDX-License-Identifier: MPL-2.0
+*/
+dml 1.4;
+
+device test;
```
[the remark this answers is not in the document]
lwaern-intel: Huh, for warnings specifically we actually are consistently using `/// COMPILE-ONLY` except when we actually have reason to run Simics. I know that we are extremely inconsistent about it when it comes to errors, as `/// COMPILE-ONLY` _should_ be meaningless for errors; the exit status prevents the Simics run. 

As an aside, if you take this to its logical conclusion, `/// COMPILE-ONLY` doesn't go far enough and you should also do `/// NO-CC`! Honestly though the point is rather moot, even with `/// COMPILE-ONLY`; the time cost is minimal (except for tests like `T_WIMMAFTER.dml` where not having Simics run is _crucial_.) The consistency aspect is the only real reason we do it.

py/dml/dmlparse.py
```diff
@@ -1028,6 +1033,40 @@ def object_statements_empty(t):
...
+    Use this directly on a list of statements yielded by a
+    'object_statements', 'device_statements' or 'template_stmts' symbol,
+    before any manipulation of that list. Otherwise this may yield false
+    positives or negatives.
+    '''
+    some_stray_is = False
+    for (i, stmt) in enumerate(body):
+        if stmt.kind == 'object':
+            (_, obj_type, _, _, obj_body) = stmt.args
+            rough_end_site = (obj_body and obj_body[-1].site) or stmt.site
```
vadholmen: the `if` expression is the idiomatic construct you are looking for

py/dml/dmlparse.py
```diff
@@ -1028,6 +1033,40 @@ def object_statements_empty(t):
...
+            j = i+1
```
vadholmen: `i + 1`

test/1.4/errors/T_WSTRAYIS.dml
mandolaerik, line 1: I would say each call to stray_is_check in the parser deserves one case here. I think this is the main cost of doing it in the parser instead of the lexer, and looks like it's a small cost. This convinces me that your approach is better than a lexer based, whose impl would probably get messier.

I think there is an intrinsic value in these test cases: it concretizes the things we warn for, and lets us contemplate if you can conceive use cases where a stray `is` can make sense.

One interesting  example I can think of is:
```
connect x { interface y; is z; }
```
where I would say it is highly unlikely that `is z` is intended for the interface, assuming the code compiles (because the only templates that are applicable to both connect and interface are the generic ones like `init` and `hard_reset`, and these are seldom applicable on `interface` objects). So this is a way to trigger a WSTRAYIS that not really is helpful. (this is not intended as an argument against your PR, it is just an interesting corner to contemplate).

Still, you can argue that `connect x is z { interface y; }` is a more idiomatic way to phrase the same, so the warning is not really wrong. With that in mind, a different way to approach the original problem could be to stricten the style rule further, and enforce (by warning) that standalone `is` statements within a block should appear before subobject declarations. Which would be way more controversial warning, so it is clearly outside the scope of this PR.

mandolaerik:
> I'm not convinced that doing it on the lexer level would be any better or easier. In particular, I'm dubious towards a regex because whatever heuristic we use for the warning has to be very context-sensitive. Even with the simplest pattern -- everything being on the same line -- we have to be careful it could not possible match within a comment, or a method, etc.

I was thinknig of analyzing the token stream directly. We do that sometimes in `port_dml`. This would *in principle* be equivalent with a very complex regexp, like "`field` followed by some tokens including one `;`, followed directly by `is`, all on the same line". That would address concerns about comments etc., but it would still be somewhat dirty, and there is also `local i_t *is; local field foo; is = NULL;`, so your approach is probably better.

py/dml/dmlparse.py
lwaern-intel: but i _like_ abusing boolean operators!
bah, fine

lwaern-intel: In truth, this was an artifact from my early development of this PR; i I had a reason to do it this way rather than `if` expressions, because this way I also catch if `obj_body[-1].site` is `None`. Since then, however, I realized that isn't possible in the cases `stray_is_check` is meant to be used.

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/24501884947): ✅ success

mandolaerik (approved):

PR #433, opened 2026-04-10
Add references to TQMICs in more places

aeliasso:


doc/1.4/language.md
```diff
@@ -2991,6 +2999,9 @@ It follows that:
   }
   ```
 
+[Template-qualified method implementation calls](#template-qualified-method-implementation-calls) is one way of resolving overrides that are ambiguous because they have the same rank.
```
lwaern-intel: Nit: conflicts don't typically arise due to two definitions of the same rank, but rather, because of two definitions of unrelated ranks (neither rank is higher than the other.) This is because ranks form a partial order rather than a total order. I'll clean this up with my commits.

Tangent: I don't think it's possible to provoke `EAMBINH` from definitions of the same rank at all. DMLC makes some checks that assumes that if two definitions are of the same rank, then they are being made in the same block; which it catches and reports as name collisions instead (`ENAMECOLL`). This is evidenced by this rather funny case that we unit-test:
```
template structural_clash {
    // EAMBINH might be a better description of this clash
    /// ERROR ENAMECOLL
    param p default 1;
    group g {
        /// ERROR ENAMECOLL
        param p default 2;
    }
}
group gg is (structural_clash) {
    group g is (structural_clash);
}
```
Here the `param p` definitions which the template places in two structurally distinct blocks are made to conflict, and stupid-dumb DMLC still reports it as a name collision rather than definitional ambiguity because both `param` definitions have the same rank.

lwaern-intel:
@aeliasso I've touched up some stuff, and added some new stuff of my own. Take a look; I wanna make sure I haven't changed anything in a way you disagree with, and also check if I've written anything weird or confusing I need to explain better.

doc/1.4/language.md
```diff
@@ -2861,6 +2885,30 @@ This section describes in detail the rules for how DML handles when there are
...
+
+Establishing dependencies may not always be desirable. In particular, it's
+possible to have conflicting templates which provide different orthogonal
+functionality, such that it would be wrong to make one depend upon the other.
+An alternate way to resolve conflicting method implementations in such cases is
+to leverage [template-qualified method implementation
+calls](#template-qualified-method-implementation-calls).
+
+The technical rules for resolution of overrides, collectively called the
+*ranking algorithm*, are as follows:
```
lwaern-intel: I couldn't resist. I figured I may as well officially coin the term since us DMLC developers use it all the time.

py/dml/messages.py
```diff
@@ -139,9 +139,13 @@ def log(self):
...
 
 class EAMBINH(DMLError):
-    """If a method or parameter has multiple definitions, then there must
+    """<a id="EAMBINH"/>
+    If a method or parameter has multiple definitions, then there must
     be a unique definition that overrides all other definitions.
+    See [Resolution of Overrides](language.html#resolution-of-overrides)
     """
+    # 'Resolution of overrides' does not exist in the 1.2 reference manual
+    version = "1.4"
```
lwaern-intel, line 148: @mandolaerik This makes me frown, but I suppose we could be OK with it. We don't care overmuch about the 1.2 reference manual rotting, do we?

py/dml/messages.py
```diff
@@ -183,7 +187,7 @@ def log(self):
 
 class EAMBDEFAULT(DMLError):
     """A method may not invoke its default implementation if multiple
-    methods are overridden, and the template inheritance graph is
+    met<a id="ENSHAREDTQMIC"/>hods are overridden, and the template inheritance graph is
```
lwaern-intel: buah, what?

lwaern-intel: pretend this never happened

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/24242094486): ✅ success

doc/1.4/language.md
mandolaerik: Technically, the list is a set of rules that defines a relation, whereas an algorithm would be a method for how to compute a representation of this relation based on a representation of code. One could argue that there is a pretty close and obvious duality between algorithms and the things they compute, so calling it an algorithm is not overly problematic, but since it is somewhat technically incorrect I would prefer at least to avoid coining the term before we have a need for it elsewhere in the documentation. We already coin the term "rank", which makes the informal term "ranking algorithm" pretty unambiguous (it is the algorithm for computing ranks)

py/dml/messages.py
mandolaerik: I agree, somewhat frown-y but not a problem. EAMBINH in 1.2 only appears in the contexts of experimental/undocumented multi-level overrides, or when importing 1.4 files. In 1.2, the sanctioned way to do overrides is to have exactly one default and one non-default, and then the non-default takes precedence.

doc/1.4/language.md
lwaern-intel: > I would prefer at least to avoid coining the term before we have a need for it elsewhere in the documentation.

We never will; I've already established a precedence of speaking of ranking and override resolution in terms of "specificity" rather than ranks in other parts of the docs for simplicity's sake. So the answer is just to remove the "collectively called the *ranking algorithm*" part and never reintroduce it.

aeliasso:
This looks good to me overall. I don't have an opinion on whether you introduce *ranking algorithm* here or not.

lwaern-intel (commented):
@mandolaerik Since I wrote half of this PR it'd be better if you are the one to approve it.

doc/1.4/language.md
```diff
@@ -2861,6 +2885,29 @@ This section describes in detail the rules for how DML handles when there are
...
+other, based on the contexts in which each definition is made. Problems related
+to override resolution can typically be understood and solved purely through
+that lens, without the need to understand the technical rules in detail.
+
+For example, if you are a modeller introducing an implementation of a method
+`m()`, only to have DMLC complain that it conflicts with an earlier definition
+made in a template `t`, then that is because DMLC is unable to spot any
+dependency establishing that the context you're working in is more specific than
+the body of `t`. The most common solution to that is to make that dependency
+clear by adding `is t` to the object/template in which you are defining `m()`.
```
mandolaerik, line 2900: I like this section, it concretizes an intuitive understanding of specificity before we deep-dive into the technical definition. Perhaps we should extend this a bit and offer an understanding of the scope of specificity, e.g. by saying 2-3 examples instead of just one. In particular, it is relevant to say that imports affect specificity.

Also, specificity is essentially a synonym of rank, so maybe we should s/rank/specificity/ further down.

Neither needs to be done within this PR.

mandolaerik (approved):

PR #435, opened 2026-04-14
Include optional interfaces in attribute docs

mandolaerik:
- **Change iface to tt**
- **Include optional interfaces in attribute docs**

### DIFF
--- a/py/dml/c_backend.py
+++ b/py/dml/c_backend.py
@@ -554,7 +554,7 @@ def generate_attribute_common(initcode, node, port, dimsizes, prefix,
         if ifaces:
             doc += (
                 '\n\nRequired interfaces: '
-                + ', '.join('<iface>' + i.name + '</iface>' for i in ifaces)
+                + ', '.join('<tt>' + i.name + '</tt>' for i in ifaces)
                 + '.')
     doc = mkStringConstant(node.site, doc)
 
--- a/py/dml/structure.py
+++ b/py/dml/structure.py
@@ -2946,14 +2946,20 @@ class InterfacesDocParamExpr(objects.ParamExpr):
     def mkexpr(self, indices):
         if self.cached is not None:
             return self.cached
-        ifaces = [i for i in self.node.get_components('interface')
-                  if param_bool(i, 'required')]
-        if ifaces:
-            self.cached = mkStringConstant(
-                self.site,
-                '\n\nRequired interfaces: '
-                + ', '.join('<tt>' + i.name + '</tt>' for i in ifaces)
-                + '.')
+        iface_nodes = self.node.get_components('interface')
+        if iface_nodes:
+            (required, optional) = ([], [])
+            for i in iface_nodes:
+                (required if param_bool(i, 'required') else optional).append(i.name)
+            text = ''
+            for (req, ifaces) in [('Required', required),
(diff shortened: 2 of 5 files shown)


syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/24413533314): ✅ success

py/dml/structure.py
```diff
@@ -2946,14 +2946,21 @@ def __init__(self, site, node):
...
-            self.cached = mkStringConstant(
-                self.site,
-                '\n\nRequired interfaces: '
-                + ', '.join('<iface>' + i.name + '</iface>' for i in ifaces)
-                + '.')
+        iface_nodes = self.node.get_components('interface')
+        if iface_nodes:
+            iface_lists = [[], []]
+            for i in iface_nodes:
+                iface_lists[param_bool(i, 'required')].append(i.name)
```
lwaern-intel: Hehe. It'd be hypocritical of me to complain about this, so I won't.
The alternative which you can use if you prefer is to do:
```
(required, optional) = ([], [])
for i in iface_nodes:
    (required if param_bool(i, 'required') else optional).append(i.name)
```

py/dml/c_backend.py
```diff
@@ -554,7 +554,7 @@ def generate_attribute_common(initcode, node, port, dimsizes, prefix,
         if ifaces:
             doc += (
                 '\n\nRequired interfaces: '
-                + ', '.join('<iface>' + i.name + '</iface>' for i in ifaces)
+                + ', '.join('<tt>' + i.name + '</tt>' for i in ifaces)
```
lwaern-intel, line 557: Not documenting optionals introduces a discrepancy vs. 1.4, but I don't really care. It might be worth noting though via a comment (_might,_ I hardly insist.)

py/dml/structure.py
```diff
@@ -2946,14 +2946,21 @@ def __init__(self, site, node):
...
+            (optional, required) = iface_lists
+            text = ''
+            for (req, ifaces) in [('Required', required),
+                                  ('Optional', optional)]:
+                if ifaces:
+                    s = 's' if len(ifaces) > 1 else ''
```
lwaern-intel, line 2958: Or just:
```f'\n\n{req} interface(s): '```
far be it from me to complain about a small extra bit of engineering, though.

lwaern-intel (approved):

py/dml/structure.py
mandolaerik: This string is documentation that we force our users to force-feed to their end-users, so better not be sloppy. Test failure strings served to fellow developers along with a stack trace is a different story; there spartan messages are ok since we can assume the reader to deduce context from the source code.

py/dml/c_backend.py
mandolaerik: Ignored on purpose, I don't want to write a test. I did the tt change because I realized `<iface>` may be a real obstacle when migrating to markdown.

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/24464676726): ✅ success

py/dml/structure.py
mandolaerik: Good point, fixed

PR #436, opened 2026-04-21
Silence Coverity complaints about overflow

jhbaarnh:


syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/24769068928): ✅ success

lib/1.4/dml-builtins.dml
```diff
@@ -2301,6 +2302,12 @@ template bank is (object, shown_desc) {
     }
 
     shared method get(uint64 offset, uint64 size) -> (uint64) throws default {
+        if (size == 0 || size > 8) {
+            log error: "%s.get() call at 0x%llx with invalid size %d",
+                this._qname(), offset, size;
+            throw;
+        }
```
lwaern-intel, line 2309: @mandolaerik why have we never done this before, again?

also do you think this warrants a Jenkins run? This particular commit is not mandatory to unblock Base.

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/24775197490): ✅ success

lib/1.4/dml-builtins.dml
mandolaerik: No, I'm not worried. It is certainly *possible* that they have written code that accidentally relies on oversized inquiry reads to return bogus data instead of rejecting (just insert a bad read call in a test in such a way  that the broken value happens to not break the test, but where an exception would break it). But we can deal with that in staging if it happens.

But perhaps it does warrant a release note.

test/1.4/lib/T_bank.dml
```diff
@@ -0,0 +1,27 @@
...
+*/
+dml 1.4;
+
+device test;
+
+/// DMLC-FLAG --breaking-change=shared-logs-locally
+
+attribute trigger_test is write_only_attr {
+    param type = "n";
+    method set(attr_value_t a) throws {
```
mandolaerik, line 13: Minor nitpick: I would have used `get`, mainly because it's 100% apparent for everyone that you got through the method if you return an actual attribute value (if the attribute did return None, then  you *must* have reached the `return SIM_make_attr_nil()` statement). Using `set` is also 100% safe, but the Simics API has a few pragmatic corners where errors are halfway forgiven, so the reader must deduce that these are not applicable.

mandolaerik: oh, never mind, the log expectation does cover this sufficiently.

mandolaerik (approved):
Consider a releasenote

lib/1.4/dml-builtins.dml
mandolaerik: It just struck me: we do have some logic to split up long accesses in 8-byte chunks. So this error path is only applicable for explicit calls within DML, not for inquiry get/set. Which makes me less worried.

lwaern-intel: >  not for inquiry get/set

Well, `io_memory_access` can still get manually invoked.

PR #441, opened 2026-05-08
Fix b3b0554 breaking Windows

lwaern-intel:


syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/25543436953): ✅ success

lwaern-intel:
[Works on a custom run](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/25543843033). Don't be fooled by it failing, that's just 'cause I limited the build targets to `core clock set-memory` -- the ones needed for t126. If you check the logs, t126 passes.

PR #442, opened 2026-05-08
t126: Don't run `list-failed-modules` on module load failure

lwaern-intel:
`list-failed-modules` causes a ton of logspam which, other than just creating a bunch of noise, actually somehow breaks Windows's stdout handling.


syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/25548851616): ✅ success

andreashedstrom-intel (approved):

mandolaerik (approved):

PR #450, opened 2026-07-14
Fix qname memory leak

TSonono:
This was revealed when running some PCIe device tests with leak sanitizer

Closes #375

### DIFF
--- a/py/dml/c_backend.py
+++ b/py/dml/c_backend.py
@@ -1741,6 +1741,7 @@ def generate_deinit(device):
     # Let the posted event deallocate the immediate after state
     out('_dev->_immediate_after_state->deleted = true;\n')
     out('}\n', preindent = -1)
+    out('_DML_free_qname_cache(&_dev->_qname_cache);\n')
 
     out('}\n\n', preindent = -1)
     splitting_point()
--- a/include/simics/dmllib.h
+++ b/include/simics/dmllib.h
@@ -2262,6 +2262,15 @@ __qname(dml_qname_cache_t *cache, const char *fmt, ...)
         return (const char *)s;
 }
 
+UNUSED static void
+_DML_free_qname_cache(dml_qname_cache_t *cache)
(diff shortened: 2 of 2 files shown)


lwaern-intel:
Closes #375. Putting the call in `_dealloc` also addresses the concern I brought up in that PR.

lwaern-intel (approved):
Approved modulo verification.

py/dml/c_backend.py
```diff
@@ -1605,6 +1605,9 @@ def generate_dealloc(device):
     start_function_definition(
         f'void {crep.cname(device)}_dealloc(conf_object_t *dev)')
     out('{\n', postindent = 1)
+    out(crep.structtype(device) + ' *_dev = ('
+        + crep.structtype(device) + ' *)dev;\n')
+    out('_DML_free_qname_cache(&_dev->_qname_cache);\n')
```
lwaern-intel: Actually -- the `_DML_free_qname_cache` call should be placed at the end of `_deinit` (see `generate_deinit`), rather than inside `_dealloc`. That's to be consistent with where we place the deallocation for the other pieces of detached state we have, those being the state for subsequent logging behavior, as well as the immediate after queue.

lwaern-intel (changes requested):

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/29324955333): ✅ success

py/dml/c_backend.py
TSonono: Ok! I'll move it. I just put it in dealloc due to the name seeming to match with what I was doing 😅

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/29374055984): ✅ success

lwaern-intel (approved):

PR #452, opened 2026-08-24
`WARNING` pragma

lwaern-intel:


RELEASENOTES.md
```diff
@@ -284,3 +284,7 @@
   ```
 - `release 7 7152`
 - `release 6 6461`
+- `note 6` Warning and error messages reported by DMLC now display the tag for
+  the warning/error kind by default. The old behavior of having the tag
+  be omitted now requires passing `--no-tags` to DMLC. With this change, the
+  `-T` flag has become a no-op and will be removed in a future version.
```
lwaern-intel: Keeping `-T` around is a spitball on my part; I'm interested in your opinion. Also instead of vaguely saying "a future version" we could tie it to the specified `--simics-api` and not accept `-T` when used with API version 8 or above.

doc/1.4/language.md
```diff
@@ -259,7 +260,50 @@ Any C line corresponding to the call to `some_function(...)` will receive
...
+line-based pragma (like `COVERITY`). For example:
+```
+param special default -1;
+param then_level default 1;
+
+method check(uint32 i) {
+    /*% WARNING WREDUNDANTLEVEL %*/
+    /*% WARNING WNEGCONSTCOMP %*/
+    log info, 1 then then_level: "%s", i == special ? "special" : "regular";
+}
```
lwaern-intel, line 292: You have *no idea* how difficult it was to concoct a semi-realistic case where applying a `WARNING` pragma twice on the same line would be warranted. We don't have enough warning kinds ;)

py/dml/dmlc.py
```diff
@@ -364,12 +364,16 @@ def main(argv):
         + ' dependency generation. Specify multiple times to have multiple'
         + ' targets.')
 
-    # <dt>-T</dt>
-    # <dd>Show tags on warning messages. The tags can be used with
-    # the <tt>-\-nowarn</tt> and <tt>-\-warn</tt> options.</dd>
+    # <dt>-\-no-tags</dt>
+    # <dd>Disables showing the tag associated with each error or warning
+    # message.</dd>
```
lwaern-intel, line 369: should we even bother?

py/dml/dmlparse.py
```diff
@@ -1547,14 +1547,19 @@ def cdecl2_ptr(t):
...
+        # compatibility
+        vsite = site(t)
+        if vsite.dml_version() != (1, 2):
+            if not breaking_changes.vect_needs_provisional.enabled:
+                report(WEXPERIMENTAL(vsite,
+                                     "vect types\nTo suppress this "
+                                     + "warning, specify "
+                                     + "'provisional simics_util_vect;' at "
+                                     + "the top of the file, after"
+                                     + 'dml 1.4;'))
```
lwaern-intel: I did this as I was concerned about the novel `WARNING` pragma being used instead of the much more preferable option to leverage `provisional simics_util_vect;`

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/32695963275): ❌ failure

py/dml/toplevel.py
```diff
@@ -226,25 +228,32 @@ def parse_pragma(filename, start_lineno, end_lineno, pragma, data):
...
+def process_pragma(t):
+    (pragma, filename, start_lineno, tgt_lineno, data) = t
+    if pragma == 'WARNING':
+        if not is_warning_tag(data):
+            report(EWARNING(SimpleSite(f'{filename}:{start_lineno}:1'), data))
+            return None
+
+    (dml.globals.line_pragmas
+     .setdefault((filename, tgt_lineno), (start_lineno, []))
+     [1].append((pragma, data)))
```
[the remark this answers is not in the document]
lwaern-intel: yawn
.dmlast files have no forward-compatibility guarantees

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/32723908899): ❌ failure

lwaern-intel:
Verification failure is unrelated.

py/dml/dmlc.py
mandolaerik: no, drop it. Nobody will use it.

doc/1.4/language.md
mandolaerik: Then possibly better to write an even simpler and clearer example with just one warning?

There is a value in documenting formally how sequences of pragmas interact, but perhaps this is better  moved to a common section? e.g., I suspect your doc still doesn't formally specify what the following means:
```
/*% WARNING ... %*/
/*% COVERITY ...  %*/
... one suspicious line ... 
```

PR #453, opened 2026-09-01
Plug memory leak

mandolaerik:
Returning 0 means "stop iteration", so only the first sub-table would
be freed.


lwaern-intel (approved):
Ah, the perils of mysterious APIs (and not reading API documentation)

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/33486332381): ✅ success

PR #455, opened 2026-09-07
Permit both versions 3.4 and 3.11 of ply

mandolaerik:
- **Rip out vestigial grammar rules**
- **Demand Ply 3.11 -- SIMICS-23703**
- **Permit both versions 3.4 and 3.11 of ply**


syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/34158109717): ❌ failure

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/34160763754): ✅ success

py/dml/toplevel.py
```diff
@@ -50,15 +47,38 @@ def get_parser(version, tabmodule=None, debugfile=None):
...
+            with tempfile.TemporaryDirectory(dir='.') as d:
+                parser = yacc.yacc(
+                    module = dml.dmlparse.grammars[version],
+                    tabmodule=tabmodule,
+                    debug=False,
+                    optimize=0,
+                    outputdir=d)
+                import shutil
+                shutil.move(os.path.join(d, tabfile), tabfile)
+            sys.path.append('.')
```
mandolaerik: The idea here is to write something horrible that is good enough to not cause any trouble until the definitive 3.11 bump. E.g., the `move` can probably cause a race on Windows, and in theory it is not sound to assume `.` is writable. But we will get away with it.

generate_parsetabs.py
```diff
@@ -12,13 +12,16 @@ def generate_parsetabs(dml_package, version, parsetab, debugfile):
...
+    # find and reuse as a cache is the previously installed
+    # dml.<parsetab> module under dml_package/dml. Remove it and any .pyc
+    # variant of it to deprive ply of the possibility to carry out any caching
+    # logic.
     for suff in ['.py', '.pyc']:
-        if os.path.lexists(parsetab + suff):
-            os.remove(parsetab + suff)
+        installed = os.path.join(dml_package, 'dml', parsetab + suff)
+        if os.path.lexists(installed):
+            os.remove(installed)
```
lwaern-intel, line 24: This code was to handle 3.11. You mean it... still works with 3.4? That's dubious. Have you tried `make dmlc` -> touch dmlparse -> `make dmlc` again?

py/dml/toplevel.py
lwaern-intel: Wow this is dirty. But after some thinking and reasoning, I **think** this should work out... I think.
Although;
1. I don't like how it unconditionally overrides `tabmodule`. Put an `assert tabmodule is None` at the beginning of this path, or something.
2. Why not move it underneath `dml` instead of `.`? That way you don't have to do `sys.path.append`. You just have to set `tabmodule = f'dml._dml{"".join(map(str, version)}parsetab'`.
3. Perhaps name the parsetab module `dml_parsetab_ply311` or something. It'd be helpful for understanding why it exists if somebody spots it.

py/dml/toplevel.py
```diff
@@ -50,15 +47,38 @@ def get_parser(version, tabmodule=None, debugfile=None):
...
     parser = yacc.yacc(
         module = dml.dmlparse.grammars[version],
         method='LALR',
         tabmodule=(tabmodule
                    or 'dml.dml%s_parsetab' % (''.join(map(str, version)),)),
         debug = debugfile is not None,
         debugfile = debugfile,
-        optimize = 0)#not debug_mode)
+        optimize = 0,
+        outputdir = '.')
```
lwaern-intel: This hunk is redundant (in ply 3.4). But it also doesn't do any damage.

lwaern-intel (changes requested):

py/dml/toplevel.py
mandolaerik: 1. good point
2. `dml.` often resides in the base package which might not be writable
3. good point

generate_parsetabs.py
mandolaerik: Yes, and also the 3.4 based Verification build is green.

py/dml/toplevel.py
lwaern-intel: ...
_Ah._
I thought before DMLC is invoked the CWD is always set to the location of the DML python module. But that's not correct; the CWD is only *relevant* if the parsetab modules can't be found, which is only the case when `generate_parsetabs` is invoked, which is working underneath `linub64/obj/modules/dmlc` as CWD; and then `copy_python` copies the genned parsetabs over to the proper DML module.

This is no good then. I looked into it, and both under make and under cmake, the CWD when DMLC is invoked will be the working directory of the target. That means the parsetabs will have to be regenned for _every DML device that gets built._ That's not acceptable. We need another solution.

syssimics:
[DML Verification : 7](https://github.com/intel-innersource/applications.simulators.simics.infra.ci-tools/actions/runs/34230066338): ❌ failure

lwaern-intel (changes requested):

py/dml/toplevel.py
```diff
@@ -50,15 +47,44 @@ def get_parser(version, tabmodule=None, debugfile=None):
         return parsers[version]
 
     lexer = lex.lex(module = dml.dmlparse.lexers[version],
-                    optimize = 0)#not debug_mode)
+                    optimize = 0,
+                    outputdir = '.')
+    cache_dir = '.'
+    if lex.__version__ != '3.5':
```
lwaern-intel: 3.5?

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/34230070931): ❌ failure

py/dml/toplevel.py
```diff
@@ -50,15 +47,44 @@ def get_parser(version, tabmodule=None, debugfile=None):
...
+        # For now we only provide pre-compiled parse tables for PLY 3.4;
+        # for 3.11 we create the parse tables on the fly instead.
+        # The plan is to switch to pre-compiled parse tables for 3.11
+        # when we bump mini-python to PLY 3.11.
+        assert tabmodule is None
+        tabmodule = "_dml%s_parsetab_ply%s" % (
+            ''.join(map(str, version)), lex.__version__.replace('.', ''))
+        cache_dir = os.environ.get('DMLC_PARSE_CACHE_DIR', '.')
```
lwaern-intel, line 61: I still prefer an error instead of defaulting to use `.`.

Usage together with ply 3.11 will very much be non-standard in this one month we are introducing it. `DMLC_PARSE_CACHE_DIR` would be an explicit escape hatch for that purpose. You won't silently get the inferior caching behavior, and instead the error tells you exactly what you need to do, and why it even is that the error occurs.

Also disregard everything I just wrote here because there's a bigger issue we need to discuss on R.C.

py/dml/toplevel.py
```diff
@@ -50,15 +47,44 @@ def get_parser(version, tabmodule=None, debugfile=None):
...
+        tabfile = f'{cache_dir}/{tabmodule}.py'
+        if not os.path.isfile(tabfile):
+            # crude good-enough caching: parsetab is typically generated once
+            # per directory. The `move` ensures atomicity in case of two
+            # parallel runs. There is possibly an unlikely race on Windows.
+            import tempfile
+            with tempfile.TemporaryDirectory(dir='.') as d:
```
lwaern-intel: `dir=cache_dir`. Would that also implicitly create it?

generate_parsetabs.py
lwaern-intel: Verification being green isn't saying much. This code is to specifically deal with the case that old, invalidated parsetabs has to get cleaned away.

lwaern-intel (changes requested):

py/dml/toplevel.py
lwaern-intel: Discussed off-site; actually, defaulting is the one way we could get this to work as we have planned.

lwaern-intel: Wait no nevermind, we *still* don't wanna default... *right now.* When it comes time to bump 1000 to package ply 3.11 parsetabs and we need to invert this logic, *then* we gonna want to default (maybe even rip out `DMLC_PARSE_CACHE_DIR` and use `.` unconditionally).

py/dml/snyk-requirements.txt
```diff
@@ -1 +1 @@
-ply==3.4
+ply==3.11
```
lwaern-intel: So uh do we know what impact this has yet?
Maybe this too should get `ply>=3.4`:d.

lwaern-intel (changes requested):

py/dml/snyk-requirements.txt
mandolaerik: Don't know. Let's keep it at 3.4 for now, that's the CI tested path.

py/dml/toplevel.py
mandolaerik: I don't see why we shouldn't default. The point with default is zero cognitive overhead: at the cost of an acceptably small perf hit, people can bump versions individually without having to know about the compat hack.

syssimics:
[DML Verification : 7](https://github.com/intel-innersource/applications.simulators.simics.infra.ci-tools/actions/runs/34244200393): ❌ failure

py/dml/toplevel.py
lwaern-intel: I explained my rationale off-site. To reiterate (in a slightly refined form:

>My suggestion would be: if and only if DMLC is run in a Ply 3.11 environment, it'll leverage DMLC_PARSE_CACHE_DIR. If it's empty, you error something like:
```
For DMLC to be used with Ply 3.11 the DMLC_PARSE_CACHE_DIR environment variable must be set (a path to house generated parsing tables)
```

>If people aren't fiddling with build systems they aren't using ply 3.11; the cognitive overhead of `DMLC_PARSE_CACHE_DIR` is specifically for people who are diverging from the default 1033 flow, and also introducing ply 3.11-
>What I like about DMLC_PARSE_CACHE_DIR (and an error message if ply 3.11 is used without it) is that it is very explicitly an escape hatch for use in the purpose of using ply 3.11
>You won't silently get the inferior caching behavior, and instead the error tells you exactly what you need to do, and why it is that the error even occurs

_However_, this is one of the things I actually don't insist on. Especially given how if we bump the ply version in base's pyproject.toml, just the act of using host python will probably pull in ply 3.11

py/dml/toplevel.py
mandolaerik: fixed. Doesn't mkdir automatically and we don't care. This is not a polished UI.

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/34244205135): ❌ failure

syssimics:
[PR Verification](https://github.com/intel-innersource/applications.simulators.simics.simics-base/actions/runs/34260490264): ✅ success

py/dml/toplevel.py
lwaern-intel: Defaulting to `.` will also allow bumping 1033 without bumping 1000 in a month's time when we make the switch proper. Although that's the tiniest of benefits. Nobody would care about bumping 1033 if it's not to enable 1000.

generate_parsetabs.py
lwaern-intel: Confirmed independently, the regen flow still works, surprisingly enough. I'm not sure why, but I'm not looking a gift horse in the mouth.

lwaern-intel (approved):

